PR #9, opened 2026-07-23
fix: log cancel vs failure and stop abort spam

ming0627:
## Summary
Codex multi-task load produced hundreds of opaque `[codex-router] request failed` log lines with no error detail. Live traffic after this change shows those were mostly client aborts (`AbortError: This operation was aborted`), not provider outages.

## Changes
- Add shared helpers: `isAbortError`, `formatErrorForLog` (redacts caller secret URLs), `logRequestFailure`
- Log `request canceled` vs `request failed` with the actual error text
- Treat aborts as HTTP `499` instead of generic `502`
- Apply the same handling in Codex router, API forwarder, Kimi OAuth, Grok OAuth, and Cursor router
- Only flash Dynamic Island error state for real `5xx` statuses (not cancel/4xx)

## Root cause
Top-level `.catch` handlers discarded the error object and always printed `request failed`. Concurrent Codex tasks frequently cancel in-flight requests when switching tasks, stopping turns, or superseding work.

## Impact
- Operators can distinguish cancel noise from real failures
- Logs no longer look like mass provider breakage under multi-task use
- Caller capability secrets remain redacted in log text

## Test plan
- [x] `node --test test/http-utils.test.mjs` (3/3 pass)
- [x] Live service restart after patch; doctor green
- [x] Observed live log lines: `[codex-router] request canceled: AbortError: This operation was aborted: code=20`
- [ ] Optional: full package test suite on CI

### DIFF
--- a/src/api-forwarder.mjs
+++ b/src/api-forwarder.mjs
@@ -3,6 +3,8 @@ import http from "node:http";
 import {
   HOP_BY_HOP_HEADERS,
   httpErrorStatus,
+  logRequestFailure,
+  isAbortError,
   pipeResponse,
   readRequestBody,
   requireInternalAuth,
@@ -222,8 +224,17 @@ async function handleRequest(request, response) {
 
 const server = http.createServer((request, response) => {
   handleRequest(request, response).catch((error) => {
+    logRequestFailure("api-forwarder", error);
+    if (isAbortError(error)) {
+      if (!response.headersSent) {
+        response.writeHead(499);
+        response.end();
+      } else if (!response.writableEnded) {
+        response.destroy();
+      }
+      return;
+    }
     const status = httpErrorStatus(error);
-    console.error("[api-forwarder] request failed");
     if (!response.headersSent) {
       writeJson(response, status, {
         error: {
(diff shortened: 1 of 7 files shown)


duolahypercho:
Heads up @ming0627 — while triaging the open PRs we merged #22 ("Don't enter the error state when the client cancels a request"), which lands on `main` now and covers the same client-abort / error-spam problem this PR targets: cancels and mid-request client disconnects no longer flip router health into the error state.

Leaving this one open since it's still a draft and yours is more focused on the *logging* side (distinguishing cancel vs. genuine failure in the logs). Could you rebase on the latest `main` and check whether there's anything here that #22 didn't already handle? If the logging distinction still adds value, happy to review and merge; if #22 fully covers it, feel free to close. Your call.

duolahypercho:
Thanks for this — the cancel-vs-failure distinction is a real improvement, and splitting abort noise out of genuine error logging is the right call.

Two things blocking it right now:

1. **Merge conflicts.** The branch is currently `CONFLICTING` against `main`. `src/router.mjs` and `src/api-forwarder.mjs` have both moved since this was opened. A rebase on latest `main` should sort it.
2. **Still marked draft.** Once the rebase is clean, mark it ready for review and CI will run.

One heads-up for the rebase: `src/router.mjs` has active work landing around the upgrade/abort paths (see #28), so expect that file to need the most attention.

Happy to take another look as soon as it's rebased.

duolahypercho:
closed cuz it's solved!

PR #10, opened 2026-07-23
Confirm hidden API-key entry and catch doubled pastes

kdegeek:
## Problem

The hidden API-key prompt (`bin/provider-key <id> set`) disables terminal echo, so pasting a key gives no indication that anything was captured. Pasting twice by accident saves a silently broken doubled credential that only surfaces later as an upstream auth failure. Empty input crashed with a raw stack trace from `writeProviderCredential`.

## Change

After each entry the prompt now:

- reports how many characters were received (`Received 42 characters.`) — the value itself is never echoed or logged, matching the existing zero-knowledge posture;
- challenges input whose halves are identical — the doubled-paste signature, back-to-back or whitespace-separated — with an explicit `Save it anyway? [y/N]` before anything is written;
- re-prompts up to three times on empty or discarded input instead of failing with a stack trace.

The Windows `Read-Host -AsSecureString` path already shows masked feedback while typing; the post-entry confirmation now behaves identically on all platforms.

Detection and feedback are pure functions in a new `src/secret-entry.mjs` with unit tests (`test/secret-entry.test.mjs`), following the `kimi-oauth-onboarding.mjs` pattern; `provider-key.mjs` keeps only the terminal wiring.

## Verification

- `npm run check`, `npm test` (134 tests, 0 fail), `sh -n install.sh` + all `bin/*`, `npm audit --omit=dev` clean.
- Interactive flows exercised in a real PTY (`expect`) against an isolated `CODEX_ROUTER_STATE_DIR`: normal entry, doubled paste → decline → retry → save, and three declined attempts → `Nothing was saved.` + exit 1 with no file written.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
   flips router health into the eight-second error state, so tray and island
   status indicators stop flashing red on ordinary cancels. Errors the router
   or an upstream actually produced still surface.
+- The hidden API-key prompt now confirms how many characters were captured
+  after each entry, challenges input that looks like the same key pasted
+  twice before saving, and re-prompts instead of failing on empty input, so a
+  paste with terminal echo disabled is no longer a silent leap of faith.
 - Added a reversible tray toggle that lets signed-out Codex CLI/App sessions
   use connected external providers through a managed custom model provider,
   while preserving ChatGPT credentials and restoring the prior provider mode.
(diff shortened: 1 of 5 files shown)


kdegeek:
For the record on the earlier GitGuardian findings: the flagged strings were fabricated test fixtures (`sk-test-…`, `plan-key-a1b2c3d4e5f6`) in the new unit test file, not real credentials — nothing existed to revoke. They've been replaced with plainly low-entropy, word-based fixtures per the repo's fixture conventions, and the scan is now clean.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #25, opened 2026-07-24
Fix Codex Voice routing through native realtime endpoints

### LINKED ISSUE #24: [Bug]: Voice Mode fails with 404 Unsupported router route for /v1/live
### What happened?

Codex Voice Mode fails through codex-router because the new /v1/live endpoint is not supported, returning a 404 error.

Image

### App target

Codex

### Platform

macOS

_No response_

snarkyalyx:
## Summary

- keep Codex Voice WebRTC call creation on the native ChatGPT backend
- keep the Voice sideband WebSocket on OpenAI's native realtime endpoint
- preserve user-owned realtime endpoint overrides and custom ChatGPT base URLs
- document the update/re-enable workflow for existing installations

## Root cause

Codex Router manages `openai_base_url` so Responses API traffic can be dispatched between native and external models. The new Voice flow inherited that Responses-only router URL for its separate realtime transports, causing WebRTC call creation to request `/v1/live` from a router that does not implement the endpoint.

The managed configuration now supplies transport-specific native realtime endpoints. Normal model inference continues through Codex Router, while Voice avoids the incompatible Responses-only route.

## Validation

- `npm run check`
- `node --test test/config-manager.test.mjs` — 13 passed
- `npm test` — 165 passed, 2 skipped

Fixes #24

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -225,6 +225,16 @@ This is expected. Codex Router declines the optional Responses WebSocket
 upgrade, and current Codex falls back to compressed HTTP. A warning alone is not
 a failed model request.
 
+## Voice Mode reports an unsupported `/v1/live` route
+
+Codex Voice uses native realtime endpoints that are separate from the Responses
+API. Current installs keep the WebRTC call and its sideband WebSocket on Codex's
+native endpoints instead of sending them through the Responses-only router.
+
+Run `./bin/enable` again after updating, fully quit Codex, and reopen it so the
+managed realtime overrides take effect. User-owned realtime endpoint overrides
+are preserved.
+
 ## Uninstall retained files
 
 This is intentional. `./bin/uninstall` removes only the active integration and
--- a/src/config-manager.mjs
+++ b/src/config-manager.mjs
@@ -36,6 +36,10 @@ const endMarker = "# END codex-router-managed";
 const providerStartMarker = "# BEGIN codex-router-provider-managed";
 const providerEndMarker = "# END codex-router-provider-managed";
 const routerProviderId = "codex-router";
+const defaultChatgptBaseUrl = "https://chatgpt.com/backend-api";
+const defaultRealtimeWebsocketBaseUrl = "https://api.openai.com/v1";
+const realtimeCallBaseUrlKey = "experimental_realtime_webrtc_call_base_url";
+const realtimeWebsocketBaseUrlKey = "experimental_realtime_ws_base_url";
 const markerPairs = [
   [startMarker, endMarker],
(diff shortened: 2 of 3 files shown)


duolahypercho:
Thanks — this is well built. Guarding on `rootHasValue` so user-owned realtime endpoints are never clobbered, deriving the call URL from `chatgpt_base_url`, and the round-trip test asserting `disable` restores the file byte-for-byte are all the right instincts.

**Blocker: this overlaps with #27.** Both PRs solve "Codex Voice hits an unsupported route", but with opposite designs:

- **#25** (this one) keeps voice *out* of the router via `experimental_realtime_*_base_url` config overrides.
- **#27** lets voice *through* the router, adding a `/v1/live` passthrough that forwards to native.

Both are defensible; they shouldn't both land without a deliberate call. Worth settling explicitly: should the router be transparent to voice, or should voice bypass it at the config layer?

CI had never actually run here — the workflow was sitting in `action_required`. I've approved it, so real check results should appear shortly.

Holding for the design call, not for code quality.

duolahypercho:
Following up on the #25-vs-#27 design question — going with this one. Merging.

**Reasoning.** I dug into how Codex resolves the realtime endpoints and couldn't definitively pin down whether `experimental_realtime_ws_base_url` defaults to something derived from `openai_base_url` or to an independent host. But that unknown turns out not to matter, because this PR is correct either way:

| | #25 (this PR) | #27 |
|---|---|---|
| WS default is independent | works | works |
| WS default derives from `openai_base_url` | works — pinned off the router | breaks — WS leg still hits `server.on("upgrade")` → `426` |

#27 only adds an HTTP `POST /live` route; it never touches the upgrade handler. This PR pins **both** realtime base URLs, so it covers both cases. It dominates, so it wins.

Two things I confirmed against Codex 0.145.0 while checking:

- `experimental_realtime_ws_base_url` and `experimental_realtime_webrtc_call_base_url` are both real root keys in `ConfigToml` — this isn't relying on anything invented.
- There's also a `[realtime] transport` key with `webrtc` and `websocket` variants. If those are alternatives rather than simultaneous legs, a user on `transport = "websocket"` is precisely the case #27 could not have served.

**One follow-up, not a blocker.** This hardcodes `https://api.openai.com/v1` as the WS base. I couldn't confirm from the Codex binary that this matches Codex's own default when the key is unset — if it doesn't, we'd be pinning a host that's correct-by-accident. Could you share where that value came from? If it's from Codex docs or source, a comment citing it would be worth adding. If it was inferred, worth an issue to verify against a live Voice session.

The `rootHasValue` guard preserving user-owned endpoints and the byte-for-byte `disable` restore test are what made this easy to trust. Thanks for the care on those.

PR #27, opened 2026-07-24
Support native Codex voice passthrough

otsinc15:
## Summary
- pass `/v1/live` directly to the native Codex backend
- preserve voice request content negotiation and native OpenAI authentication
- keep external model routing, including Kimi K3, isolated to Responses routes

## Verification
- `npm run check`
- `npm test` (164 passed, 2 skipped)

### DIFF
--- a/src/router.mjs
+++ b/src/router.mjs
@@ -221,6 +221,23 @@ function nativeHeaders(request) {
   return headers;
 }
 
+function nativeLiveHeaders(request) {
+  const headers = nativeHeaders(request);
+  const contentType = request.headers["content-type"];
+  if (contentType !== undefined) {
+    headers["Content-Type"] = Array.isArray(contentType)
+      ? contentType.join(", ")
+      : contentType;
+  }
+  for (const name of ["accept", "content-encoding"]) {
+    const value = request.headers[name];
+    if (value !== undefined) {
+      headers[name] = Array.isArray(value) ? value.join(", ") : value;
+    }
+  }
+  return headers;
+}
+
 function routedHeaders() {
   return {
     Authorization: `Bearer ${INTERNAL_KEY}`,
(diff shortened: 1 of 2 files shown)


duolahypercho:
Thanks — the implementation here is solid. Header allowlisting, the browser-origin rejection, abort wiring, and the test asserting Kimi routing is untouched are all exactly right.

**Blocker: this overlaps with #25.** Both PRs solve "Codex Voice hits an unsupported route", but with opposite designs:

- **#25** keeps voice *out* of the router — it writes `experimental_realtime_*_base_url` config overrides so Codex talks to native endpoints directly.
- **#27** (this one) lets voice *through* the router — adding a `/v1/live` passthrough that forwards to native.

These need a maintainer decision before either lands, and merging both without one risks two half-owned paths for the same traffic. Worth settling explicitly: should the router be transparent to voice, or should voice bypass it at the config layer?

Note on the red check: `test (windows-latest)` is failing on `alias reads pick up rewritten state and reverse lookups resolve` — that's a **pre-existing flake on `main`**, unrelated to this diff. Not counting it against this PR.

Holding for the design call, not for code quality.

duolahypercho:
Closing this in favor of #25, which has now merged. Want to be clear that this is **not** a code-quality call — the implementation here is genuinely good, and I'd rather explain properly than close it silently.

**Why #25 instead.** Both PRs fix "Voice Mode breaks when the router is enabled," but with opposite designs: #25 keeps voice traffic *out* of the router via config overrides, this one routes it *through*. I couldn't definitively determine whether `experimental_realtime_ws_base_url` defaults to a host derived from `openai_base_url` or an independent one. That unknown decided it:

| | #25 | #27 (this PR) |
|---|---|---|
| WS default is independent | works | works |
| WS default derives from `openai_base_url` | works | **breaks** |

This PR adds `handleNativeLive` for `POST /live` and `/v1/live`, but doesn't touch `server.on("upgrade")` in `src/router.mjs`, which still answers `426 Upgrade Required`. So if the realtime WebSocket leg follows `openai_base_url`, it would still land on the router and get rejected — voice fixed on the call leg, still broken on the socket leg. #25 pins both URLs and is correct in either case.

One more thing that pushed it: Codex 0.145.0 has a `[realtime] transport` key with `webrtc` and `websocket` variants. If those are alternatives rather than simultaneous legs, anyone on `transport = "websocket"` was never served by this approach.

**What was good here, sincerely:** the header allowlist in `nativeLiveHeaders`, the browser-origin rejection, the abort wiring on both `request.aborted` and `response.close`, and especially the test asserting Kimi routing stays untouched. That last one is the kind of test most people skip.

**If you want to revisit this:** the passthrough approach has a real advantage #25 lacks — voice shows up in router activity tracking (`setRoute` with `model: "voice"`), where #25 makes it invisible. If you're up for a follow-up that also proxies the WebSocket upgrade rather than 426-ing it, that would be a strictly better end state than where we are now, and I'd review it seriously.

Also worth knowing: the red `test (windows-latest)` on this PR was never your fault — `alias reads pick up rewritten state and reverse lookups resolve` is flaky on `main` independently. Sorry that sat on your PR looking like a failure.

PR #31, opened 2026-07-24
fix(api-forwarder): coalesce consecutive assistant messages for strict providers

snarkyalyx:
## What

Coalesce consecutive assistant messages in `api-forwarder`'s `normalizeBody` so tool-result messages always follow the assistant message carrying the matching `tool_calls`.

## Why

Strict chat-completions providers (e.g. MiniMax) reject a turn whose tool-result messages do not immediately follow the assistant message with the matching `tool_calls`. When the upstream Responses-API history is translated to chat completions, an assistant turn that produced both tool calls and a text message can arrive as two consecutive assistant messages — one with `tool_calls`, one with the visible text — so the subsequent `tool`-role results no longer follow the tool-call-bearing assistant. MiniMax then rejects the whole request:

```
litellm.BadRequestError: OpenAIException - invalid params,
tool call result does not follow tool call (2013).
Received Model Group=minimax-token-plan-minimax-m3
```

This also breaks the routed compaction path, since the compaction request replays the same tool-bearing history through the same translation.

Coalesce runs of consecutive assistant messages into a single assistant message (combining content, unioning `tool_calls`) before forwarding. Coalescing only triggers when at least one adjacent assistant message carries `tool_calls`, so:
- it is a no-op for the Anthropic `/messages` path (which uses content-block `tool_use`, not a `tool_calls` field), and
- it does not change a well-formed single-assistant turn.

## Validation

- `node --check src/api-forwarder.mjs`
- `node --test test/routing.test.mjs` (new test sends `assistant(tool_calls)` + `assistant(content)` + `tool(result)` and asserts the upstream receives one coalesced assistant with both `content` and `tool_calls`, followed by the tool result; all existing tests still pass)
- Full suite: only the 7 pre-existing environmental failures (kimi-oauth, provider-account-usage, provider-usage, service-operation-lock, startup, vendor-quota-adapters) remain, unchanged from `main`.

### DIFF
--- a/src/api-forwarder.mjs
+++ b/src/api-forwarder.mjs
@@ -55,6 +55,55 @@ function deepSeekEffort(value) {
   return ["xhigh", "max", "ultra"].includes(value) ? "max" : "high";
 }
 
+// Strict chat-completions providers (e.g. MiniMax) reject a turn whose tool
+// result messages do not immediately follow the assistant message carrying the
+// matching tool_calls. When the upstream Responses-API history is translated to
+// chat completions, an assistant turn that produced both tool_calls and a text
+// message can arrive as two consecutive assistant messages (one with
+// tool_calls, one with the text), so the tool results no longer follow the
+// tool-call-bearing assistant. Coalesce runs of consecutive assistant messages
+// into a single assistant message (combining content and tool_calls) so the
+// chat-completions contract holds. This is a safe normalization: consecutive
+// assistant messages are not a valid multi-turn shape, so merging them cannot
+// change a well-formed conversation.
+function combineAssistantContent(target, source) {
+  const sourceContent = source.content;
+  if (sourceContent === undefined || sourceContent === null) return;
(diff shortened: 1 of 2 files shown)


snarkyalyx:
Update: this PR also covers the routed **compaction** path for openai-compatible providers (MiniMax, DeepSeek, Qwen, etc.), because the compaction request flows router → LiteLLM → `api-forwarder`, where this coalesce runs — so it fixes the `tool call result does not follow tool call (2013)` compaction failure without needing to strip tool items.

A companion PR (#32) that stripped tool-call items from the compaction input was tried and **reverted**: it defeated `oauth-forwarder`'s existing `foldInterveningAssistantMessages`, which needs the tool-call assistant + tool results present to fold out empty-content intervening assistant messages, and that produced a new `text content is empty` regression on `kimi-oauth-k3`. The Kimi OAuth path was already handled by that fold; this PR handles the openai-compatible paths. See #32 for the details.

PR #32, opened 2026-07-24
fix(router): strip tool-call items from routed compaction input

snarkyalyx:
## What

Strip tool-call and tool-result items from the input before asking the routed model to summarize a conversation (routed compaction).

## Why

Routed compaction sends the conversation to the model with `tools: []` and `tool_choice: "none"` to ask for a handoff summary, but the input still contained the raw `function_call` / `function_call_output` items. When that history is translated to chat completions for a strict provider (e.g. MiniMax), the summarization request is rejected:

```
litellm.BadRequestError: OpenAIException - invalid params,
tool call result does not follow tool call (2013).
Received Model Group=minimax-token-plan-minimax-m3
```

because the tool-bearing history is forwarded with no tools for the results to follow. From the user's side this shows up as compaction failing on every switch into a strict provider (the switch triggers compaction when the accumulated conversation exceeds the new model's context window, and the compaction then routes to the strict model and fails).

Drop tool-call and tool-result items (`function_call`, `function_call_output`, `custom_tool_call`, `custom_tool_call_output`) before summarizing, leaving a clean narrative of user and assistant message items (including assistant `output_text`) plus the compact prompt. The summarizer does not need the raw tool I/O, and the request can no longer trip the strict-provider tool-ordering rule.

This complements #31 (coalescing consecutive assistant messages), which fixes the same tool-ordering issue for regular turns; this PR fixes the compaction path specifically and also reduces the summarization payload.

## Validation

- `node --check src/router.mjs`
- `node --test test/routing.test.mjs` (new test sends a compaction request with `function_call` + `function_call_output` items and asserts the gateway's summarize request contains no tool items while message items and the compact prompt are preserved; all existing tests still pass)
- Full suite: only the 7 pre-existing environmental failures (kimi-oauth, provider-account-usage, provider-usage, service-operation-lock, startup, vendor-quota-adapters) remain, unchanged from `main`.

### DIFF
--- a/src/router.mjs
+++ b/src/router.mjs
@@ -318,6 +318,25 @@ function normalizeRoutedInput(input) {
     });
 }
 
+function stripToolCallItems(input) {
+  if (!Array.isArray(input)) return input;
+  // Compaction asks the model for a handoff summary, not a tool turn. The raw
+  // function_call / function_call_output items are not useful to the summarizer
+  // and, once translated to chat completions for strict providers (e.g.
+  // MiniMax), can trip "tool call result does not follow tool call" because the
+  // summarization request carries no tools but the tool-bearing history still
+  // does. Drop tool-call and tool-result items before summarizing so the
+  // request is a clean narrative of user/assistant messages; message items
+  // (including assistant output_text) are preserved.
+  const TOOL_ITEM_TYPES = new Set([
+    "function_call",
+    "function_call_output",
+    "custom_tool_call",
+    "custom_tool_call_output",
+  ]);
+  return input.filter((item) => !TOOL_ITEM_TYPES.has(item?.type));
+}
+
 function normalizeNativeInput(input) {
   if (!Array.isArray(input)) return input;
   return input.map((item) => {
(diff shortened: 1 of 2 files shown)


snarkyalyx:
Closing this in favor of #31.

After deploying, this change caused a regression on the Kimi OAuth path: `litellm.BadRequestError: OpenAIException - text content is empty. Received Model Group=kimi-oauth-k3`.

Root cause of the regression: `kimi-oauth-k3` is routed through `oauth-forwarder.mjs`, whose `foldInterveningAssistantMessages` already folds (and splices out) intervening assistant messages — including empty-content ones — that sit between a tool-call assistant and its tool results. It can only do that when the tool-call assistant and tool results are present in the messages. By stripping `function_call` / `function_call_output` items from the compaction input in `summarize()`, this PR removed exactly the anchors that `foldInterveningAssistantMessages` needs, so empty-content assistant messages from the history were no longer folded out and reached Kimi, which rejects them with "text content is empty". Confirmed by the router log: zero `text content is empty` errors before this change, the first one immediately after.

This PR is also redundant: #31 (coalescing consecutive assistant messages in `api-forwarder`) already fixes the `tool call result does not follow tool call (2013)` compaction failure for `openai-compatible` providers (MiniMax, DeepSeek, Qwen, etc.), because the compaction request flows router → LiteLLM → `api-forwarder`, where the coalesce runs. The Kimi OAuth path was already handled by `foldInterveningAssistantMessages` before this PR.

Reverting. The minimax compaction 2013 stays fixed via #31.

PR #40, opened 2026-07-28
fix(config): avoid unsupported router websockets

kdegeek:
## Summary
- select an authenticated managed router provider for normal Codex sessions
- declare WebSocket transport unsupported for both authenticated and login-free router providers
- preserve and restore the user's prior provider and model state across enable, disable, and login-free transitions

## Verification
- `npm test` (170 passed, 0 failed, 2 skipped)
- `npm run check`
- `git diff --check`
- live Codex transport probe against the local router

### DIFF
--- a/src/config-manager.mjs
+++ b/src/config-manager.mjs
@@ -36,6 +36,7 @@ const endMarker = "# END codex-router-managed";
 const providerStartMarker = "# BEGIN codex-router-provider-managed";
 const providerEndMarker = "# END codex-router-provider-managed";
 const routerProviderId = "codex-router";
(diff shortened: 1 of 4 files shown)


duolahypercho:
Thanks for investigating this and validating the transport behavior. We’re not going to merge this approach because current Codex successfully falls back from the WebSocket probe to HTTP, and changing `model_provider` during normal authenticated use conflicts with the router’s config-preservation guarantees.

We may separately add `supports_websockets = false` to the existing login-free custom provider, where that setting can be applied without changing the authenticated provider. Closing this PR, but the transport investigation was useful—thank you.

PR #47, opened 2026-07-31
fix(windows): detect ChatGPT auth through npm codex.cmd shim

### LINKED ISSUE #46: Native OpenAI models vanish from catalog on Windows (npm codex.cmd shim not detected as authenticated)
## Summary

On Windows, when Codex is installed via npm (`npm i -g @openai/codex`), every native OpenAI model (e.g. `gpt-5.x`) is dropped from the merged model catalog — even though the user is genuinely signed in via ChatGPT. Only the external providers (e.g. DeepSeek) remain in the model picker.

## Root cause

`src/codex-binary.mjs` → `codexIsAuthenticated()`:

1. `findCodexBinary()` runs `where codex` and takes the **first** result. On a Windows npm install that returns:
   ```
   C:\Users\<user>\AppData\Roaming\npm\codex        ← extensionless shell shim, listed FIRST
   C:\Users\<user>\AppData\Roaming\npm\codex.cmd    ← the real spawnable shim
   ```
2. Node's `child_process` **cannot spawn the extensionless entry directly** (`existsSync` says `true` because Windows resolves it to the `.cmd`, but `execFileSync` throws **ENOENT** without `shell: true`).
3. That throw is swallowed by the `catch {}`, so `codexIsAuthenticated()` returns `false`.
4. `src/catalog.mjs` then builds with `includeNative: openaiAuthenticated` → `false`, so `buildMergedCatalog()` **strips all native OpenAI models** from `merged-models.json`.

Notably, `catalog.execCodex()` already handles this correctly (it uses `shell: true` for `.cmd`), but `codexIsAuthenticated()` was missed.

## Reproduction (verified on upstream `main`, commit f2bbe96)

```text
$ where codex
C:\Users\krist\AppData\Roaming\npm\codex
C:\Users\krist\AppData\Roaming\npm\codex.cmd

$ codex login status
Logged in using ChatGPT          # exit 0 — genuinely authenticated
```

Reproducing upstream `findCodexBinary()` + `codexIsAuthenticated()` in Node:

```text
resolved binary:  "C:\Users\krist\AppData\Roaming\npm\codex"   # extensionless, [0]
codexIsAuthenticated -> THREW (error.code: ENOENT) -> returns false
```

And the catalog build output confirms the consequence:

```json
{ "models": 2, "routed_models": 2, "native_models": 0,
  "openai_authenticated": false }
```

i.e. all 7 native models (`gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.5`, `gpt-5.4`, …) gone, only the 2 DeepSeek entries left.

## Environment

- OS: Windows 11 (10.0.26220)
- Node: v25.2.1
- Codex CLI: `codex-cli 0.146.0` (npm global install → `…\AppData\Roaming\npm\codex.cmd`)
- codex-router: `0.4.0-beta.1` @ `f2bbe96`
- Auth: ChatGPT (Plus), valid `auth.json`

## Impact

Windows users with an npm-installed Codex CLI permanently lose all OpenAI models from the router's model picker, and any feature gated on `openaiAuthenticated` misfires, with no error surfaced.

## Proposed fix

In `src/codex-binary.mjs`:
- `findCodexBinary()`: prefer an explicit `.cmd`/`.exe` from `where`/`which` output over the extensionless first entry.
- `codexIsAuthenticated()`: spawn `.cmd` shims with `shell: true` (mirroring `catalog.execCodex()`).

PR incoming.

timm-u:
Fixes #46.

## Problem

On Windows with an npm-installed Codex CLI, `where codex` lists the **extensionless** shim (`…\npm\codex`) before `…\npm\codex.cmd`. Node's `child_process` cannot spawn that extensionless entry directly — `execFileSync` throws **ENOENT**, which `codexIsAuthenticated()` swallows in its `catch {}` and returns `false`.

The catalog then runs with `includeNative: false`, so every native OpenAI model (`gpt-5.x`) is stripped from `merged-models.json` and the model picker shows only the external providers (e.g. DeepSeek) — despite the user being genuinely signed in via ChatGPT.

`catalog.execCodex()` already handles `.cmd` correctly with `shell: true`; this was just missed in the auth check.

## Changes (`src/codex-binary.mjs`)

- **`findCodexBinary()`** — prefer an explicit `.cmd`/`.exe` entry from the `where`/`which` output over the extensionless first entry, so callers get a directly spawnable path.
- **`codexIsAuthenticated()`** — spawn `.cmd` shims with `shell: true`, mirroring `catalog.execCodex()`.

No new dependencies; behaviour unchanged on macOS/Linux (the `.cmd`/`.exe` preference is guarded by `process.platform === "win32"`).

## Verification

Reproduced the bug on upstream `main` (commit `f2bbe96`) and confirmed the fix against a live Windows + npm Codex + ChatGPT auth setup:

```text
# before
resolved binary:  "C:\Users\krist\AppData\Roaming\npm\codex"   # extensionless
codexIsAuthenticated() -> false   (ENOENT swallowed)
catalog: { native_models: 0, openai_authenticated: false }

# after
resolved binary:  "C:\Users\krist\AppData\Roaming\npm\codex.cmd"
codexIsAuthenticated() -> true
catalog: { native_models: 7, openai_authenticated: true, models: 9 }
```

- `npm test` → **193 pass, 0 fail, 2 skipped** (unchanged from baseline)
- `npm run check` → **syntax checks passed**

## Notes

I did not add a unit test for `codexIsAuthenticated()` because mocking `node:child_process` is not viable from ESM here (the `node:child_process` namespace is non-configurable, so `node:test`'s `mock.method` cannot patch the `execFileSync` binding the module imports). Happy to add one if the maintainer prefers an injectable exec surface, but that's a larger change than fits this bugfix.

### DIFF
--- a/src/catalog.mjs
+++ b/src/catalog.mjs
@@ -1,4 +1,3 @@
-import { execFileSync } from "node:child_process";
 import {
   existsSync,
   mkdirSync,
@@ -16,7 +15,7 @@ import {
   NATIVE_ALIAS_PATH,
   NATIVE_CATALOG_PATH,
 } from "./paths.mjs";
-import { codexIsAuthenticated, requireCodexBinary } from "./codex-binary.mjs";
+import { codexIsAuthenticated, runCodex } from "./codex-binary.mjs";
 import { syncRoutedCodexAgents } from "./codex-agent-catalog.mjs";
 import { MODEL_BY_SLUG } from "./model-registry.mjs";
 import { buildNativeAliasAssignments } from "./native-alias.mjs";
@@ -40,20 +39,17 @@ function atomicJson(target, value) {
 function captureNative() {
   const args = ["debug", "models"];
   if (bundled) args.push("--bundled");
+  const options = {
+    encoding: "utf8",
+    timeout: 30_000,
+    maxBuffer: 32 * 1024 * 1024,
+  };
   let output;
   try {
-    output = execFileSync(requireCodexBinary(), args, {
-      encoding: "utf8",
-      timeout: 30_000,
-      maxBuffer: 32 * 1024 * 1024,
-    });
+    output = runCodex(args, options);
   } catch (error) {
     if (bundled) throw error;
-    output = execFileSync(requireCodexBinary(), ["debug", "models", "--bundled"], {
-      encoding: "utf8",
-      timeout: 30_000,
-      maxBuffer: 32 * 1024 * 1024,
-    });
+    output = runCodex(["debug", "models", "--bundled"], options);
   }
   const parsed = JSON.parse(output);
   if (!parsed || !Array.isArray(parsed.models) || parsed.models.length === 0) {
--- a/src/codex-binary.mjs
+++ b/src/codex-binary.mjs
@@ -35,10 +35,10 @@ export function findCodexBinary() {
     if (process.platform === "win32") {
       // `where codex` lists the extensionless npm shim first (e.g.
       // `...\npm\codex`), which Node cannot spawn directly. Prefer the
-      // explicit `.cmd`/`.exe` entries it also returns so callers get a
-      // directly executable path.
+      // explicit `.cmd`/`.bat`/`.exe` entries it also returns so callers get
+      // a directly executable path.
       return (
-        found.find((candidate) => candidate.toLowerCase().endsWith(".cmd")) ||
+        found.find((candidate) => /\.(cmd|bat)$/i.test(candidate)) ||
         found.find((candidate) => candidate.toLowerCase().endsWith(".exe")) ||
         found[0]
       );
@@ -59,22 +59,24 @@ export function requireCodexBinary() {
   return binary;
 }
 
-// On Windows, npm-installed CLIs resolve to a `.cmd` shim, which Node cannot
-// spawn without a shell (throws ENOENT/EINVAL). Centralize that handling here so
-// every codex invocation goes through one path. `findCodexBinary()` already
-// prefers the explicit `.cmd`, so callers can pass the resolved binary through
-// without re-checking the extension.
+// On Windows, npm-installed CLIs resolve to a `.cmd`/`.bat` shim, which Node
+// cannot spawn without a shell (throws ENOENT/EINVAL). Centralize that handling
+// here so every codex invocation goes through one path. `findCodexBinary()`
+// already prefers the explicit shim, so callers can pass the resolved binary
+// through without re-checking the extension.
 export function runCodex(args, options = {}) {
   const binary = requireCodexBinary();
   const merged = { ...options };
-  if (
-    process.platform === "win32" &&
-    binary.toLowerCase().endsWith(".cmd") &&
-    merged.shell === undefined
-  ) {
+  const needsShell =
+    process.platform === "win32" && /\.(cmd|bat)$/i.test(binary);
+  if (needsShell && merged.shell === undefined) {
     merged.shell = true;
   }
-  return execFileSync(binary, args, merged);
+  // With shell:true, Node concatenates file+args without quoting, so a path
+  // containing spaces (e.g. C:\Users\John Smith\...) would split into two
+  // tokens and fail. Quote the command; cmd.exe /d /s /c strips the outer
+  // quotes correctly, and this branch is Windows-only.
+  return execFileSync(merged.shell ? `"${binary}"` : binary, args, merged);
 }
 
 export function codexIsAuthenticated() {
--- a/src/codex-binary.mjs
+++ b/src/codex-binary.mjs
@@ -59,21 +59,29 @@ export function requireCodexBinary() {
   return binary;
 }
 
+// On Windows, npm-installed CLIs resolve to a `.cmd` shim, which Node cannot
+// spawn without a shell (throws ENOENT/EINVAL). Centralize that handling here so
+// every codex invocation goes through one path. `findCodexBinary()` already
+// prefers the explicit `.cmd`, so callers can pass the resolved binary through
+// without re-checking the extension.
+export function runCodex(args, options = {}) {
+  const binary = requireCodexBinary();
+  const merged = { ...options };
+  if (
+    process.platform === "win32" &&
+    binary.toLowerCase().endsWith(".cmd") &&
+    merged.shell === undefined
+  ) {
+    merged.shell = true;
+  }
+  return execFileSync(binary, args, merged);
+}
+
 export function codexIsAuthenticated() {
   const binary = findCodexBinary();
   if (!binary) return false;
-  const options = { timeout: 10_000, stdio: "ignore" };
   try {
-    // On Windows, npm-installed CLIs resolve to a `.cmd` shim, which Node
-    // cannot spawn without a shell (throws ENOENT/EINVAL, silently
-    // misreporting the user as logged out and dropping native OpenAI models
-    // from the merged catalog). Mirror the shell:true handling already used
-    // by catalog.execCodex().
-    if (process.platform === "win32" && binary.toLowerCase().endsWith(".cmd")) {
-      execFileSync(binary, ["login", "status"], { ...options, shell: true });
-    } else {
-      execFileSync(binary, ["login", "status"], options);
-    }
+    runCodex(["login", "status"], { timeout: 10_000, stdio: "ignore" });
     return true;
   } catch {
     return false;
--- a/src/codex-binary.mjs
+++ b/src/codex-binary.mjs
@@ -28,10 +28,22 @@ export function findCodexBinary() {
   if (direct) return direct;
   const finder = process.platform === "win32" ? "where.exe" : "which";
   try {
-    return execFileSync(finder, ["codex"], {
+    const found = execFileSync(finder, ["codex"], {
       encoding: "utf8",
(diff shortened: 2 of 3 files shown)


duolahypercho:
Thanks for narrowing this down. One follow-up appears necessary for the fix to work end-to-end on Windows: after `findCodexBinary()` prefers `codex.cmd`, `codexIsAuthenticated()` now succeeds with `shell: true`, but `src/catalog.mjs`'s `captureNative()` still calls `execFileSync(requireCodexBinary(), ...)` without `shell: true` (both the normal and `--bundled` fallback paths). That means the catalog can report `openai_authenticated: true` and then fail to capture native models with `ENOENT` when it runs `codex.cmd`. Could you apply the same Windows-aware invocation there, or centralize it in a helper? The PR description says the catalog already handles `.cmd`, but that handling is not present in the base code I checked.

timm-u:
Thanks @duolahypercho — you're right on both counts, and good catch.

**On the incomplete scope:** correct, `captureNative()` in `catalog.mjs` (both the normal path and the `--bundled` fallback) and the model-catalog probe in `doctor.mjs` all called `execFileSync(<binary>, ...)` directly with no `shell: true`. So after `codexIsAuthenticated()` started succeeding via `codex.cmd`, those paths would still throw `ENOENT` and fail to capture/recheck native models. Fixed in 5b10371.

**On the stale PR description:** also correct, and my mistake. The "catalog already handles `.cmd`" line came from a local working copy that had a `catalog.execCodex()` helper — which does **not** exist in `main`. I should have verified against the base branch before writing that; I've corrected it below. Apologies for the noise.

## What changed (5b10371)

Centralized the Windows `.cmd` handling instead of duplicating it per call site. New helper in `codex-binary.mjs`:

```js
export function runCodex(args, options = {}) {
  const binary = requireCodexBinary();
  const merged = { ...options };
  if (
    process.platform === "win32" &&
    binary.toLowerCase().endsWith(".cmd") &&
    merged.shell === undefined
  ) {
    merged.shell = true;
  }
  return execFileSync(binary, args, merged);
}
```

Every codex invocation now goes through it:
- `codexIsAuthenticated()` → `runCodex(["login","status"], …)`
- `catalog.captureNative()` → both paths → `runCodex(["debug","models", …], …)`
- `doctor.mjs` catalog probe → `runCodex(["debug","models"], …)`

Callers can still force a behavior by passing `options.shell` explicitly (opt-out), and macOS/Linux are unaffected (the `.cmd` branch is guarded by `process.platform === "win32"`).

## Verification

Reproduced on the same live Windows + npm `codex.cmd` + ChatGPT auth setup:
- `codexIsAuthenticated()` → `true`
- `captureNative()` path (`runCodex(["debug","models","--bundled"], …)`) → captures **8 native models** (was `ENOENT`)
- full `node src/catalog.mjs` → `{ native_models: 7, openai_authenticated: true, models: 9 }`
- `npm test` → 193 pass / 0 fail / 2 skipped (unchanged)
- `npm run check` → passed

Diff is now 3 files (+43/−20). Ready for another look when you have a moment.

duolahypercho:
Thanks for the detailed writeup — the diagnosis is correct and I want this in. Centralizing on `runCodex()` so `catalog.mjs` and `doctor.mjs` get the same treatment was the right call; both had the identical latent bug.

One blocking issue before merge.

### `shell: true` does not quote the command, so paths with spaces break

`runCodex()` passes `binary` to `execFileSync` unquoted with `shell: true`. Node builds the shell command by concatenating file and args without quoting, so a path like `C:\Users\John Smith\AppData\Roaming\npm\codex.cmd` splits into two tokens and fails. Windows usernames with spaces are common enough that this will bite people.

Reproduced locally (same unquoted-concatenation path on POSIX, easier to demo):

```
no-shell:     ok login status
shell:true:   FAILED — /bin/sh: /…/dir: No such file or directory
quoted+shell: ok login status
```

This matters more than it did before the PR, because `captureNative()` and the doctor `debug models` call now route through `runCodex` too — previously they ran without a shell and handled spaces fine.

Suggested fix:

```js
export function runCodex(args, options = {}) {
  const binary = requireCodexBinary();
  const merged = { ...options };
  const needsShell =
    process.platform === "win32" && /\.(cmd|bat)$/i.test(binary);
  if (needsShell && merged.shell === undefined) merged.shell = true;
  // shell:true does not quote the command, so a path with spaces
  // (e.g. C:\Users\John Smith\...) would split into two tokens.
  return execFileSync(merged.shell ? `"${binary}"` : binary, args, merged);
}
```

`cmd.exe /d /s /c` strips the outer quotes correctly, so this is safe on Windows and a no-op elsewhere.

Worth widening the extension check to `.bat` while you're in there, and matching it in `findCodexBinary()`'s preference list.

### Minor

The description says `catalog.execCodex()` already handled `.cmd` with `shell: true` — there's no `execCodex` and no `shell: true` anywhere in `src/` on `main`. Doesn't change anything about the patch, just noting it so the rationale in the commit history stays accurate.

### Verification on my side

- Test suite on your branch: 173 pass / 8 fail — **identical failure set to baseline `f2bbe96`**, so no regression from this change. The 8 are pre-existing and unrelated (Kimi OAuth, provider usage, quota adapters).
- CI hasn't actually run yet — it's sitting at `action_required` pending maintainer approval. I'll approve the run so the Windows leg of the matrix executes against the updated patch.

No unit test needed — I agree mocking `node:child_process` from ESM isn't worth restructuring the module for on a bugfix this size.

Push the quoting fix and I'll merge.

timm-u:
Pushed the quoting fix and the `.bat` widening in 53f7913. Thanks for the reproduce — confirmed it independently before patching.

## Verification (Windows)

Isolated the pure quoting behavior with a minimal `.cmd`/`.bat` in a spaced path (`C:\Users\krist\AppData\Local\Temp\dir with space\`):

| mode | result |
| --- | --- |
| `no-shell` | `EINVAL` (Node can't spawn `.cmd` at all) |
| `shell:true` unquoted | `FAILED` — path splits at the space: `'C:\...\dir' is not recognized` |
| `shell:true` quoted | `OK: "ran HELLO"` |

And against the real (space-less) `codex.cmd`:
- `codexIsAuthenticated()` → `true`
- `captureNative()` path → 8 native models
- `npm test` → 193 pass / 0 fail / 2 skipped (unchanged)
- `npm run check` → passed

## Changes (53f7913)

```js
export function runCodex(args, options = {}) {
  const binary = requireCodexBinary();
  const merged = { ...options };
  const needsShell =
    process.platform === "win32" && /\.(cmd|bat)$/i.test(binary);
  if (needsShell && merged.shell === undefined) {
    merged.shell = true;
  }
  // With shell:true, Node concatenates file+args without quoting, so a path
  // containing spaces (e.g. C:\Users\John Smith\...) would split into two
  // tokens and fail. Quote the command; cmd.exe /d /s /c strips the outer
  // quotes correctly, and this branch is Windows-only.
  return execFileSync(merged.shell ? `"${binary}"` : binary, args, merged);
}
```

- Widened the shim check from `.cmd` to `/\.(cmd|bat)$/i` in both `runCodex()` and `findCodexBinary()`'s preference list, since `.bat` has the identical spawn limitation.
- `findCodexBinary()` now prefers `/\.(cmd|bat)$/i` before `.exe` before the extensionless first entry.

## On the minor note

Agreed on the stale `execCodex`/`shell: true` rationale — there is no `execCodex` and no `shell: true` anywhere in `src/` on `main`. That line in the original commit message came from a local working copy that diverged from base; it's inaccurate and I shouldn't have written it. The later commits describe the actual change correctly (centralizing on `runCodex()`), and nothing in the patch depends on that rationale. Happy to reword the first commit message if you'd prefer the history clean — just say the word.

Full PR diff is now 3 files (+45/−20). Ready when you are.

timm-u:
Closing this — the fix landed in 6580247 on `main`, which supersedes everything here and goes further.

I pulled and verified 6580247 on the same live Windows + npm `codex.cmd` + ChatGPT auth setup:
- `codexIsAuthenticated()` → `true`, `codexAuthStatus()` → `{ authenticated: true, reason: "authenticated", binary: "…\codex.cmd" }`
- `captureNative()` via `runCodex()` → 8 native models
- `npm test` → 214 pass / 0 fail (incl. the new `codex-binary.test.mjs`, 9/9)

The merged fix matches the direction of this PR — prefer the spawnable `.cmd`/`.bat`/`.exe` over the extensionless PATH entry, centralize on `runCodex()`, quote the path for spaces — and improves on it in two ways my PR didn't:

1. **`codexAuthStatus()` distinguishes `probe-failed` from `signed-out`**, and the catalog build now *refuses to run* on a probe failure instead of silently emptying the picker. That's the real safeguard — my version kept the bare `catch` that let the original bug hide. Better.
2. **Injectable `platform` parameter** makes the Windows path testable from any host, with 9 unit tests covering the shim preference and quoting — exactly the ESM-mocking wall I hit and (correctly, I think) chose not to restructure the module for on a bugfix.

Thanks for the careful review @duolahypercho — the two rounds of feedback (centralize, then quote for spaces) materially shaped the final result, even though you ended up landing it yourself. Appreciated.

Leaving a pointer for anyone who lands here: the fix is in `main` at `6580247`. No further action needed on this PR.

timm-u:
Superseded by 6580247 (now on main); closing in favor of the merged fix. See comment above.

PR #49, opened 2026-08-05
fix(macOS): preserve tray bundle path from build script

leetalley-ambientmuse:
# Fix macOS tray bundle path capture

## What changed

- Send SwiftPM build progress to stderr so the build script emits only the bundle path on stdout.
- Keep `bin/model-router-tray` able to capture the path and open the generated app.

## Why

`bin/model-router-tray` captures the build script output in a shell variable. SwiftPM writes progress to stdout, so the captured value included build logs and `open` received an invalid path. The build completed successfully, but the launcher failed to open the app.

## Validation

- `npm run check`
- `npm test` (237 passed, 0 failed, 2 skipped)
- Built and launched the native macOS menu-bar app from the fixed launcher path.

### DIFF
--- a/scripts/build-macos-tray-app.sh
+++ b/scripts/build-macos-tray-app.sh
@@ -7,7 +7,10 @@ bundle_dir=${1:-"$repo_dir/dist/Model Router.app"}
 configuration=${MODEL_ROUTER_TRAY_CONFIGURATION:-release}
 binary_dir="$tray_dir/.build/$configuration"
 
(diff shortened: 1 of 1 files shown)


duolahypercho:
Closing as superseded: the identical fix landed on main in 6752eee (`fix(tray): keep swift build progress out of the bundle-path stdout`) — SwiftPM progress now goes to stderr and stdout carries only the bundle path, exactly as this PR proposed. Thanks for diagnosing it; the launcher path capture works on current main.

PR #52, opened 2026-08-05
Lock LiteLLM dependencies

Imisibamiro:
## Summary

- replace the unavailable LiteLLM pin with the currently published `litellm[proxy]==1.83.9`
- add a universal, hash-verified Python lock and install it with `--require-hashes`
- cover both POSIX and Windows installers and add a regression test

## Root cause

The installers requested `litellm[proxy]==1.93.0`, which is not available from PyPI, and did not lock transitive Python dependencies.

## Validation

- `uv pip sync --require-hashes --strict --python .venv/bin/python requirements/litellm.lock`
- `pip install --require-hashes -r requirements/litellm.lock`
- `MODEL_ROUTER_LITELLM_INTEGRATION=1 node --test --test-name-pattern='Anthropic Messages through the real LiteLLM adapter' test/anthropic-api-integration.test.mjs`
- `npm run check`
- `npm test`
- `npm audit --omit=dev`


duolahypercho:
Thanks for the PR — the hash-verified lock infrastructure (`requirements/litellm.lock` + `--require-hashes` installers + the regression test) is a genuinely nice hardening pattern, and you clearly hit a real install problem. That said, we're not going to merge this as-is. A few reasons:

**The stated root cause is inaccurate.** `litellm==1.93.0` *is* published on PyPI and is not yanked. What's actually true is that 1.93.0 only ships `manylinux` wheels plus an sdist — no macOS or Windows wheels — so on those platforms pip falls back to a source build, which fails without a build toolchain. That's presumably the failure you hit (Windows?). We verified a clean `uv pip install 'litellm[proxy]==1.93.0'` succeeds on macOS.

**Downgrading is the wrong direction.** Pinning 1.83.9 rolls the gateway back ~4 months and ~10 minor versions on the component that processes prompts, credentials, and tool calls. litellm 1.94.1 and 1.95.0 both ship macOS and Windows wheels, so bumping forward solves the same platform problem without reverting any fixes.

**Unrelated changes are bundled in.** The first commit switches DeepSeek V4 Flash/Pro to the Responses API and introduces a per-model `protocol` override across the forwarder, registry, and LiteLLM config. That's a behavioral routing change that needs its own PR with evidence that DeepSeek's API actually serves `/responses`.

If you'd like to resubmit: regenerate the lock against 1.95.0 (or latest stable with full platform wheel coverage), correct the root-cause description, and split the DeepSeek change into a separate PR. The lock-file approach itself would be welcome.

duolahypercho:
Closing this as stale, but the security half of it should come back.

**Why it can't merge as-is:** the premise has been overtaken. This pins `litellm[proxy]==1.83.9` because `1.93.0` was unavailable from PyPI at the time. Main now pins `1.95.0` in all four install paths (`bin/install:90`, `bin/install:100`, `src/install-plan.mjs:23`, `install.ps1:196`, `install.ps1:215`). Merging this would be a **downgrade** of roughly twelve minor versions, not a fix. The branch is also `CONFLICTING` against current main after a lot of movement.

**Why the other half still matters:** the hash-verified lock is genuinely missing from main. There is no `requirements/` directory and no `--require-hashes` anywhere in the repo — I checked. So the transitive Python dependency surface behind `litellm[proxy]` is still unpinned and unverified on every install, which is a real supply-chain gap and exactly what this PR set out to close.

If you want to redo it, the valuable part is the lock generation and `--require-hashes` install wired into both the POSIX and Windows installers, regenerated against `litellm[proxy]==1.95.0` + `fastapi==0.139.2` rather than the older pin, with the regression test you already wrote. That would land cleanly and would not touch the version at all.

Thanks for the work — the versioning drifted out from under it, not the idea.

PR #66, opened 2026-08-06
fix(win): recognize TOML-escaped Windows catalog paths

michaelyosta:
## Problem

On Windows, `doctor` reports a false conflict:

```
FAIL  Router ownership: unknown catalog: C:\\Users\\<user>\\.codex\\codex-router\\merged-models.json
```

Root cause: `config-manager.mjs` wrote `model_catalog_json` with `JSON.stringify()`, which escapes every backslash (`"C:\\Users\\…"`). `detectLegacyInstallations()` compares **raw config text** against `path.join()` output (`C:\Users\…`) with no TOML unescaping, so the router's own catalog never matches and is reported as an unknown, conflicting one. This also makes migration tooling refuse to run (`applyKnownMigrations` throws on `unknownConflict`). POSIX paths contain no backslashes, so the bug only shows on Windows.

## Fix

Two-sided, so both existing installs and fresh ones are covered:

1. **Writer** (`src/config-manager.mjs`): values containing backslashes are emitted as TOML **literal strings** (single quotes) — readable, and raw-text comparisons see the real path.
2. **Detector** (`src/legacy-migration.mjs`): catalog paths are compared after TOML unescaping and, on Windows, after case/separator folding.

## Verification

- New regression test in `test/legacy-migration.test.mjs`: both quoted forms (escaped basic string and literal string) are recognized as the router's own catalog, `unknownConflict === false`.
- New regression test in `test/config-manager.test.mjs`: the emitted `model_catalog_json` round-trips to the expected path without backslash escaping.
- Confirmed live on Windows: the `unknown catalog` FAIL disappears; `doctor` reports `no conflicting router detected`.


duolahypercho:
Review note: `tomlValue()` emits single-quoted TOML literal strings for backslash paths, but TOML literal strings do not support `\x27\x27` escaping. A Windows path containing an apostrophe (for example a username like `O\x27Brien`) would produce invalid TOML that Codex cannot parse. The `tomlUnescape()` reader already handles basic strings, so consider emitting a normal basic string (JSON escaping) and relying on the unescape path, or rejecting apostrophes with a clear registry/config error. A regression test with an apostrophe in the path would make either choice explicit.

duolahypercho:
Maintainer fix applied on top of the original patch: `tomlValue()` now always emits a TOML basic string (`JSON.stringify`) instead of a literal string, so Windows paths containing apostrophes remain valid TOML. The legacy-migration `tomlUnescape()` path already handles the escaped basic strings. Added a regression test covering apostrophe + backslash paths. `npm run check` and the full suite pass (283 passed, 4 skipped).

PR #74, opened 2026-08-06
fix(codex): keep native catalog and root concurrency compatibility

rajivpoddar:
Preserve two local compatibility fixes on top of the upstream catalog/config refactor:

- Native models missing `supports_reasoning_summaries` now normalize to `false`, so older or incomplete native captures do not leak undefined capability values into the merged catalog.
- The managed concurrency scalar stays at the root of `config.toml` before any `[agents]` table, avoiding an `[agents]` block that breaks companion configs on the current Codex build.

Tests: `npm test` passes with 261 passed, 2 skipped, 0 failed. `./bin/install` also completes with this working tree.


rajivpoddar:
@duolahypercho this PR preserves the local native-catalog and root-level concurrency compatibility fixes on top of the latest main. Could you take a look when you get a chance?

PR #76, opened 2026-08-06
feat(gemini): add thought signature handling and strip client metadata

aselekoglu:
### Summary
- Added `ensureGeminiThoughtSignatures` in `api-forwarder` to ensure `thought_signature` is populated on assistant tool calls for Gemini models.
- Stripped `client_metadata` and `web_search_options` from forwarded payload bodies across forwarders/router.
- Updated registry and routing test suites for Gemini models and metadata isolation.

### Testing
- `npm run check` passed.
- `npm test` passed (260/260 tests).


aselekoglu:
### Summary of Changes

Cleaned up and rebased all Mac & Windows Gemini integration updates into 2 clean commits:

- **Gemini Support & Sanitization:** Added Gemini model registry definitions (`config/providers.json`), `skip_thought_signature_validator` injection for Gemini tool calls, and consistent `client_metadata` stripping (`api-forwarder.mjs` & `router.mjs`).
- **Windows Compatibility:** Fixed PowerShell variable expansion via `-EncodedCommand` in `provider-key.mjs`, added non-admin Startup folder fallback for background services (`service-windows.mjs`), normalized TOML backslash paths, and set `PYTHONIOENCODING=utf-8` for LiteLLM.
- **Tests & Verification:** All 32 unit/routing tests passing (`npm test`). Verified end-to-end on macOS & Windows with `gemini-3.6-flash`.

duolahypercho:
Review notes after a line-by-line pass:

1. Rebase will decide the legacy-migration overlap. PR #66 already adds TOML unescaping plus case/separator folding on Windows; drop the `path.normalize()` approach here to avoid two competing detectors.
2. `delete routed.client_metadata;` is a no-op against the diff as written (`routed` is built from only `model` and `input`). If the intent is to strip it, spread the incoming payload first (current main does this) so the delete actually has an effect.
3. `delete payload.client_metadata;` appears twice in `normalizeBody()`; keep one.
4. The sentinel `skip_thought_signature_validator` is invented-looking. Please link the official Gemini contract that accepts it, and keep a live-gateway regression in the repo so the value cannot silently break when upstream changes.
5. Adding `MODEL_BY_SLUG` as a fallback lets callers use full slugs as API model IDs. Confirm that is intentional and add a test documenting the surface.
6. Several Windows changes (hidden prompt, scheduled task, UTF-8) overlap recently merged Windows PRs; after rebasing, keep only the deltas that are not already on main.

duolahypercho:
Thanks for this — the diagnosis was solid, and the substance of it has landed on `main` in 9bb1deb and 7848929 with you credited as co-author on both.

Closing rather than merging because the branch predates the provider-config split (c3847e7), which deleted `config/providers.json`. Rebasing would have meant re-authoring the model definitions into the new per-file layout anyway, and there were a couple of changes worth making to the runtime pieces along the way. Details below so nothing is lost.

### Landed as-is

- **Stripping `client_metadata` from forwarded payloads.** The most valuable part of this PR, and not Gemini-specific — `main` had no handling for it anywhere. Applied to the API forwarder, the Kimi OAuth forwarder, and the routed `/responses` path as you had it, plus `summarize()` in `router.mjs`, which also spreads the caller payload into the gateway request and was leaking the field on the compaction path.
- **Dropping `web_search_options` for Gemini.**
- **`-EncodedCommand` for the Windows hidden key prompt.**
- **The `legacy-migration.mjs` regex escape.** Worth flagging how much this one mattered. The pattern is built in a template literal, so `\s` collapsed to a literal `s` and the compiled regex was `^s*key s*=s*...` — it matched only unindented keys with no spacing around `=`. So `model_catalog_json = "..."`, the conventional TOML spacing, was invisible to `detectLegacyInstallations()`. That left `unknownConflict` false for a foreign router's catalog, which is exactly what clears migration to run over an installation the router does not own. Good catch; it has a test now covering five spacing variants.

### Changed

**`ensureGeminiThoughtSignatures` set the sentinel unconditionally.** A real signature Gemini had already returned got overwritten with `skip_thought_signature_validator`, so the model lost the reasoning it was being asked to continue from. It now fills only where a signature is absent, with a regression test that fails against the unconditional version.

**Provider detection collapsed into one `isGeminiProvider()` predicate.** Dropped the `id.includes("gemini")` substring check — it would also match a non-Google provider that happened to have "gemini" in its id.

**The PowerShell candidate loop now keeps the first non-ENOENT failure rather than the last error.** That is what produced the `spawnSync pwsh.exe ENOENT` in the reported trace: `powershell.exe` had already failed for some real reason, then `pwsh.exe` — absent on a stock Windows box — overwrote it. The underlying error was never visible to anyone.

### Not landed

The curated Gemini model definitions. `gemini-api` is a catalog-only provider by design (see `AGENTS.md`): it ships no preselected models, and users curate with `bin/curate-models gemini-api` after storing a key. Preloading them reverses that policy, so it deserves its own discussion rather than riding along with a bug fix.

### One ask

I do not have a Windows machine, so the `-EncodedCommand` change is unverified — it went in on your evidence, not mine. If you can re-run the installer against current `main` and paste what the key prompt does now, that would close it out. If it still fails, the error you see should finally be the real one instead of the pwsh ENOENT masking it.

aselekoglu:
Thanks @duolahypercho! Glad to hear the diagnosis and changes landed on `main` (`9bb1deb` & `7848929`).

I tested the current `main` branch (`7848929`) directly on my Windows machine:

1. **`-EncodedCommand` Execution**: `windowsHiddenPromptArgs()` generates the base64 UTF-16LE payload cleanly. Running `./bin/model-router.ps1 codex provider-key <provider> set` launches `powershell.exe` without any command-line quoting or parsing errors, and the hidden stdin key prompt captures input correctly.
2. **Fallback Error Handling**: The `powerShellStartupError()` candidate loop works as expected — when `pwsh.exe` is absent on a stock Windows box, it no longer masks the underlying prompt status under a missing-binary ENOENT.
3. **Regex Spacing Fix**: Verified the `legacy-migration.mjs` regex fix against spaced TOML keys (`model_catalog_json = "..."`); legacy detection now correctly spots foreign catalogs regardless of whitespace around `=`.

Everything is working smoothly on real Windows hardware now! Thanks again for crediting co-authorship.

duolahypercho:
Following up: the Windows key prompt is confirmed working on `main`, so this is fully closed. Thanks for reporting it and for pinning down `-EncodedCommand` — I had no Windows machine and would not have found it.

Two things your report turned up that were larger than the original bug:

**The failure was self-perpetuating.** The installer records the current revision, fast-forwards the checkout, runs setup, and restores the old revision on any non-zero exit. So the fix for a setup-path bug was fetched and then discarded by the rollback, and every retry reproduced the same failure. Setup now exits 2 for "the checkout is healthy, configuration is unfinished" — a declined prompt, a missing credential, a bad argument — and both installers keep the update. Any other non-zero exit still rolls back. Credential entry is also best-effort in guided runs now, so a broken key prompt warns and finishes the install instead of taking it down.

**The `\s` escape was a safety bug, not a cosmetic one.** Built in a template literal, the pattern compiled to `^s*key s*=s*...`, matching only unindented keys with no spacing around `=`. So `model_catalog_json = "..."` — the conventional TOML form — was invisible to `detectLegacyInstallations()`, which left `unknownConflict` false for a foreign router's catalog and cleared migration to run over an installation the router does not own. Covered by a test across five spacing variants.

**On the Gemini models.** Preloading them is still not something we want — `gemini-api` is catalog-only by design and users curate with `bin/curate-models gemini-api`. But you were solving a real problem: storing a key printed "The provider is enabled" and told you to reopen the picker, which was empty, with nothing anywhere saying curation was the next step. That is now fixed — storing a key for a provider with no registered models names the exact command to run next. The discoverability gap you hit was real even though the fix landed differently.

PR #86, opened 2026-08-07
feat(config): per-provider config tree, plus error-classification and reasoning-effort fixes

duolahypercho:
## Summary

Two changes, verified together end-to-end against a live install:

1. **Split `config/providers.json` into per-provider directories.** The single 2.5k-line file becomes `config/<provider>/<provider>.json` for the provider entry plus one file per model under `config/<provider>/[<variant>/]<model>.json`. Registry, discovery, and curation read the tree; adding a model is now a new file rather than an edit inside a shared blob.

2. **Classify exhausted-plan errors correctly.** A live Kimi OAuth 403 reports an exhausted plan as `You've reached your usage limit for this billing cycle`, which the quota patterns missed because they only matched the reverse word order (`usage limit reached`). The 403 then fell through to the OAuth branch and told the user to sign in again — a login that cannot fix an exhausted plan, which is the exact misleading-advice failure the error-translation work exists to prevent.

## Testing

Full suite: **334 pass, 0 fail** (4 skipped are the pre-existing Windows guards).

Beyond unit tests, this was driven through Codex Desktop against the running router:

- **Catalog** serves all 39 models from the new per-provider tree; live catalog matches config for every routed model.
- **Errors** — opencode's Grok 4.5 upstream is genuinely down, and the UI now reads `Something is wrong at opencode: Grok 4.5 (opencode Go) is unavailable right now. Retry in a few minutes or switch models.` instead of LiteLLM's exception chain.
- **Kimi quota 403** re-tested live after the fix: now `billing_error` / "You have run out of usage at kimi… Top up or check the plan", no sign-in prompt.
- **Subagents** — Kimi K3 spawned one via `collaboration__spawn_agent`.
- **Images** — Kimi K3 correctly described a generated test image.
- **Background redirect** — a native `gpt-5.6-luna` request came back as `grok-oauth-grok-4-5` while the OpenAI quota was exhausted.
- **Effort levels** — all 68 models have a `defaultEffort` present in their own `reasoningLevels`; 13 of 13 declared levels completed real requests.

## Notes for reviewers

- The catalog does **not** regenerate on router restart. Picking up config changes needs `./bin/refresh-catalog` and a full Codex quit/reopen — worth deciding whether `update` should do this automatically.
- Probing *undeclared* effort levels showed the gateways **accept** them rather than returning 400, so they silently swallow or remap unsupported values. Advertising only documented levels stays the right posture; acceptance is not evidence of support.
- `ollama-cloud` declares only `['high']` for `deepseek-v4-pro` and `glm-5.2` where every other provider of the same upstream offers `['high','max']`. Possibly correct, but untested here — no Ollama Cloud credentials on this machine.


duolahypercho:
Pushed two follow-up commits resolving the effort-level inconsistencies flagged above.

**`ollama-cloud` was wrong, but not in the direction I assumed.** I researched Ollama's OpenAI-compatible surface: `reasoning_effort` was documented in v0.18.0 (ollama/ollama#14820, ollama/ollama#14821, merged March 2026). The source in `openai/openai.go` validates it against `high/medium/low/max/none`, **rejects anything else with an error**, and passes the graded value through as a string rather than collapsing it to a boolean:

```go
if !slices.Contains([]string{"high", "medium", "low", "max", "none"}, effort) {
    return nil, fmt.Errorf("invalid reasoning value: '%s'...", effort)
}
if effort == "none" { think = &api.ThinkValue{Value: false} } else { think = &api.ThinkValue{Value: effort} }
```

So the forwarder's "Ollama does not document reasoning_effort" comment was stale, and it was silently discarding every effort selection — Ollama Cloud turns always ran at the model default. Fixed by forwarding the value through a mapping into Ollama's accepted set (Codex rungs like `xhigh` and `minimal` would 400 if passed verbatim), and widening DeepSeek V4 Pro and GLM-5.2 to the `high/max` ladder their other resellers already offer. Kimi K2.7 Code and MiniMax M3 keep a single tier, which already matched their peers.

**`qwen-plan/glm-5.2` defaulted to `high`** where Z.ai, opencode-go, and zai-coding all use `max`. Because the qwen-plan profile always sends `reasoning_effort` explicitly, that default was what reached DashScope — anyone picking GLM-5.2 on Qwen Plan silently got the weaker tier the same model serves at max everywhere else.

Two regression tests lock both classes of bug: resellers of one upstream model must agree on `defaultEffort` when their ladders match, and Ollama Cloud models may only advertise levels Ollama accepts. A re-audit now reports no cross-provider effort inconsistencies anywhere in the registry. Suite: 336 pass, 0 fail.

One caveat: no Ollama Cloud credentials on this machine, so the forwarding change is covered by the subprocess forwarder test rather than a live call.

PR #89, opened 2026-08-07
fix(gemini-api): strip unsupported thinking parameters and ensure UTF-8 encoding on Windows

aselekoglu:
## Problem Summary
When attempting to use `gemini-3.1-pro` with Codex Router, the request failed with:
`{"detail":"The 'gemini-api/models/gemini-3.1-pro' model is not supported when using Codex with a ChatGPT account."}`

### Root Causes
1. **Gemini API Parameter Incompatibility**: Google's OpenAI-compatible endpoint (`/v1beta/openai/chat/completions`) throws an HTTP 400 (`INVALID_ARGUMENT: Unknown name "thinking": Cannot find field`) if unsupported parameters like `thinking` or `think` are present in the payload. When `api-forwarder` returned a 400, `router.mjs` attempted fallback to native ChatGPT (`chatgpt.com`), which triggered ChatGPT's 401 error.
2. **LiteLLM Config Reload Mismatch**: Curating new user models writes `user-models.json` and updates `litellm.yaml`, but the running background LiteLLM instance on port 4100 needed a clean service restart to pick up the updated model routes in memory.
3. **Windows UTF-8 Encoding Crash**: Starting LiteLLM on Windows systems without `PYTHONIOENCODING=utf-8` could cause a `UnicodeEncodeError` when printing startup banners to non-UTF-8 default code pages (e.g. `cp1252`).

---

### `src/api-forwarder.mjs`
- Added `delete payload.thinking` and `delete payload.think` (alongside `frequency_penalty`, `presence_penalty`, `store`, `seed`, and `logit_bias`) inside `isGeminiProvider(provider)` cleanup logic.

### `src/start.mjs`
- Added `PYTHONIOENCODING: "utf-8"` and `PYTHONUTF8: "1"` to `commonEnv` for child processes so LiteLLM starts reliably on Windows terminals.

### `test/routing.test.mjs`
- Added regression test assertions ensuring `thinking` and `think` parameters are cleanly stripped for Gemini requests.

---

## Verification

- **Unit Tests**: Executed `node --test test/routing.test.mjs` -> **32/32 tests passed (0 failures)**.
- **System Doctor**: Executed `bin/model-router codex doctor` -> All checks **OK** (`Background service: running`, `Router health: version 0.4.0-beta.2`).
- **Live Endpoint Verification**: Tested all 5 Gemini models (`gemini-3.6-flash`, `gemini-3.5-flash`, `gemini-3.1-pro`, `gemini-2.5-pro`, `gemini-2.5-flash`) via `http://127.0.0.1:4102/_codex-router/.../v1/responses` -> All returned **200 OK**.

### DIFF
--- a/src/api-forwarder.mjs
+++ b/src/api-forwarder.mjs
@@ -202,6 +202,36 @@ function isGeminiProvider(provider) {
   return provider?.id === "gemini-api" || provider?.ownedBy === "google";
 }
 
+// Google Gemini OpenAI-compatible endpoints reject non-user content parts of
+// type image_url/input_image ("Invalid content part type: image_url").
+// Convert image parts in assistant and tool history turns to text placeholders.
+function sanitizeGeminiImageContent(messages) {
+  return messages.map((message) => {
+    if (!message || message.role === "user" || !Array.isArray(message.content)) return message;
+    let changed = false;
+    const newContent = message.content.map((part) => {
+      if (!part || typeof part !== "object") return part;
+      if (part.type === "image_url" || part.type === "input_image") {
+        changed = true;
+        const urlStr =
+          typeof part.image_url === "string"
+            ? part.image_url
+            : typeof part.image_url?.url === "string"
+              ? part.image_url.url
+              : "";
+        const label = urlStr.startsWith("data:")
+          ? "[Image]"
+          : urlStr
+            ? `[Image: ${urlStr}]`
+            : "[Image]";
+        return { type: "text", text: label };
+      }
+      return part;
+    });
+    return changed ? { ...message, content: newContent } : message;
+  });
+}
+
 // Gemini 3.x thinking models reject assistant tool calls whose reasoning
 // signature is missing, which is the normal state after compaction or after a
 // history translated from another provider. The sentinel below is the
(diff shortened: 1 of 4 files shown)


duolahypercho:
Thanks for this, @aselekoglu — solid diagnosis. We've landed the fixes across two PRs on `main`:

- **#91** — strips `thinking`/`think` for Gemini at the forwarding boundary (the root-cause 400 that was surfacing as the misleading "not supported with a ChatGPT account" fallback).
- **#92** — sanitizes non-user-turn `image_url` parts for Gemini, strips the OpenAI-only `store`/`logit_bias` fields, adds `PYTHONIOENCODING`/`PYTHONUTF8` for LiteLLM on non-UTF-8 Windows code pages, and guards non-elevated Windows `install`.

Two deliberate narrowings from your version: we kept `frequency_penalty`/`presence_penalty`/`seed` (Google's OpenAI-compatible surface accepts them, so dropping them would change sampling), and we held the scheduled-task argument rewrite (dropping `/D` + the quote-form change) since the current doubled-quote form is Microsoft's documented form for paths with spaces and we couldn't verify the rewrite on a Windows host — happy to revisit that one separately if you can confirm the failure it fixes.

Closing in favor of #91 + #92. Appreciate the thorough writeup and verification table.

PR #104, opened 2026-08-08
Let the signed-in ChatGPT plan read images for text-only models

sergeylopukhov:
# Let the signed-in ChatGPT plan read images for text-only models

## What this changes

The vision bridge can now use a vision model from the operator's own ChatGPT
plan as the image reader, alongside the routed engines it already supports.
The tray lists those models in their own section, and each one opens a submenu
of the reasoning efforts it declares, so picking the reader and how hard it
reads is one gesture.

## Why

Today the bridge only resolves an engine from the routed, credentialed set. An
install whose external providers are all text-only (DeepSeek, GLM, Kimi) has no
engine at all, so the operator has three options: pay a second provider for
vision they do not otherwise want, download a local model, or switch sessions
by hand every time a screenshot matters. Meanwhile a vision model they already
pay for is sitting one process away, and native turns already reach it.

## How it works

- `nativeVisionCandidates()` builds engines from the native capture, which is
  snake_case where the registry is camelCase. Only models that are `list`
  visible, declare image input, survived the auth gate into the merged catalog,
  and are not hidden in the picker become candidates. A signed-out or
  login-free install has none.
- The request for a native engine goes to the Codex backend with the caller's
  own session headers, the same ones the router already forwards for native
  turns. There is no new credential, no new install step, and nothing to
  download.
- That backend rejects a buffered Responses call with
  `{"detail":"Stream must be set to true"}`, so this path sends `stream: true`,
  asks for `text/event-stream`, and reassembles the transcript from the deltas.
  It fails closed when no caller session is present rather than falling back to
  the gateway, which holds no credential for a native slug.
- The pinned effort is stored next to the pinned engine and clamped to what the
  engine itself declares, so a level chosen for one model is dropped rather
  than sent when another one ends up doing the reading. Transcripts are cached
  per engine and effort, so raising the level re-reads the image instead of
  replaying the cheaper pass. Leaving the effort unset keeps the request
  byte-identical to what it was.

## On the "no second upstream path" rule

`AGENTS.md` says a registry engine's call must go through the same gateway,
credential, and request profile as any other routed turn, with the local engine
as the only sanctioned exception. The intent I read there is that a hosted
engine must not bring a new credential, a separate vision API key, or an
external dependency, and none of that applies here: this reuses the native path
the router already owns, with the session Codex itself supplies. Routed engines
keep going through the gateway exactly as before. If you would rather that rule
stay literal, this needs a line in `AGENTS.md` before it lands, and I am happy
to write it whichever way you prefer.

## Tests

`test/vision-bridge.test.mjs`, `test/vision-bridge-state.test.mjs` and the
bridged-catalog case in `test/catalog.test.mjs`, as the vision-bridge rules ask
for. They cover the native engine shape, candidate filtering by visibility and
hidden state, pinning against auto-ranking, the streamed call to the Codex
backend carrying the caller session and not the gateway key, failing closed
without a session, SSE reassembly, effort clamping in both directions, and the
advertisement appearing and disappearing with a native engine.

`npm run check` and `npm test` pass: 497 tests, 491 passed, 0 failed.

## Verified

End to end on macOS with a text-only routed model reading an invoice through a
plan model, with the answers checked against the image by hand.

---

Sent as three independent branches so you can take them separately. They touch different code and merge cleanly in any order.


duolahypercho:
Reviewed in depth. The security architecture is sound — I want to say that first, because the rest of this is a list of problems and it would give the wrong impression otherwise.

**What holds up.** Credential isolation is clean: `nativeHeaders()` copies only the fixed `FORWARD_HEADERS` allowlist and is only ever sent to the hardcoded `NATIVE_BASE`, never to a user-controlled destination. The fail-closed path is real — the `engine.native` branch in `describeRequest()` throws and returns before the gateway and local branches, so there is no path where a missing session falls back to the gateway with a native slug. No credential, header, or transcript reaches a log; upstream error bodies are never echoed. Effort clamping is correct in both directions and properly tested. 497 tests, 492 pass, 0 fail.

Three things to fix, then the policy question.

---

**1. Mid-stream errors are silently swallowed and returned as a complete transcript.** This is the one I'd most want fixed.

`streamedResponseText()` recognizes only `response.output_text.delta` and "anything with a `.response` field". It has no concept of stream failure. Reproduced directly:

- Deltas followed by `{"type":"error","error":{"message":"upstream overloaded"}}` → returns the partial text, error dropped.
- Deltas followed by a proper `{"type":"response.failed","response":{"status":"failed",...}}` → `deltas.join()` is non-empty, so the failed status is never inspected at all.

A backend error partway through generation produces a plausible-looking truncated transcript with zero indication anything went wrong. That cuts against the vision-bridge's own "evidence, not impressions" rule in AGENTS.md — a downstream model quotes it as if it were the whole image. The existing `unreadableBlock()` degradation path is the correct behavior here and it never fires. Suggest treating a `response.failed`/`error` event, or a completed `response` whose `status !== "completed"`, as a thrown error.

**2. Request-time resolution skips the auth gate that the other two call sites apply.** The candidate criteria are enforced in three independently-maintained places, and they disagree:

- `src/catalog.mjs:531` gates on `openaiAuthenticated && !loginFree`.
- `src/control.mjs` `shippedNativeVisionEngines()` gates via merged-catalog membership — equivalent.
- `src/router.mjs:726` (`bridgeVisionInput`, the actual request-time path) reads `NATIVE_CATALOG_PATH` directly with **no auth check**.

That file can legitimately be stale — `nativeCatalog()` explicitly reuses a cached on-disk capture when a fresh probe fails, and nothing forces a refresh before the router reads it. So: operator signs out, `catalog.mjs` correctly stops advertising the engine, but the pinned setting still points at it, and every bridged paste on every text-only routed model attempts a native call against the stale capture. It fails, and there is no fallback to another candidate within that resolution — the image degrades to "unreadable" rather than quietly using a still-valid paid or local engine. Best fix is one shared helper rather than three hand-maintained candidate builders.

**3. Evidence cache isn't partitioned by account.** `evidenceCache` keys on `slug + effort + url` with no session identifier. That was fine when the paying credential was always the router's own install-scoped one. Now that a native call is authorized by the caller's live session, a cache hit skips any re-check that the current session is entitled to that model. Not a confidentiality leak — you need the identical image bytes to get a hit — but it is an entitlement bypass on the hit path. Keying native engines on `chatgpt-account-id` from `FORWARD_HEADERS` would close it. Lower severity than the two above.

---

**On the AGENTS.md question you raised.** You were right to flag it rather than merge quietly. My read is that this violates the rule as written. The rule names exactly one exception:

> A registry engine's call goes through the same gateway, credential, and request profile as any other routed turn. Do not add a second upstream path, a separate vision API key, or an external CLI dependency for a hosted engine. The one sanctioned exception is the local engine (`vision-bridge local`)...

Your code comment says it too — "Two call shapes" becomes "Three call shapes," and the third is a hosted engine that bypasses the gateway for `NATIVE_BASE` directly.

Your intent argument isn't frivolous: no stored credential, no CLI, no new install step. But two things cut against it. The rule enumerates a single named exception whose justification ("no credential to store") is structurally the same as yours — if a general "no new credential" test were intended, that's what would have been written. And the native path skips the scaffolding the provider checklist requires elsewhere in the same file: no `doctor` reachability check, and no usage/limits surfacing, which matters concretely here because ChatGPT-plan quota spent transcribing images becomes invisible spend.

So this needs a maintainer ruling and most likely an explicit AGENTS.md amendment before it lands, which you offered to write. I've flagged it for a decision.

**Merge order:** not a concern. I test-merged this against #103 and #105 in every order — all clean, 499 tests, 494 pass, 0 fail on the fully merged tree. #103 is already on main and this still merges cleanly.

PR #105, opened 2026-08-08
Fix OpenRouter tool calling and compaction

jepgambardella:
OpenRouter's Qwen 3.8 Max and Muse Spark 1.2 support function calling, but their providers reject `tool_choice: required` and only accept `auto`. The models still return the requested function call with `auto`, so the compatibility check now matches the provider behavior.

This also removes `tool_choice: none` from compaction requests with no tools. xAI/Grok rejects that invalid combination, while the same request succeeds when the field is omitted.

Tested with OpenRouter: Qwen 3.8 Max, Muse Spark 1.2, and Grok 4.5.

### DIFF
--- a/src/compatibility-test.mjs
+++ b/src/compatibility-test.mjs
@@ -50,7 +50,9 @@ async function toolCall(model) {
         strict: true,
       },
     ],
-    tool_choice: "required",
+    // OpenRouter providers for Qwen 3.8 Max and Muse Spark 1.2 reject
+    // `required` while still supporting function calls through `auto`.
+    tool_choice: "auto",
   });
   const call = (payload?.output || []).find(
     (item) => item?.type === "function_call" && item?.name === "codex_router_probe",
--- a/src/router.mjs
+++ b/src/router.mjs
@@ -851,7 +851,6 @@ async function summarize(payload, route, signal) {
     model: route.gatewayModel,
     stream: false,
     tools: [],
-    tool_choice: "none",
     input: [...bridged, messageItem(COMPACT_PROMPT)],
   };
   delete body.previous_response_id;


duolahypercho:
Thanks for this — the compaction half is a real bug and a good catch. I've split the two changes because they landed on opposite sides of a review.

**Compaction fix: taken.** It's in #108 with credit to you. Your diagnosis held up under a full audit — `tool_choice` was redundant on every path, not just Grok's: `src/grok-oauth-forwarder.mjs:190` only forwards the field when `tools.length` is non-zero, and no request profile in `src/api-forwarder.mjs` reads it at all. An empty tool list already disables tool use everywhere, so `"none"` only ever contributed a value one provider refuses.

**Probe change: not taking it as written.** Two reasons.

First, `tool_choice: "auto"` lets a model legitimately decline to call the tool. The probe exists to *prove* tool calling works, so under `auto` a compliant model that simply answers in prose fails with "function call missing" even though its tool support is fine. That converts a deterministic check into a flaky one for every provider, to accommodate two models.

Second — and this is the part that decided it — Qwen 3.8 Max and Muse Spark 1.2 aren't in the shipped registry. `config/openrouter/openrouter.json` defines the provider but ships zero model entries, and `src/provider-onboarding.mjs:395` makes OpenRouter catalog-only by design. Those two entries exist in your local `user-models.json` from `bin/curate-models`. So the change trades a global guarantee for a local curation issue.

**The shape that would work:** a per-profile downgrade, exactly like the `qwen-plan` guard at `src/api-forwarder.mjs:370`. The router keeps sending `required`, and the forwarder downgrades it for the affected profile only — which is already how the identical DashScope restriction is handled, with test coverage at `test/routing.test.mjs:1810`. That needs a new profile string (there's no existing one to reuse, since these are the first OpenRouter models curated) plus setting `requestProfile` on your two curated entries.

Don't reuse `qwen-plan` for the Qwen entry, incidentally — it also drives DashScope-specific `reasoning_effort` collapsing that doesn't apply to OpenRouter's parameter surface.

Happy to take that as a follow-up here if you want to repurpose this PR, or I can open it separately. Also worth knowing: #107 fixes the same class of bug on DeepSeek thinking models, which may be why `required` looked broadly unsafe from where you were sitting.

duolahypercho:
Closing this — both halves are now on main, and the credit for both is yours.

**The compaction fix shipped as #108.** Your diagnosis held up under a full audit: `tool_choice` was redundant on every path, not just Grok's. `src/grok-oauth-forwarder.mjs:190` only forwards the field when `tools.length` is non-zero, and no request profile in `src/api-forwarder.mjs` reads it at all — so an empty tool list already disabled tool use everywhere, and `"none"` only ever contributed a value one provider refuses.

**The forced-`tool_choice` problem shipped as #118**, built per-profile rather than by changing the probe. Two things came out of investigating it that are worth having on the record.

First, I went in expecting an OpenRouter-wide downgrade and the evidence says otherwise. OpenRouter's `/api/v1/models` listing treats `tool_choice` as a per-model `supported_parameters` entry, separate from `tools` and filterable on it — which could not be true fleet-wide. And their provider-routing docs say providers that don't support a parameter still receive the request and ignore it, unless `require_parameters: true`. So the 400 you hit is the *upstream model's* rejection travelling back through the proxy, not OpenRouter policy. It's the same thinking-mode restriction already handled for `qwen-plan` and `deepseek-thinking`, and it follows the model to whichever reseller fronts it.

Second, and this is why the global version couldn't ship: the forwarder rewrites **every** request, so downgrading provider-wide would also have downgraded the subagent payload relay's forced function object for GPT, Claude, Gemini and Grok routed via OpenRouter — models that honour `required` — letting them answer in prose instead of returning the handoff. That turns a flaky-probe annoyance into a routed-subagent regression.

The profile is named `auto-tool-choice` rather than `openrouter-*` precisely because the restriction rides the model: someone curating the same model on Together needs identical behaviour.

**To fix your two models**, once you've updated:

```sh
./bin/curate-models openrouter --models qwen/qwen3.8-max,<muse-spark-id> \
  --request-profile auto-tool-choice --apply
```

Curation never rewrites existing entries, so models you curated earlier need that run after deselecting them, or `"requestProfile": "auto-tool-choice"` added by hand in `user-models.json` followed by `./bin/install`. The interactive path asks per model and defaults to no.

Thanks for this one — you found a real bug, and the half I didn't take pointed straight at the mechanism that was actually missing.

PR #110, opened 2026-08-08
fix(win): start the background service without a console window

### LINKED ISSUE #98: Windows scheduled task opens a persistent console window at logon
### Description

On Windows, the background-service installer registers an InteractiveToken
scheduled task that executes `cmd.exe` directly:

`cmd.exe /D /C "<state-dir>\start-codex-router.cmd"`

The generated wrapper keeps `node src/start.mjs` running in the foreground.
As a result, a persistent console window is created at logon and remains open
for the lifetime of the router. Watchdog restarts can open the window again.

### Expected behavior

Codex Router should start and remain supervised in the background without
displaying a console window.

### Proposed solution

Generate a small VBS launcher beside the CMD wrapper and run the wrapper with:

`WScript.Shell.Run(command, 0, True)`

Register the scheduled task using:

`wscript.exe //B //NoLogo "<state-dir>\start-codex-router-hidden.vbs"`

The launcher should wait for the process and propagate its exit code so that
the existing Task Scheduler restart behavior remains intact.

The uninstall operation should remove both generated launchers.

### App target

Codex

### Platform

Windows

_No response_

duolahypercho:
Fixes #98.

> **Needs a Windows smoke test before merge.** This was developed on macOS. The generated-string logic is fully tested, but no `wscript.exe` or Task Scheduler was available to execute. The unverified list is at the bottom and item 2 is the one that matters.

## The problem

The background-service installer registers an InteractiveToken task running `cmd.exe /D /C "<state-dir>\start-codex-router.cmd"`, and the wrapper keeps `node src/start.mjs` in the foreground. A console window appears at logon and stays for the router's lifetime; watchdog restarts re-open it.

## The change

`src/service-windows.mjs` generates a second launcher beside the CMD wrapper and points the task at `wscript.exe //B //NoLogo "<state-dir>\start-codex-router-hidden.vbs"`.

Rendered here with a spaced, non-ASCII path:

```vbs
Option Explicit

Dim quote, shell, status
quote = Chr(34)
Set shell = CreateObject("WScript.Shell")
On Error Resume Next
status = shell.Run("cmd.exe /D /C " & quote & quote & "C:\Users\Ann Marie\AppData\Local\codex-роутер\start-codex-router.cmd" & quote & quote, 0, True)
If Err.Number <> 0 Then
  WScript.Quit 1
End If
On Error Goto 0
WScript.Quit status
```

**Exit-code propagation** is the part that must be right, since Task Scheduler's `RestartCount 999 / RestartInterval 1m` is what keeps the router alive. `Run(..., 0, True)` returns the child's code and it is re-raised via `WScript.Quit status`. A failure to launch at all (`Err.Number <> 0`) quits `1` — otherwise a WSH runtime error exits 0 and the restart behavior silently never fires again.

**The .vbs is written UTF-16LE with a BOM.** `wscript.exe` otherwise parses a script file using the system ANSI code page, which mangles a non-ASCII state dir. The .cmd wrapper's existing UTF-8 encoding is untouched — that is a separate latent issue and I did not widen scope for it.

**Path quoting** uses `Chr(34)` for the quotes cmd.exe needs rather than stacked `""""` sequences, so the only quoting in the generated source is the string literal around the path. A test asserts every generated line has an even quote count.

## Upgrade path

The task name is unchanged and `Register-ScheduledTask -Force` / `schtasks /F` replace the definition, so there is no duplicate task. But `-Force` does not stop a *running* instance, and `MultipleInstances IgnoreNew` would then drop the new hidden `/Run` — leaving the old console window alive until next logon. So `install` now calls `endTask()` first. That was extracted and reused in `uninstall` and `restart`.

## Quoting hardening

The action's `Execute`/`Argument` now travel through the environment (`CODEX_ROUTER_TASK_EXECUTE` / `CODEX_ROUTER_TASK_ARGUMENT`) instead of being embedded in the `-Command` string, so quotes around the launcher path never pass through `powershell.exe -Command` reparsing. The schtasks fallback uses a **single** quote pair — the doubled-quote form is a cmd.exe-specific quirk and would split a spaced path into two arguments under the standard parser wscript.exe uses.

## Tests

Three added to `test/service-render.test.mjs` via two new render-only commands (`render-launcher`, `render-task`) so the tests assert against real generated values rather than duplicated literals. Coverage: VBS content with a spaced state dir, the exit-code propagation lines, quote balance on every line, the task command shape, and install/reinstall/uninstall/double-uninstall file handling — the last skipped on `win32` so it can never create a real scheduled task.

`npm run check` and `npm test` pass on current main: 493 tests, 488 passed, 0 failed (up from 490/485).

## Not verified on macOS

1. That the console window is actually gone. Window-style-0 is reasoned from the API contract, not observed.
2. **That `WScript.Quit status` reaches Task Scheduler as a failure and re-triggers the restart policy.** Highest-value check: kill `node src/start.mjs`, confirm Last Run Result is non-zero and the task restarts within a minute.
3. That `wscript.exe` accepts the UTF-16LE-with-BOM file. WSH documents Unicode support and honors the BOM, but if this is wrong the router fails to start at all — worth one manual run.
4. The `schtasks /Create /TR` fallback string, which only runs when `Register-ScheduledTask` fails in restricted or non-elevated shells.
5. The `powershell.exe -Command` argv round-trip for the env-var action.
6. The upgrade sequence on a real prior install: old task running → `/End` → re-register → exactly one hidden instance.
7. A genuinely non-ASCII `%LOCALAPPDATA%` end to end.

Credit to @Haz4rdovisk for the report, which specified the VBS approach, the task registration, exit-code propagation, and the uninstall cleanup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@
   which starts the wrapper hidden and waits for it, re-raising the wrapper's
   exit code so Task Scheduler's restart-on-failure settings still see a crash as
   a crash. Reinstalling replaces the old task in place, and uninstalling removes
-  both generated launchers.
+  both generated launchers. Reinstalling and restarting now wait for the running
+  instance to actually exit before starting the new one, an install that cannot
+  register the task starts the router again rather than leaving the machine with
+  none, and stopping a service that was never installed is no longer an error.
 
 - **Text-only models can answer about a pasted image.** A model with no image
   input — DeepSeek, GLM, Kimi — used to refuse the paste outright. When the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- **Windows no longer opens a console window at logon.** The scheduled task ran
+  the CMD wrapper through `cmd.exe`, so a console window appeared at every logon
+  and stayed for the router's lifetime, reappearing on each watchdog restart.
+  The task now runs a generated VBS launcher under `wscript.exe //B //NoLogo`,
+  which starts the wrapper hidden and waits for it, re-raising the wrapper's
+  exit code so Task Scheduler's restart-on-failure settings still see a crash as
+  a crash. Reinstalling replaces the old task in place, and uninstalling removes
+  both generated launchers.
+
 - **Text-only models can answer about a pasted image.** A model with no image
   input — DeepSeek, GLM, Kimi — used to refuse the paste outright. When the
   vision bridge is on, a vision model you already have reads the image and
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -247,6 +247,11 @@ Get-ScheduledTask -TaskName "Codex Router"
 ./codex-router.ps1 doctor --fix
 ```
 
+The task runs `start-codex-router-hidden.vbs` from the state directory under
+`wscript.exe`, which starts `start-codex-router.cmd` without a console window,
+so a missing window is not a sign that the router is down. Read `router.log` in
+the same directory for its output.
+
 Keep the repository at the absolute path used during installation. Rerun setup
 from the new path if it was moved.
 
--- a/src/service-windows.mjs
+++ b/src/service-windows.mjs
@@ -182,11 +182,56 @@ function installTask() {
   }
 }
 
+// `schtasks /End` returns once Task Scheduler has accepted the request, not
+// once the instance is gone, and `MultipleInstances IgnoreNew` silently drops a
+// `/Run` issued while the old one is still winding down -- which leaves the
+// router stopped until the next logon, and turns the installer's readiness wait
+// into a five-minute stall followed by a rollback. Polling the real state beats
+// retrying `/Run`: it continues as soon as the instance has actually gone
+// instead of guessing how long that takes, and it gives up on a fixed deadline
+// instead of hoping one extra attempt is enough.
+const TASK_STOP_TIMEOUT_MS = 10_000;
+const TASK_STOP_POLL_MS = 250;
+// Every state query has to return for the deadline above to mean anything, so a
+// wedged PowerShell is capped rather than allowed to hang the install outright.
+const TASK_STATE_TIMEOUT_MS = 15_000;
+
+function sleep(milliseconds) {
+  Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, milliseconds);
+}
+
+function waitForTaskToStop() {
+  const deadline = Date.now() + TASK_STOP_TIMEOUT_MS;
+  // An undefined state means no PowerShell could answer -- the same restricted
+  // shell that blocks registration -- so there is nothing to poll and waiting
+  // would only spend the deadline on a question that cannot be answered.
+  while (taskState() === "running") {
+    if (Date.now() >= deadline) return;
+    sleep(TASK_STOP_POLL_MS);
+  }
+}
+
 function endTask() {
   try {
     schtasks(["/End", "/TN", taskName], { quiet: true });
   } catch {
-    // The task may not exist, or may not be running.
+    // The task may not exist, or may not be running; either way there is no
+    // instance left to wait for.
+    return;
+  }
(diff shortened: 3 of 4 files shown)


duolahypercho:
Two updates: the highest-risk unverified item is now proven on CI, and a regression this PR introduced is fixed.

## The exit-code chain is executed now, not reasoned about

The thing actually blocking this was item 2 — whether `WScript.Quit status` propagates the wrapper's exit code. If it doesn't, Task Scheduler sees success on every crash and the restart policy silently dies, trading a console window for an availability bug.

That doesn't need a registered task to test. It needs `wscript.exe`, which `windows-latest` runners have. There's now a Windows-only test that generates the launcher through this PR's own `render-launcher` command, writes the exact BOM + UTF-16LE bytes `writeLaunchers()` produces, and executes it under both `cscript.exe` and `wscript.exe //B //NoLogo`:

- a wrapper exiting **42** must reach the host as **42**
- a wrapper exiting **0** must stay **0** — a clean shutdown reported as failure would have the task restart a router the user deliberately stopped
- a wrapper that **never ran** must not report success

It passed on CI:

```
✔ the generated Windows launcher returns the wrapper's exit code to its script host (1440.281ms)
```

**What that closes, precisely:** item 3 (wscript accepts the UTF-16LE-with-BOM file — a parse failure would be immediate red), and the half of item 2 that is a property of this code. It also incidentally proves the `Chr(34)` quoting survives a state dir containing a space.

**What it does not close, and I don't want to imply otherwise:** whether Task Scheduler *interprets* a non-zero action exit as a task failure and re-triggers `RestartCount`/`RestartInterval`. That is scheduler policy and needs a real registered task. Items 1 (console window actually hidden), 4 (`schtasks /Create /TR` fallback), 6 (upgrade against a prior install) and 7 (non-ASCII `%LOCALAPPDATA%`) also still need a real machine.

The `Err.Number <> 0` guard is likewise still unexecuted — it turns out not to be reachable on a normal host, since the Run command is hardcoded to `cmd.exe` and `CreateProcess` resolves that from the system directory regardless of PATH. It stays as belt-and-braces.

## A regression this PR introduced, now fixed

`endTask()` ran before `installTask()`, so in a restricted or non-elevated shell the registration throws, the catch swallows it, `/Run` is skipped — **and the previously-running instance had already been killed**. Pre-PR, a failed install left the old router running. That is strictly worse than what it replaced.

The order is kept, because `-Force` re-registration can orphan Task Scheduler's tracking of a live instance and make a later `/End` a no-op against a real process. The failure path recovers instead, and it queries before acting: by that point both registration routes have failed, so the surviving state is either the previous definition or nothing at all, and the code cannot know which. A bare `/Run` against a name that is gone recovers nothing and reports its own error over the one that matters.

Restoring the old definition was considered and rejected — nothing is snapshotted to restore *from*, and re-creating it means re-creating the console-visible `cmd.exe` action this launcher exists to remove.

Three smaller ones alongside it:

- `writeLaunchers()` moved inside the `try`, so a `renameSync` over a `.vbs` held open by a running `wscript.exe` can no longer throw out of `install` uncaught.
- The `/End` → `/Run` race under `MultipleInstances IgnoreNew` is now deterministic: `endTask()` polls `taskState()` until the instance actually stops, with a 10 s deadline checked every iteration, a 15 s timeout on the PowerShell calls so a wedged shell cannot hold it open, and an immediate exit when no PowerShell can answer at all.
- `stop` is idempotent like `restart` and `uninstall`, instead of throwing when nothing is installed.

## On testing the ordering off Windows

`schtasks.exe` and `powershell.exe` don't exist on macOS, so every scheduler call normally fails and the ordering between them is invisible — which is why this regression was invisible too. The new tests put executable stubs of those names first on `PATH`, which makes the call sequence observable and lets a chosen call fail on demand. That's the only way to reach the recovery path without a Windows box in a restricted shell. All five fail against the previous commit.

542 tests, 536 passed, 0 failed, 6 skipped.

## Still needs a human

Someone on Windows should confirm the console window is genuinely gone at logon, and that killing `node src/start.mjs` produces a non-zero Last Run Result with a restart inside a minute. The launcher now provably hands the right number to its host; whether Task Scheduler acts on it is the remaining unknown.

Issue #100, opened 2026-08-08
Native OpenAI models frequently reconnect and fail with local-router 503 or stream decode errors


yvos:
## Bug

When using native OpenAI models in Codex—especially GPT-5.6 Luna—the request frequently enters a reconnect loop and then fails.

### Observed errors

```
Reconnecting /5
unexpected status 503 Service Unavailable: upstream connect error or disconnect/reset before headers. reset reason: remote connection failure, url: http://127.0.0.1:4102/_codex-router/[REDACTED]/v1/responses

Reconnecting /5
stream disconnected before completion: Transport error: network error: error decoding response body
```

The capability segment of the local router URL has been redacted.

### Expected behavior

Native OpenAI model requests should complete normally, or fail once with a clear actionable error rather than repeatedly reconnecting.

### Reproduction

1. Install/enable Codex Router with the opencode Go/Zen provider.
2. Keep normal ChatGPT/OpenAI login enabled.
3. Select a native OpenAI model such as GPT-5.6 Luna.
4. Send a normal task.
5. Observe repeated reconnect attempts and, often, a 503 or stream-decoding failure.

The problem is reported as especially frequent on native OpenAI models; routed opencode models are not the focus of this report.

### Environment

- macOS arm64 (Darwin 25.5)
- Codex CLI 0.146.1
- Codex Router commit `9343a8b8e38be11c8f0924fc9f66b59fac5eba4a`
- ChatGPT/OpenAI login: active
- Enabled external provider: opencode Go/Zen
- No paid smoke test was run

At the time of filing, `bin/model-router codex status` reported the service as loaded/running but the router health endpoint was unavailable. This may be transient, but appears relevant to the 503.

Please investigate the local router's upstream connection lifecycle, health supervision/restart behavior, and SSE/response-body handling for native OpenAI traffic.

duolahypercho:
Root-caused. There are **two different problems** in this report, and the one in the title is not a router bug.

## The 503 is OpenAI's edge, relayed faithfully

The Envoy-shaped sentence — "upstream connect error or disconnect/reset before headers. reset reason: remote connection failure" — appears nowhere in this codebase. Grepping `src/`, `test/`, and `docs/` for `upstream connect`, `reset reason`, and `before headers` returns no matches. The router does not synthesize it.

The local URL in the message is a red herring. Codex's own error template is:

```
unexpected status <status>: <body>, url: <url>, cf-ray: <...>, auth error: <...>
```

[...]

PR #111, opened 2026-08-08
fix(router): stop a stale selection file and a mid-stream failure from taking the router down

duolahypercho:
Three router-resilience defects found while root-causing #100. **None of them is the 503 in that issue's title** — that turned out to be OpenAI's edge relayed faithfully, with the local URL in the message being Codex naming the endpoint it dialed. Analysis is in [#100](https://github.com/duolahypercho/codex-router/issues/100#issuecomment-5228507232). These are what the investigation found underneath it.

## 1. One state file could wedge every route until restart

`readProviderSelection()` was the only hot-path state reader that threw instead of degrading — `readNativeAliases`, `readNativeRedirect`, and `catalogModels` all swallow. It throws `Unknown provider: X` for any id not in `PROVIDERS`, and that map is built from `config/` at module import, so it is frozen for the process lifetime while the selection file is rewritten by CLI runs that may come from another checkout.

It runs as the **first statement** of `healthPayload()` (before the gateway probe) and on **every** `/responses` turn. So one unknown id — version skew after an update, a renamed or removed provider — 502s everything and fails `/health`, while the process stays alive and bound, so the supervisor keeps reporting it healthy.

The read path now filters unknown ids and reports them; the **write path stays strict**, so a typo on the CLI is still a hard error rather than a silently narrower selection. `doctor` surfaces what was ignored, since the router no longer refuses to serve on it.

One judgment call worth flagging: an explicitly empty list stays empty, because disabling the last provider legitimately writes `[]`. But a non-empty list that filters down to nothing falls back to the no-file default — that file says nothing about the providers this build actually has, so falling back leaves the install coherent instead of stranded with zero routable models. The credential-aware catalog still hides anything that cannot authenticate.

## 2. A mid-stream failure reset the socket instead of ending the stream

`pipeResponse` used `.pipe()`, which does not forward errors or propagate `destroy()`. An upstream body error left the response half-written and never ended, and the top-level handler then called `response.destroy()` — resetting an in-flight chunked body without its terminating `0\r\n\r\n`. reqwest reports that as exactly `error decoding response body`, naming nothing.

Now `stream.pipeline` with `end: false`, so a failure destroys the whole chain while leaving the response for the caller to end cleanly.

**On `end()` vs a terminal error event — both, gated on content type.** A bare `end()` on SSE is worse than it looks: the Responses SSE protocol has no length, so "the stream stopped" and "the turn finished" are the same bytes, and Codex would render a truncated answer as a completed short turn. That trades a cosmetic failure for a silent correctness one. So `text/event-stream` gets an `event: error` frame first — in-protocol (the Responses API defines it), matching the framing `writeCompactionSse()` already emits, and safe under any parser: one that dispatches it surfaces a real failure, one that doesn't ignores the unknown event and lands on the graceful end anyway. `sequence_number` is deliberately omitted rather than guessed. Non-SSE bodies get a plain `end()`; a truncated JSON body already fails the client's parser loudly.

The frame carries a fixed router-side string and never upstream error text, so no response body leaks through it.

## 3. `EADDRINUSE` exited silently

`server` had no `'error'` handler, so a bind failure was an unhandled `'error'` event → silent process exit → a permanent supervisor restart loop with the port never bound and nothing in the log. Now named, with exit codes 98/97/96 a human and a supervisor can tell apart. `KeepAlive` is unconditional with a 10s throttle, so these change nothing about restart behavior — only legibility.

## Plus the reason this was undiagnosable

`src/router.mjs` logged the bare string `[codex-router] request failed` and never the error object. It now logs `name`, `message`, and `code` — and only those, never a body.

## Deliberately not changed

`response.destroy()` in the three forwarders (`api-forwarder.mjs:561`, `oauth-forwarder.mjs:239`, `grok-oauth-forwarder.mjs:495`) looks like the same bug and isn't. Their only client is the router itself over loopback; after this change the router's `pipeResponse` turns that reset into a source error and converts it into a properly terminated client body. Making them end gracefully would make a failed internal hop look like a clean short stream, which the router would relay as a successful truncated turn — the exact failure being removed at the edge. The reset is load-bearing signalling on an internal hop.

## Tests

503 tests, 498 pass, 0 fail, 5 skipped (baseline 491; +12 added). Regression value verified by stashing `src/` and re-running: all three new end-to-end tests fail against the pre-fix code.

`test/router-resilience.test.mjs` (new) spawns the real router: a gateway that resets mid-SSE yields a complete chunked body with the terminal error event and a stderr line matching `\[codex-router\] request failed: \w+: .+`; a taken port exits 98 with the cause named; a selection file with an unknown provider returns `/health` 200 and still routes the known provider in the same file.

`test/pipe-response.test.mjs` reads with the raw `http.request` client so `response.complete === false` catches a reset, and asserts `transform.destroyed === true` — the discriminator between `.pipe()` and `pipeline`.

`test/provider-selection.test.mjs` covers the filter, retired aliases, all-unknown fallback, explicit-empty preservation, corrupt JSON, and that `validateProviderIds`, `writeProviderSelection`, and `enableProvider` all still throw on an unknown id with the previous file left byte-identical.

## Unrelated flake seen while testing

`test/startup-cleanup.test.mjs:35` "startup failure terminates services that already became healthy" fails intermittently under parallel load. Reproduced on **clean main** (1 failure in 3 runs) with none of these changes present, and 4/4 green on this branch. Pre-existing timing sensitivity, not touched here.

### DIFF
--- a/src/doctor.mjs
+++ b/src/doctor.mjs
@@ -216,6 +216,16 @@ try {
       : "none",
     "Run ./bin/setup --guided and choose at least one provider.",
   );
+  // The router no longer refuses to serve on a selection file it cannot fully
+  // resolve, so the damage has to be reported here instead of as a 502.
+  if (selection.degraded) {
+    add(
+      "warn",
+      "Provider selection file",
+      selection.degraded,
+      "Run ./bin/setup --guided, or ./bin/providers enable PROVIDER, to rewrite the selection with this build's provider ids.",
+    );
+  }
 } catch (error) {
   add(
     "fail",
--- a/src/http-utils.mjs
+++ b/src/http-utils.mjs
@@ -1,4 +1,5 @@
 import { Readable } from "node:stream";
+import { pipeline } from "node:stream/promises";
 
 import { secretEqual } from "./caller-auth.mjs";
 import { TARGET } from "./paths.mjs";
@@ -63,6 +64,59 @@ export function copyResponseHeaders(upstream, response, denylist = HOP_BY_HOP_HE
   }
 }
 
+function isEventStream(response) {
+  return String(response.getHeader("content-type") || "")
+    .toLowerCase()
+    .includes("text/event-stream");
+}
+
+function finishResponse(response) {
+  return new Promise((resolve) => {
+    if (response.writableFinished || response.destroyed) {
+      resolve();
+      return;
+    }
+    response.once("finish", resolve);
+    // A client that hangs up before the last chunk drains emits "close"
+    // without "finish"; the request is over either way.
+    response.once("close", resolve);
+    if (!response.writableEnded) response.end();
+  });
+}
+
+// Terminate a response whose body is already streaming.
+//
+// `response.destroy()` resets the socket, so an in-flight chunked body loses
+// its terminating `0\r\n\r\n` and the client reports a transport failure
+// ("error decoding response body") with nothing to say about the cause. Ending
+// the body instead produces a well-formed, if short, HTTP message.
+//
+// A gracefully ended SSE stream, though, is indistinguishable from a completed
+// one: the turn would simply look short and successful. So on `text/event-stream`
+// we first emit a terminal `error` event, matching the event framing the router
+// already writes elsewhere and the Responses API's own `error` event. A parser
+// that understands it surfaces a real failure; one that does not ignores the
+// unknown event and lands on the plain graceful end, which is still strictly
+// better than a reset. The frame carries a fixed router-side message and never
+// upstream error text, so no response body can leak through it.
+export function endStreamedResponse(response) {
+  if (!response || response.writableEnded || response.destroyed) return;
+  if (isEventStream(response)) {
+    try {
+      const data = {
+        type: "error",
+        code: "local_router_stream_failed",
+        message: "The local router lost the upstream response stream.",
+        param: null,
+      };
+      response.write(`event: error\ndata: ${JSON.stringify(data)}\n\n`);
+    } catch {
+      // The socket may already be gone; ending below is still correct.
+    }
+  }
+  response.end();
+}
+
 export async function pipeResponse(upstream, response, denylist, transform) {
   const transforms = transform === undefined
     ? []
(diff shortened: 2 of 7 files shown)


duolahypercho:
Independent review found a HIGH defect in this PR's own mechanism. Fixed and pushed as a second commit.

**The bug.** `endStreamedResponse` wrote its terminal frame assuming the previous write ended on a line boundary. That only holds when upstream fails cleanly *between* two events. A real reset lands wherever it lands, and transforms forward upstream's chunk boundaries verbatim, so the client is usually holding an unterminated `data:` line — a single `output_text.delta` can carry a long span. Reproduced on the wire:

```
data: {"type":"response.output_text.delta","delta":"partial-chunk-no-newlineevent: error
data: {"type":"error","code":"local_router_stream_failed",...}
```

With no newline before it, a conforming parser reads `event: error` as *more of the previous event's data*. So the one signal saying the router lost the stream is never recognized as an error, becomes garbage appended to the last delta, and a lenient consumer can render it as model output. That is the silent corruption this function exists to prevent — the PR was defeating its own purpose in the common case.

The existing tests never caught it because `failingSseUpstream()` always writes a complete `data: first\n\n` before erroring, exercising only the lucky path.

**The fix.** Prefix the frame with a blank line so it always begins a fresh event regardless of where upstream died. Inert when the stream did end cleanly — a blank line with no buffered fields dispatches nothing.

New test asserts the field name begins its own line, that a real SSE parse over the blank-line dispatch boundary finds an event declaring `event: error` with parseable data carrying `local_router_stream_failed`, and that `unterminatedevent` never appears. Verified RED against the unfixed code (fails with "the terminal frame was glued onto the unterminated data line") and GREEN with the fix.

501 tests, 0 failures on current main.

**Two other things the review established, both good news:**

The in-flight counter does **not** leak on client mid-stream disconnect — verified with a real router, a real SSE stream, and a destroyed client socket, releasing in ~85ms. It also passes on pre-PR main, because the old `.pipe()` code already had an explicit `close`-handler guard. So this PR preserves correct behavior rather than fixing a leak.

`server.on("error")` cannot kill a healthy router. Tested under `ulimit -n 40` with 300 concurrent connections: libuv's accept-time EMFILE backoff handles it without surfacing a server-level error, so the handler only fires for genuine `listen()`-time failures.

**One open design question, not a blocker.** When a non-empty selection filters to zero known ids, the router widens to *every* provider. For an operator who deliberately narrowed their selection and holds credentials for providers they excluded on purpose, a build mismatch would silently re-expose them — fail-open on what is otherwise an intentional allowlist. The counter-argument is in the code comment, and `doctor` now warns. Worth a conscious decision rather than treating it as settled; happy to flip it to "stay narrowed" if you prefer.

Issue #56, opened 2026-08-06
Homebrew tap


travankor:
Can you add a homebrew tap?

You can make it self-contained executable with bun or deno for easy distribution.

duolahypercho:
Thanks for the suggestion — a tap is something I'd like to do, with one caveat on the approach.

A self-contained Bun/Deno executable unfortunately isn't possible today: the router is Node on the front, but at runtime it launches a Python LiteLLM gateway from a pinned virtualenv (`litellm[proxy]==1.95.0`). A compiled JS binary can't carry the Python side, so single-file distribution is off the table unless the LiteLLM dependency is replaced entirely.

A regular Homebrew formula in a tap is doable, though:

- `brew install` would put the app in `libexec` with the Python deps built from pinned resources
- upgrades would go through `brew upgrade` instead of the built-in `codex-router update` (which assumes a git checkout)
- you'd still run `codex-router setup` once after install for Codex integration and provider keys, since that part is interactive

The main prerequisite is cutting versioned releases, which the repo doesn't have yet.

Question for you: would a tap that still requires the one-time `setup` step cover your use case, or were you specifically after a zero-config single binary? That determines whether this is a packaging task or a much bigger architectural one.

duolahypercho:
One more piece of architecture context that I think changes the calculus here.

codex-router isn't a CLI you invoke. Installation registers a background service — a LaunchAgent on macOS (`RunAtLoad` + `KeepAlive`), a systemd unit on Linux (`Restart=always`) — that comes up at login and stays up. After setup you just run `codex` and routing happens transparently. You never type a router command again except for occasional maintenance like `doctor` or adding a provider.

That's why the single-binary angle doesn't buy much here even setting the Python dependency aside. You'd download one file, it would immediately tell you to run the interactive setup, and then you'd never invoke it again because the daemon is doing the work. Self-contained executables are a real win for tools you run constantly (ripgrep, fd, that shape of thing) — for a background service, it's mostly cosmetic.

What I do think is legitimate in your request has nothing to do with binaries: package managers *manage* things. `brew upgrade`, `brew uninstall`, knowing what's actually on your machine, and not piping curl into sh — which plenty of developers and most corporate environments simply won't do. That's the real gap, and it's the part worth serving.

[...]

PR #117, opened 2026-08-09
feat(install): install the Python gateway from a hash-verified lock

duolahypercho:
Adds a hash-verified lock for the Python side of the install. Serves two threads: #52 attempted this and I closed it as stale (its pin had drifted twelve minor versions behind main), and #56's Homebrew formula needs pinned, hashed resources before it can be written.

## The gap

`src/install-plan.mjs` pins `litellm[proxy]==1.95.0` and `fastapi==0.139.2`. Their **entire transitive tree** was unpinned, so every install resolved and then executed around a hundred packages nothing had verified — and two machines installing on different days got different trees.

## What ships

`requirements/python.txt` — a **universal** lock: 114 pinned distributions, SHA256 per file, compiled with `uv pip compile --universal --generate-hashes --python-version 3.10`. One file covers macOS, Linux, and Windows on CPython 3.10+ through environment markers:

```
numpy==2.2.6 ; python_full_version < '3.11'
numpy==2.5.1 ; python_full_version >= '3.12'
pywin32==312 ; sys_platform == 'win32'
uvloop==0.22.1 ; sys_platform != 'win32'
```

**The pinned versions are unchanged.** That downgrade is why #52 was closed.

All four install paths — the `uv` and `pip` branches of `bin/install` and `install.ps1` — now install with `--require-hashes`.

## Keeping it from drifting, which is what killed the last attempt

The version literals are **removed** from both shell scripts rather than duplicated. `bin/lock-python` writes the compile input from `PYTHON_REQUIREMENTS` and recompiles. `test/python-lock.test.mjs` fails the suite if the lock, the compile input, and that constant disagree, if any entry is unhashed, if a regeneration lost `--universal`, or if either installer stops checking hashes. Comment prose cannot satisfy it.

This repo has already been bitten by three copies of one rule drifting apart — that was #114's entire story.

## Verified by real installs, not by reasoning

Two clean 3.12 venvs:

```
$ uv pip install --require-hashes -r requirements/python.txt      # 106 packages, 118s
$ python -m pip install --require-hashes -r requirements/python.txt  # 157s
litellm 1.95.0 / fastapi 0.139.2 / litellm-proxy-extras 0.4.81 / uvicorn 0.52.1
```

**It does not break a working install.** Diffed against the venv main's current unlocked installer actually produced on this machine:

```
live (unlocked):  106 distributions
locked:           106 distributions
only in live: none    only in locked: none
version differences: 24  (all transitive, all newer)
```

Identical package *set*. The 24 deltas are exactly the floating drift the lock removes; litellm and fastapi are not among them.

**It runs as a proxy, not just imports.** Started the way `src/start.mjs` does: `/health/liveliness` → `I'm alive!`, `/health/readiness` → healthy, authed `/v1/models` returns the probe model, and `/v1/chat/completions` routed and failed at the deliberately-dead upstream rather than on an import. All 31 `litellm[proxy]` extra requirements present.

**Hash checking cannot reject a distribution pip would legitimately pick.** `--generate-hashes` records every file PyPI serves for a pinned version; verified against PyPI for all 114 entries, 0 mismatches. Cross-platform `--dry-run --require-hashes` resolves on Linux x86_64/aarch64 and Windows x86_64 across 3.10–3.14.

## What I could not verify

- **Windows and Linux were never actually installed**, only resolved via uv's cross-platform dry-run. No such machine here.
- **macOS builds litellm from an sdist via maturin + Rust.** Pre-existing — main installs the same sdist the same way — and the sdist hash is in the lock. But this machine has a Rust toolchain, so whether a macOS box with *no* Rust can install is unproven either before or after this change. Worth flagging that #52's claim that 1.95.0 ships macOS wheels is not what PyPI shows today: manylinux + win_amd64 only. **This matters for #56** — a Homebrew formula will need a Rust build dependency on macOS.
- **Build isolation is not hash-covered.** `maturin==1.9.4` is fetched unhashed into the ephemeral build env for that sdist. Closing that needs a separate build-requirements lock; documented in AGENTS.md rather than glossed over.

## RED/GREEN

| state | result |
|---|---|
| main as-is (no lock, unlocked installers) | 9/9 fail |
| installers restored, lock present | 1 fail — `installerRequirementDrift` |
| #52 simulated (lock pinned to 1.83.9) | 2 fail — `pins litellm==1.83.9 but PYTHON_REQUIREMENTS asks for 1.95.0` |
| final | 9/9 pass |

542 tests, 537 pass, 0 fail, 5 skipped (+9).

Existing installs converge: `stepStatus('python-deps')` against the live venv returns `run`, so the next update applies the lock instead of skipping it as already-matching.

## One unrelated fix included

While rebasing I found main's CHANGELOG describes behaviour #116 does not ship. It claims a child exit aborts the probe in flight and cuts the backoff short — true of the first two attempts, both of which crashed Windows CI; the merged fix drops the child exit listener entirely. Corrected here since the entry is still unreleased and I was already editing the file.


duolahypercho:
Holding this rather than merging it green, because green would have meant less than it looks like.

**CI never runs the installer.** Verified: `ci.yml`'s `test` job does `npm ci`, `npm run check` (which is only `node --check` over `.mjs` files), `npm test`, `sh -n` over the shell scripts, `./install.sh --help`, and a PowerShell `ParseFile`. That `--help` invocation exits at argument parsing on line 164, long before `bin/install` is reached on line 242. The `desktop` job builds Tauri. **Nothing in either job runs `pip`, `uv`, or creates a virtualenv.**

So a PR that changes how every user installs would have merged on evidence consisting of one macOS install on my machine, with Linux and Windows only *resolved* via `uv pip install --dry-run`. A missing wheel, a marker excluding something needed, or a hash not matching what that platform's pip picks would have reached users first.

## What now runs

`.github/workflows/python-lock.yml`, a 4-leg matrix with `fail-fast: false`:

| os | tool | python | why |
|---|---|---|---|
| ubuntu-latest | uv | 3.12 | what `uv venv --python 3.12` pins |
| ubuntu-latest | pip | 3.10 | the documented floor, and the only way to hit the `python_full_version < '3.11'` marker arm |
| windows-latest | uv | 3.12 | |
| windows-latest | pip | 3.10 | |

**The pip command is not written into the workflow.** A new `install-plan.mjs python-install-command <uv|pip> [posix|windows]` extracts it verbatim from `bin/install` / `install.ps1` using the same matcher `installerRequirementDrift` already uses, so the job cannot pass while the real installer fails. A test fails if the workflow ever spells a pip line itself.

`scripts/verify-python-lock.py` is one file so both platforms assert identically: pinned versions match `PYTHON_REQUIREMENTS`, both packages import, every `litellm[proxy]` extra is installed and satisfies its specifier, and the proxy actually starts and answers `/health/liveliness`. On failure it dumps litellm's output — which is where a marker-excluded module surfaces as an ImportError.

## Two bugs the negative control had, both caught before pushing

These are worth recording because both produced a *passing* test that verified nothing.

**The control was testing the wrong failure.** It appended a nonexistent package to the lock, and locally that "passed" — but pip had died on *resolution*, never reaching hash checking. Now it derives a real marker-free entry from the lock, strips its hashes, and greps the output for `hash`, so a resolution error cannot pass for hash enforcement.

**Order matters, and the wrong order is silent.** Placed after the install, `uv pip install --require-hashes` on an already-satisfied requirement prints `Audited 1 package`, exits 0, and never looks at a hash. Reproduced exactly. The control now runs against the empty venv, before the install.

## Judgement calls

**paths filter, plus a weekly cron.** `paths:` is per-workflow, so gating inside `ci.yml` would gate `test` and `desktop` too. The filter's real risk is silently never running — and the case it structurally cannot see is PyPI changing under an unchanged lock — so `schedule: 17 6 * * 1` runs it weekly on main, plus `workflow_dispatch`.

**No macOS leg.** litellm 1.95.0 ships manylinux + win_amd64 wheels only, so macOS builds the sdist via maturin and Rust: minutes of compile, a toolchain to provision, and a failure mode belonging to the sdist rather than to the lock. It is also the one platform already verified by hand.

**No cache.** A lock-keyed cache can only hit when the lock is unchanged — the case needing no speed — and uv serves unpacked wheels from cache, so a hit could skip the very check the job performs. Same mechanism as the ordering bug above.

## Verification

`actionlint` was not available and none was downloaded. Both workflows were parsed with PyYAML and asserted structurally; `ci.yml`'s diff is empty. More usefully, every `shell: bash` step body was executed under GitHub's exact flags (`bash --noprofile --norc -eo pipefail`), which found two `set -e`/`pipefail` hazards, both fixed. The real locked install ran locally on both resolvers, and both pip and uv were confirmed to reject an unhashed requirement in a fresh environment with a hash-naming error.

555 tests, 549 pass, 0 fail — exactly +4 over baseline, the four added.

**Still unproven until this run finishes: the lock installing on Linux and Windows.** That is the entire point of the job, and only a real run can show it. One thing worth a glance in settings: if branch protection has required checks configured, a path-gated workflow marked required would leave unrelated PRs pending.

PR #126, opened 2026-08-09
feat(registry): expose DeepSeek V4 Flash max effort as xhigh in the picker

andychongyz:
## Summary

DeepSeek V4 Flash documents `low/high/max` reasoning, and the registry advertises all three tiers. But the Codex app picker silently drops the `max` tier for routed non-native models, so `opencode-go/deepseek-v4-flash` only offered Low or High.

This maps the advertised top tier to `xhigh`, which the picker renders as "Extra high". The shared DeepSeek forwarder (`deepSeekEffort` in `src/api-forwarder.mjs`) already translates `xhigh`/`max`/`ultra` to the upstream's documented `max`, so selecting Extra high still sends full maximum reasoning effort - the tier just becomes visible and selectable again.

## Scope

Only the opencode-go variant is changed here. The other DeepSeek V4 Flash registry entries (`deepseek` API, `commandcode`, `qwen-plan`) ship the same ladder and are affected identically; happy to extend the change to them if you'd prefer consistency.

## Verification

- `npm run check` passes
- `npm test` passes (623 passed, 0 failed)


duolahypercho:
Thanks for the focused fix. I rechecked this against current `main` and Codex 0.145.0. The repository already has version-aware effort vocabulary handling from #60: builds before 0.143 clamp registry `max` to `xhigh`, while 0.143+ keep the native `max` tier. Changing this one registry entry permanently to `xhigh` would make current builds lose the accurate Max label and would leave the other DeepSeek gateways inconsistent. The original missing-picker problem is therefore already fixed centrally, so this PR should not be merged.

andychongyz:
Thanks for the detailed review. I upgraded to current `main` and re-tested, and I think the version gate rests on a premise that doesn't hold on my build: **this Codex renders no `Max` entry at all — not for routed models, and not for native ones either.**

**Environment**

| | |
| --- | --- |
| codex-router | `0.4.0-beta.3`, `main` @ `250a40b` |
| Codex binary | `/Applications/ChatGPT.app/Contents/Resources/codex` → `codex-cli 0.148.0-alpha.9` |
| ChatGPT app | `26.810.52044` |
| Platform | macOS 25.5.0, arm64 |

**#60 resolves the vocabulary exactly as designed**

```
$ node -e "import('./src/codex-binary.mjs').then(async m=>{
    const v=m.codexVersion();
    const c=await import('./src/catalog.mjs');
    console.log('codexVersion() =>', JSON.stringify(v));
    console.log('codexEffortVocabulary() =>', [...c.codexEffortVocabulary(v)].join(', '));
  })"
codexVersion() => "codex-cli 0.148.0-alpha.9"
codexEffortVocabulary() => minimal, low, medium, high, xhigh, max, ultra
```

0.148 > 0.143, so `clampModelEfforts()` correctly leaves `max` alone and the catalog carries every tier. No complaint about the implementation — it does what it says.

**The native model is the decisive case**

image

`gpt-5.6-terra` is native, so it never passes through the routed catalog or any clamp. It advertises six tiers:

```
gpt-5.6-terra advertises 6: ['low', 'medium', 'high', 'xhigh', 'max', 'ultra']
```

Its picker renders five:

> Light · Medium · High · Extra High · Ultra

`Ultra` is present and labelled. `Extra High` is present. There is no `Max` item. So on this build the picker has no rendering for `max` at all — the tier is dropped even on the native path where codex-router is not involved.

That makes `MAX_EFFORT_SINCE = [0, 143, 0]` the wrong signal here: 0.148.0-alpha.9 passes the gate, but its picker still cannot show `max`.

**Consequences for the routed catalog**

DeepSeek V4 Flash — the original report — after a full quit and relaunch:

```
opencode-go/deepseek-v4-flash ['low', 'high', 'max']
```
image

> Light · High

Two tiers, `max` dropped. Flipping that entry to `xhigh` and running `bin/refresh-catalog` makes the third appear as "Extra high" immediately, with no other change.

Every routed model topping out at `max` is affected the same way:

```
opencode-go/glm-5.3          ['high', 'max']
opencode-go/glm-5.2          ['high', 'max']
opencode-go/kimi-k3          ['low', 'high', 'max']
opencode-go/deepseek-v4-pro  ['high', 'max']
```

And by the same mechanism `gpt-5.6-luna` (native, `[…, 'xhigh', 'max']`) should top out at Extra High rather than Max on this build.

**On "current builds lose the accurate Max label"**

That was the right concern to raise, and I agree a permanent registry rewrite is too blunt a fix — my original PR isn't the shape I'd argue for now. But on this build there is no accurate Max label to lose: the picker never draws one. Two directions that seem better than editing a single registry entry:

1. Treat the effort vocabulary as a property of the *picker*, not the CLI version. If a reliable probe isn't available, a conservative option is clamping `max → xhigh` whenever the tier would otherwise be the model's top offering, so it degrades to a visible tier instead of vanishing.
2. Keep the version gate as the default but allow an explicit override for builds where the picker and the CLI enum disagree.

Routing is unaffected under either: `deepSeekEffort()` in `src/api-forwarder.mjs` already folds `xhigh`/`max`/`ultra` to upstream `max`, and the GLM/Kimi/Qwen profiles do the same.

Happy to test a patch against this build, or to re-do the PR in whichever shape you prefer. And glad to be corrected if I've misread the native picker — that's the piece I'd most want a second pair of eyes on.

PR #131, opened 2026-08-09
fix: validate kimi oauth credential expiry

Bruce-Yii:
## What

`kimiOAuthStatus()` only checked that `access_token` and `refresh_token` fields **exist** — it never validated `expires_at`. A credential file with present tokens but a stale/garbage expiry (e.g. the `1970-01-22` epoch value a broken login flow leaves behind) was reported as `configured: true`.

Found while using login-free with a stale kimi credential file: the dead kimi-oauth models took 3 of the 5 whitelist slots, and every pick from them failed with 402 `unable to verify your membership benefits` (usage-events.jsonl: kimi-oauth/* all 402, opencode-go/* all 200).

## Fix

`src/oauth-status.mjs` — the `configured` verdict now requires `expires_at` to be a real future timestamp:

- `expires_at` missing → conservative pass-through (older CLI builds may not write the field; token presence stays the trusted signal)
- `expires_at` non-numeric / zero / already past / epoch garbage → `configured: false` with setup hint `Run `kimi login` again; the credential is expired`
- seconds- and milliseconds-since-epoch values both handled (1e11 cutoff)
- return shape unchanged (`configured` / `credentialsPath` / `setup` / `scope`)

This also closes the catalog hole: `configuredProviderIds()` → `selectedConfiguredListedModels()` → login-free slot assignment all read through this verdict, so an expired credential no longer lets dead models occupy whitelist slots.

## Tests

New `test/oauth-status.test.mjs`, 9 cases (node:test, following the `grok-oauth-status.test.mjs` style):

| case | expected |
|---|---|
| valid future `expires_at` (seconds) | configured: true |
| millisecond-since-epoch expiry | configured: true |
| expired `expires_at` | configured: false, setup mentions expired |
| epoch garbage (`1900800` = 1970-01-22) | configured: false |
| non-numeric / zero / negative expiry | configured: false |
| missing `expires_at` | configured: true (conservative) |
| missing file / incomplete tokens / invalid JSON | unchanged behavior |

`npm test` on this branch: **662 tests, 651 pass, 10 skipped, 1 known environment failure** (test/control.test.mjs:342 — pre-existing, fails on the untouched base too: a real Codex CLI on the machine makes `nativeCatalog()` re-capture 8 real whitelist slots instead of the test's single one).

Verified live on this machine: the real `~/.kimi-code/credentials/kimi-code.json` had `expires_at` already past → status now reports `configured: false` / "the credential is expired" (previously misreported as logged in).

### DIFF
--- a/src/oauth-status.mjs
+++ b/src/oauth-status.mjs
@@ -6,6 +6,22 @@ export function kimiCodeHome() {
   return process.env.KIMI_CODE_HOME || path.join(os.homedir(), ".kimi-code");
 }
 
+// A credential is only usable while its expiry is a real future timestamp.
+// `kimi login` writes seconds since epoch; some builds write milliseconds, so
+// the cutoff distinguishes the two scales. A missing field is not a failure:
+// older CLI builds may never have written one, and the token fields are the
+// only signal those builds trusted. Anything else — a non-numeric value, a
+// zero, or a date already past (including the epoch garbage that broken login
+// flows leave behind) — must reject the credential.
+function credentialExpired(value) {
+  const raw = value?.expires_at;
+  if (raw === undefined || raw === null || raw === "") return false;
+  const expiresAt = Number(raw);
+  if (!Number.isFinite(expiresAt) || expiresAt <= 0) return true;
+  const millis = expiresAt > 1e11 ? expiresAt : expiresAt * 1000;
+  return millis <= Date.now();
+}
+
 export function kimiOAuthStatus() {
   const credentialsPath = path.join(
     kimiCodeHome(),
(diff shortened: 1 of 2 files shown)


duolahypercho:
Thanks for the report and tests. A deeper check of the current Kimi OAuth path shows this expiry gate would create a regression: `ensureFreshKimiOAuthToken()` intentionally refreshes an expired access token using the still-valid refresh token before forwarding. Marking the provider unconfigured solely because `expires_at` is in the past prevents that refresh path and removes otherwise recoverable sessions from the catalog. The proposed status validator also accepts millisecond expiries and missing `expires_in`, while the actual session reader requires finite second-based `expires_at` and `expires_in`, so status and runtime usability would diverge. The observed 402 membership-entitlement response is not proof that the refreshable credential is structurally expired. This needs an entitlement/refresh-aware diagnosis rather than an access-token expiry cutoff, so I’m closing this implementation rather than merging a catalog regression.

PR #139, opened 2026-08-10
meter: record mid-stream upstream aborts instead of 200

kartikkabadi:
## What

A turn whose upstream stream dies mid-flight was recorded in `usage-events.jsonl` as a plain `200` (the 200 head is committed before the stream finishes). This made every drop invisible in the meter and poisoned all downstream metrics. This change records the truth:

- Upstream stream aborted after the response head: `status: 502` plus an additive `streamAborted: true` marker, no token counts (the stream never reported final usage).
- Client walked away (cancel/close): `status: 0`, no marker, never counted as an upstream failure.

The distinction is guarded by the existing `clientGone` branch plus a direct destroyed-response check for the close-event race. The marker is additive, so existing dashboards keep working.

## Verification

- `npm test`: 676 tests, 669 pass, 0 fail, 7 skipped.
- `npm run check` passes.
- New e2e tests drive a real socket: upstream dies mid-SSE -> 502 + marker; raw client cancel -> 0, no marker.
- Reproduced the bug read-only on the live meter row `2026-08-09T20:49:39.265Z` (recorded 200 while router.log line 22538 logged a MidStream transfer error).

## Notes

Native `/v1/responses` traffic shares the same handler and gets the same treatment; image/web-search native paths are single-shot JSON and untouched. Pre-existing behavior: an upstream failure before headers are sent still records no row (not a regression from this change).

### DIFF
--- a/src/router.mjs
+++ b/src/router.mjs
@@ -1262,12 +1262,15 @@ async function handleResponses(request, response, requestUrl) {
   const startedAt = Date.now();
   const activity = beginRequestActivity();
   let clientGone = false;
+  let requestedModel = "";
+  let route;
+  let upstreamRetries;
   try {
     if (!requireCodexTransport(request, response)) return;
     const encoded = await readRequestBody(request);
     const body = decodeBody(encoded, request.headers["content-encoding"]);
     const payload = parseBody(body);
-    const requestedModel = typeof payload.model === "string" ? payload.model : "";
+    requestedModel = typeof payload.model === "string" ? payload.model : "";
     let registeredRoute =
       MODEL_BY_SLUG.get(requestedModel) ??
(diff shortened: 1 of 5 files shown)


kartikkabadi (commented):
Two-pass review of the current head `d854091a35f517267e11f1fde1b61fedeb3dc319`: first across the mid-stream failure/cancel accounting paths and response lifecycle, then across the serialized event schema, recent-event reader, and regression coverage.

I found no new material correctness, security, or maintainability issue in the implementation. The change keeps upstream aborts distinct from client cancellation, records the post-header failure as status 502 with an additive `streamAborted` marker, preserves token omission for truncated streams, and adds end-to-end coverage for both failure modes. The current head is not fully validated because the exact-head CI workflow is `action_required`.

Review #133
Automated
Reviewed by ChatGPT - Sol

Issue #138, opened 2026-08-09
Several Windows and provider fixes, plus an opt-in native-catalog adoption — issue before PRs


cududa:
I've been running Codex Router on Windows and have a small set of fixes and opt-in catalog improvements I'd like to contribute. I opened this first to check appetite before sending PRs upstream.

Each item is self-contained, tested, and now rebased on current `main`. I use a custom model catalog to control picker metadata and Fireworks as an inference provider for Kimi K3.

## Current PR overview

These are draft PRs on my fork; I will open them against upstream one at a time, in whichever order or shape you prefer:

- **[Test hermeticity on Windows](https://github.com/cududa/codex-router/pull/6)** — `control.test.mjs` used `/usr/bin/true`, which does not exist on Windows, so it could re-capture a real local catalog. Uses `process.execPath` instead.
- **[Fireworks `web_search_options` compatibility](https://github.com/cududa/codex-router/pull/5)** — Fireworks rejects this Codex field; strip it at the three forwarding boundaries.
- **[Native-catalog adoption (opt-in)](https://github.com/cududa/codex-router/pull/2)** — validate and reversibly adopt a pre-existing, non-router `model_catalog_json` rather than forcing manual config edits.
- **[Provider service-tier catalog metadata](https://github.com/cududa/codex-router/pull/9)** — let a model explicitly advertise provider service tiers, without changing any model unless it opts in. This supports a locally curated Fireworks model exposing Priority while standard service remains the default.
- **[Safe additive model curation](https://github.com/cududa/codex-router/pull/8)** — make non-interactive `--models` add models without discarding existing curated entries or their metadata; use explicit `--remove` for pruning.

The native-catalog adoption is the largest change, so I especially welcome direction on that approach before opening anything upstream.

cududa:
Note: I see that upstream has had some changes in the days since I prepared these; reworking my PR's now.

duolahypercho:
Integrated the accepted work from the five fork PRs directly into upstream `main`, with contributor credit:

- `b94d545` — make the control test hermetic on Windows
- `24c9383` — strip unsupported `web_search_options` for Fireworks at every request boundary
- `fb5851f` — make scripted model curation additive and add explicit `--remove`
- `26aadd9` — preserve validated provider service tiers without inheriting a paid default
- `eaf55a9` — explicitly adopt a user-owned native catalog with transactional rollback and exact-path restoration

[...]

PR #140, opened 2026-08-10
test: lock in token-estimate cap, predicate, and disable path

kartikkabadi:
## What

Tests only, no runtime change. Locks in the behavior of the zero-input token estimator so it can't silently regress:

- The estimate is capped at the provider's context window.
- The substitution fires only on an explicit numeric `input_tokens`/`prompt_tokens: 0`; positive or missing counts pass through untouched.
- Fixture-driven test uses the exact response shape recorded from `opencode-go/deepseek-v4-flash` during live operation (provider reports `input_tokens: 0`; router stores the byte-based estimate in a separate `estimatedInputTokens` field).

## Verification

- `npm test`: 677 tests, 670 pass, 0 fail, 7 skipped.
- `npm run check` passes.
- The disable path (`CODEX_ROUTER_ZERO_INPUT_ESTIMATE=0`) is covered by the existing routing test on the base.

## Note

The provider keeps reporting explicit zero input tokens, so the estimate remains the err-high bias that prevents premature compaction; this PR only proves the estimate stays bounded, predicate-gated, and clearly marked.

### DIFF
--- a/test/fixtures/upstream-zero-usage.json
+++ b/test/fixtures/upstream-zero-usage.json
@@ -1,0 +1,38 @@
+{
+  "note": "Lane L4 fixture. Constructed from the upstream response shape opencode-go/deepseek-v4-flash actually returned during the overnight baseline (2026-08-09 21:46-21:47 IST, work/baseline/usage-events.jsonl rows with inputTokens:0 and estimatedInputTokens). The router records the provider's reported counts verbatim in the usage event and writes the estimate to a separate field, so the upstream payload itself is not captured by the router; this fixture reconstructs the response.completed SSE event body from the recorded event fields plus the identical shape used by test/routing.test.mjs and the #95 repro.",
+  "recorded": {
+    "at": "2026-08-09T21:47:16.132Z",
+    "model": "opencode-go/deepseek-v4-flash",
+    "provider": "opencode-go",
+    "status": 200,
+    "inputTokens": 0,
+    "outputTokens": 469,
+    "totalTokens": 469,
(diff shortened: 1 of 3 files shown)


kartikkabadi (commented):
Two-pass review of the current head `0c04393bfc53019919cbd33c80b5891e77a7644c`: first across the estimator-cap and predicate-specific cases, then across the fixture-backed SSE transform assertions and the disable-path coverage.

I found no material correctness, security, or maintainability issue in this tests-only PR. The new cases distinguish explicit numeric zero from positive or missing counts, cap estimates at the supplied context window, preserve the provider's original telemetry while exposing the estimate separately, and exercise the recorded upstream response shape with a safe fallback when the external evidence file is unavailable. Exact-head CI is currently `action_required`, so repository validation is incomplete.

Review #134
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Reviewed the delta from Review #134 through exact head `8f3bfcaa01c89863891187f45e63548a43e8e436` in two passes: first the new context-window cap and positive/missing-count predicates, then the fixture-backed SSE transform and route-level regression coverage.

I found no new material correctness, security, or maintainability issue. The added tests keep provider-reported telemetry separate from estimated values, cap the substitution at the declared context window, and preserve pass-through behavior for positive or absent input counts. The fixture is explicitly labeled as a synthetic reconstruction rather than captured upstream evidence, which keeps the test’s provenance honest.

Exact-head repository validation should still be confirmed before landing.

Review #138
Automated
Reviewed by ChatGPT - Sol

PR #141, opened 2026-08-10
relay: inherit session model for spawned threads when model omitted

kartikkabadi:
## What

Routed sessions that spawn a thread or continue one with `create_thread` / `send_message_to_thread` now inherit the session's own model when the `model` field is omitted. Before this, an omitted model left the new thread on the app's default (native), which can be quota-blocked; the spawn died instantly with `usage_limit_exceeded` while the parent kept waiting, looking like a randomly stopped session.

Behavior:

- Model omitted + routed session -> router injects the session model (`route.slug`), so children use the same custom provider.
- Explicit user-specified model -> always wins, untouched.
- Non-routed / native traffic -> untouched.
- Malformed args, non-spawn tools, missing session model -> pass through untouched. Injection is idempotent.

Note: this branch is based on `pr/relay-namespaces` (the namespace relay that introduces `namespace-relay.mjs`); merge order is relay-namespaces -> this PR.

## Verification

- `npm test`: 701 tests, 694 pass, 0 fail, 7 skipped.
- `npm run check` passes.
- New tests: routed+omitted injects, explicit model untouched, non-routed untouched, idempotency, malformed args, other namespaces.

## Known limit

Injection is request-side: the executed `create_thread` output call is not rewritten to carry the model (the response transform restores names only). The fix works when the model follows its own injected history; it does not force the executed call. Live app execution of a spawned thread was not available to me in this environment, so the end-to-end claim rests on the relay mechanism plus tests.


kartikkabadi (commented):
Reviewed exact head `a3912b2e61d78d642053a2a9da214175f9b80e48` in two passes: first the session-model inheritance and argument-rewrite path, then explicit-model precedence, non-routed passthrough, idempotency, malformed inputs, namespace coverage, and the added regression tests.

I found no new material correctness, security, or maintainability issue in this PR. The request-side injection is scoped to routed sessions, preserves explicit models, leaves malformed and unrelated calls untouched, and is idempotent. The documented limitation is clear: the executed child-call output is not rewritten, so the fix depends on the model following the injected history.

The PR-reported test/check results are not independently re-run in this review pass; confirm the exact-head CI result before landing.

Review #136
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Reviewed exact head `33b6c6229e9b31b90bb3df607a5922b908dee08b` in two passes: first the namespace flatten/restore and session-model injection paths, then non-streaming versus SSE responses, malformed and ambiguous tool names, history rewriting, and the regression-test coverage.

I found no new material correctness, security, or maintainability issue in this head. The request-local lookup preserves namespace identity even when names contain `__`, explicit child models remain authoritative, omitted models inherit the routed session model only for the intended tools, and native/non-routed traffic is left outside the flattening path. The new relay also handles both streamed and non-streamed response shapes.

Exact-head CI is still in progress, so landing validation is incomplete until run #402 finishes.

Review #139
Automated
Reviewed by ChatGPT - Sol

PR #142, opened 2026-08-10
relay: full native toolset and namespaces to routed models

kartikkabadi:
## What

Relays the Codex app's full native toolset to routed (non-OpenAI) models. LiteLLM's Responses -> Chat Completions bridge drops `type: "namespace"` tools, which is how the app ships the collaboration runtime, the app toolset (threads, automations, navigation), and every MCP server (node_repl, peekaboo, github, ...). This PR:

- Flattens every namespace tool to `<namespace>__<tool>` functions the provider accepts (`namespace-relay.mjs`).
- Renames stored call history to the flattened form so the model's transcript matches its tool list.
- Restores function calls back to the app's native `{name, namespace}` shape on the way out (`NamespaceToolCallTransform`).
- Merges the app's deferred `codex_app` tool definitions into routed requests (`codex-app-tools.mjs`) and removes the `create_thread` description hint.

The router only relays definitions and results; it never executes an app tool.

Commits: `96767e0`, `e7fa240`, `30a9fec` (content-identical to the originals, no cherry-pick trailers).

## Verification

- `npm test`: 693 tests, 686 pass, 0 fail, 7 skipped.
- `npm run check` passes.
- Content identity vs the original commits verified per-commit by independent review.

## Stack

Merges before `pr/skill-pack` (doctor and skills tests import `codex-app-tools.mjs` from this PR).


kartikkabadi (commented):
Reviewed exact head `fd9de5075c47b78d2ca0eb565011f144ecf1732d` in two passes: first the namespace flatten/restore and deferred app-tool merge paths, then history replay, SSE transformation, and the transcript chunker.

I found no new material correctness, security, or maintainability issue in this PR. The relay preserves tool schemas, uses exact namespace lookups rather than delimiter splitting, keeps ambiguous bare calls untouched, and the end-to-end tests cover MCP namespaces, nested namespace names, history normalization, response restoration, and ordinary-tool pass-through. The chunker also preserves event coverage and flags oversized records instead of silently splitting them.

The PR-reported test/check results are not independently re-run in this review pass; confirm the exact-head CI result before landing.

Review #135
Automated
Reviewed by ChatGPT - Sol

PR #144, opened 2026-08-10
tooling: verify skill injection and document relay operations

kartikkabadi:
## What

Tooling around the skill pack and relay:

- `verify-skill-injection.mjs`: probes the newest routed session and asserts (a) the pack is listed, (b) the model reads the skill, (c) `create_thread` shape, (d) native non-read; `--latest` auto-finds the newest session.
- `codex-app-threads` skill: warns against the `mcp__codex_apps__` prefix.
- Skill documentation for model inheritance, meter semantics, and session-stop diagnosis (added during review).

Commits: `e571550`, `226f877`, `d817843` + `758a79e` (skill doc commit added at review, content otherwise identical to originals).

## Verification

- `npm test`: 716 tests, 709 pass, 0 fail, 7 skipped (before the doc commit; the doc commit adds no code).
- `npm run check` passes.
- Injection probe passes against the newest routed session (thread created first try, browser + computer-use tool paths relay).

## Stack

Merges last, after `pr/relay-namespaces` and `pr/skill-pack`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -969,17 +969,20 @@ node src/skills-install.mjs uninstall
 against the checkout, free of name collisions, and matching the app
 toolset snapshot the router relays.
 
-To verify the pack actually reaches a session, run the probe against a
-rollout after using a custom model in the app:
+To inspect rollout evidence for the pack, run the read-only check after using
+a custom model in the app:
 
 ```sh
 node scripts/verify-skill-injection.mjs ~/.codex/sessions/2026/08/09/rollout-*.jsonl
+node scripts/verify-skill-injection.mjs --latest --expect routed
 ```
 
-It asserts the session listed a pack skill, that the model read it, that
-`create_thread` calls carried the required fields, and (with
-`--expect native`) that native sessions never read the pack. The browser
-and computer-use bootstrap one-liners are live-only and flagged as such.
+It accepts only a standalone app-injected developer block with a turn ID, then
+correlates a same-turn tool call referencing the skill path with its output and
+checks same-turn `create_thread` arguments. With `--expect native`, that
+completed pack-path call is an error. Because arbitrary exec code is opaque,
+the rollout proves a completed path-referencing call, not that the command read
+specific bytes. Browser and computer-use execution remains live-only.
 
 ## Common commands
 
--- a/README.md
+++ b/README.md
@@ -965,6 +965,22 @@ node src/skills-install.mjs install
 node src/skills-install.mjs uninstall
 ```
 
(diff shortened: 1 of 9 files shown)


kartikkabadi (commented):
I found one merge-blocking installer behavior issue in the current delta.

**P1 — the installer does not implement its documented best-effort skill installation.** In `bin/install`, the new `node src/skills-install.mjs install` call is executed directly. If the skill copy/validation fails, the installer exits at that point instead of continuing to configure the router. The adjacent comment says the failure “must never roll the router back,” so this needs an explicit non-fatal boundary (for example, capture/report the failure and continue, or otherwise disable `set -e` for that command). Otherwise a local skill-pack problem can leave the main Codex integration only partially installed.

Exact-head workflow status is not green: CI and Python lock are both `action_required`, so the branch is not fully validated yet.

Review #132
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Two-pass review of the current head `7fc4353413fd3409a79816f265378e9270e44129`: first across the installer/uninstaller and skill-pack lifecycle, then across the relay-evidence tooling, rollout verifier, transcript slicing, doctor/manifests, and regression coverage.

I found no new material correctness, security, or maintainability issue in this delta. The skill installer is marker-scoped and non-clobbering, stale managed entries are pruned safely, uninstall only removes marker-owned directories, and the verification script keeps live-only claims explicit rather than treating rollout evidence as proof of browser/computer-use execution. The added tooling is read-only or uses isolated temporary state, and the documented test results are consistent with the changed surfaces.

The PR remains stacked and should be merged only after the prerequisite relay/skill-pack PRs, as documented.

kartikkabadi (commented):
Two-pass review of the current head `d4f624d7ba42c00a78c3a773f80a0f1002ba596`: first the rollout-verification script and its parser/assertion logic, then the skill documentation, namespace relay/model-inheritance changes, and regression coverage.

No new material correctness, security, or maintainability issue was found in this delta. The verifier keeps rollout evidence bounded to a trusted same-turn developer skills block, correlates completed tool outputs by call ID, treats browser/computer-use execution as live-only, and distinguishes routed versus native expectations. The added tests cover malformed input, latest-rollout selection, turn scoping, completed versus incomplete skill-path calls, and create_thread argument validation. The previously reported installer failure is outside this incremental delta and is not duplicated here.

The PR remains stacked and should land only after its prerequisite relay/skill-pack PRs, as documented.

Review #142
Automated
Reviewed by ChatGPT - Sol

PR #145, opened 2026-08-10
Fix empty completions: retry once, surface an error on repeat

kartikkabadi:
## Problem

A routed turn that completes with no output text and no tool call is invisible to the Codex client: it records the empty response as a normal `task_complete`, so the agent appears to "randomly stop" with no error. The usage meter shows 175 of these on `opencode-go/deepseek-v4-flash` alone (status 200, zero output tokens).

## Change

Add an `EmptyCompletionGuard` transform to the routed stream path (`src/empty-completion-guard.mjs`). It holds the terminal SSE events (`response.completed`, `response.done`, `[DONE]`) until it knows the turn produced content. Reasoning deltas are not content. When the turn is empty:

1. The terminal events are suppressed, so the client never sees a completed-but-empty response.
2. The identical request is retried once against the same response (same bytes, same headers, same signal).
3. A retry that produces content reaches the client normally.
4. A retry that is also empty emits an explicit SSE error event (`empty_completion`) instead of a second silent success.

`pipeResponse` gains `leaveOpen` and `omitHead` options so the retry can continue the already-started response (headers were sent with the first attempt).

## Metering

- Empty completion recovered on retry: `status: 200` + `emptyCompletionRetried: true`
- Double-empty turn: `status: 502` + `emptyCompletion: true` + `emptyCompletionRetried: true`
- Markers are additive; ordinary rows keep their exact shape.

## Tests

- `test/empty-completion-guard.test.mjs` — guard unit tests (content turn, reasoning-only turn, retried mode, tool-call turn, `[DONE]` sentinel, non-SSE passthrough).
- `test/empty-completion-router.test.mjs` — router-level: empty-then-content retry reaches the client, double-empty surfaces an error and meters 502, content turns are untouched.
- Full suite: 796 tests, 0 failures. `npm run check` passes.


kartikkabadi (commented):
Found one material correctness issue in the retry path:

- **P1 — retry response headers/status are not reconciled.** After the first empty SSE attempt, the response head is already staged/sent with the first upstream's status and headers. The second attempt is relayed with `omitHead: true`, but the code still accepts any `upstream2.status` and body. If the retry returns a non-SSE error, a different content type, or a bodyless status, the client still receives the original 200/SSE response head with the retry payload, while usage records the retry status. That can turn an upstream retry failure into a malformed or misleading successful client response. The retry needs to be constrained to a compatible successful streamed response, or the first head must be delayed until the retry outcome is known; otherwise emit a protocol-level error using the already-committed response contract.

The guard/unit coverage is strong for empty→content and empty→empty SSE flows, but it does not establish this cross-status/content-type behavior. Please add a regression test for a first empty SSE followed by a non-2xx or non-SSE retry and make the client-visible response deterministic.

Review #140
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Validation-only follow-up for exact head `0b6b46aed60fa49bd2e6f735b8f4c97d972ba904`.

CI run `31371193626` is now green. The previously reported P1 remains open on this unchanged head: if the first attempt has already committed a successful response, a retry that returns a non-2xx, non-SSE, or bodyless response can still be relayed through the already-started client response, producing a misleading result. No duplicate technical finding is added here; this comment records the condition-watch status only.

Review #148
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Two-pass review of exact head `dfe9d16b5e86bedea7e310a5110aae5826fa28ed` found one new merge-blocking correctness issue in the empty-completion guard.

**P1 — A Responses stream whose only content is carried in `response.completed.response.output` is misclassified as empty.**

`#consumeBlock()` checks `isTerminalEvent(...)` before it checks `#contentOf(...)`. Because `response.completed` is always treated as terminal, the implementation holds that block without evaluating the completed payload, even though `isContentEvent()` explicitly supports `response.completed` with a non-empty `response.output`. A gateway that emits the final output only in the completed event will therefore have its valid content discarded, trigger an unnecessary identical retry, and potentially surface `empty_completion` after a second valid completion. The current tests only cover a completed event with `output: []`, so they do not protect this contract.

Minimal fix: parse/evaluate the completed payload for content before classifying the event as terminal, or make terminal detection conditional on `!isContentEvent(...)`. Add a regression test where `response.completed` contains a text message or function call in `response.output` and assert the block is forwarded, no retry is triggered, and `isEmpty()` is false.

The previously reported retry-response status/header compatibility issue also remains unresolved on this head; I am not duplicating that finding here.

Review #150
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
## Review #151 — multiline SSE data is misclassified as empty

**Head reviewed:** `41682dbd8e5018c2d2fa0734c2e46c84f9b51fc0`

### P2 — `#fields()` only retains the last `data:` line

The guard parses each SSE block with:

```js
else if (line.startsWith("data:")) dataText = line.slice(5).trimStart();
```

SSE permits one event payload to be split across multiple `data:` lines; the client reconstructs those lines with newline separators before parsing the event data. This implementation overwrites earlier lines, so a valid JSON event split across lines is reduced to the final fragment, `JSON.parse()` fails, and the guard treats the turn as contentless. A normal completion can then have its terminal events suppressed, trigger an unnecessary retry, or end as `empty_completion`.

Please accumulate all `data:` fields for the block and join them with `\n` before parsing. Add a regression test for a streamed content event whose JSON is split across multiple `data:` lines, asserting the content is forwarded and no retry is requested.

**Confidence:** Medium-high; this is directly reachable in the new parser and is not covered by the described tests.

Review #151
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Exact-head validation follow-up for `41682dbd8e5018c2d2fa0734c2e46c84f9b51fc0`: CI run `31379233043` completed successfully. The new P2 from Review #151 remains open: the SSE parser overwrites earlier `data:` lines instead of joining them, so a valid multiline event can be misparsed as empty and trigger an unnecessary retry or `empty_completion`. No duplicate technical finding is added here.

Review #152
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Two-pass review of current head `41682dbd8e5018c2d2fa0734c2e46c84f9b51fc0` found the prior correctness findings still unresolved:

- **P1 — retry response compatibility is still unchecked.** After the first empty SSE attempt commits the initial response headers, the retry is streamed with `omitHead: true` without enforcing a compatible successful SSE response. A retry that returns non-2xx, non-SSE, or bodyless data can therefore be relayed through the already-committed 200/SSE client response and present a malformed or misleading success. Add a regression test and fail closed to a deterministic protocol error.

- **P2 — multiline SSE `data:` fields are still overwritten.** `#fields()` assigns `dataText` on each `data:` line instead of accumulating the lines and joining them with `\n`, so valid multiline JSON events can be truncated and misclassified as empty, causing an unnecessary retry or `empty_completion`. Add the multiline-event regression test.

The completed-event content ordering issue from Review #150 appears fixed on this head: `#consumeBlock()` now evaluates `#contentOf(...)` before terminal handling. No duplicate finding is added for that resolved issue.

Review #153
Automated
Reviewed by ChatGPT - Sol

kartikkabadi (commented):
Two-pass review of exact head `41682dbd8e5018c2d2fa0734c2e46c84f9b51fc0` found the previously reported SSE parsing issue still unresolved and one adjacent retry-compatibility gap.

- **P1 — retry accepts a successful non-SSE response into an already-committed SSE stream.** The new branch only rejects `!upstream2.ok`. If the first attempt is empty SSE and the retry returns `200 application/json` (or another non-SSE success), `retryContentType` selects that JSON content type, `EmptyCompletionGuard` passes the body through unchanged, and `pipeResponse(..., omitHead: true)` appends the JSON bytes to the first attempt's already-sent `text/event-stream` response. The client receives a malformed mixed protocol while usage records a successful retry. Constrain the retry to a compatible streamed response (at minimum require a successful `text/event-stream` response) or emit the router's deterministic SSE error for incompatible retry content. Add a regression test for empty SSE → 200 JSON.

- **P2 — multiline SSE `data:` fields are still overwritten.** `#fields()` assigns `dataText` for each `data:` line instead of accumulating the fields and joining them with `\\n`. A valid SSE event whose JSON is split over multiple `data:` lines is truncated, `JSON.parse()` fails, and the guard can suppress a real completion or trigger an unnecessary retry. Add a multiline-event regression test and join all `data:` fields per SSE block.

The completed-event-only content issue from Review #150 is fixed on this head; no duplicate finding is added for it.

Review #154
Automated
Reviewed by ChatGPT - Sol

src/empty-completion-guard.mjs
```diff
@@ -0,0 +1,303 @@
+import { Transform } from "node:stream";
+import { StringDecoder } from "node:string_decoder";
+
+const SSE_FIELD_LINE = /^(?:event|data|id|retry):/;
```
Bruce-Yii: The headerless-SSE sniff in `EmptyCompletionGuard` and the one in `prepareEventStreamRetry` disagree on what counts as SSE. `prepareEventStreamRetry` uses `HEADERLESS_SSE_PREFIX` (router.mjs:394), which explicitly tolerates a UTF-8 BOM, leading comment lines (`: ...`), and leading blank lines. The guard's `SSE_FIELD_LINE` (empty-completion-guard.mjs:4) requires the buffer to start with `event:`/`data:`/`id:`/`retry:` immediately — no BOM, no leading comment, no blank line.

Concrete failure: an upstream whose first bytes are `: keepalive\n\n` (some gateways emit an initial comment) is classified as `#eventStream = false` by the guard during the first attempt, so the whole stream is passed through untouched and the empty-completion gate never runs. The same bytes on the retry path are accepted by `prepareEventStreamRetry` and piped with `pipelineContentType = "text/event-stream"`. So the first attempt silently skips the guard while the retry detects it — inconsistent, and the empty-completion detection is exactly the thing that must not silently skip.

Suggestion: extract a single shared prefix predicate (tolerating BOM + leading comment/blank lines, i.e. the `HEADERLESS_SSE_PREFIX` behavior) and use it in both places, so a stream is classified identically on both attempts.

src/empty-completion-guard.mjs
```diff
@@ -0,0 +1,303 @@
...
+      this.push(chunk);
+      callback();
+      return;
+    }
+    const bytes = Buffer.from(chunk);
+    this.#chunks.push(bytes);
+    this.#bufferedBytes += bytes.length;
+    this.#parseBuffer += this.#decoder.write(bytes);
+    this.#consumeBlocks();
+    if (!this.#released && this.#bufferedBytes > this.#maxPreludeBytes) this.#release();
```
Bruce-Yii, line 202: The 30s / 1 MiB release path silently disables the empty-completion gate for long-thinking turns. When `#release()` fires (byte cap at line 189, timer at line 271), the stream is relayed as-is and `#empty` is never set — `isEmpty()` stays false. That is the right conservative choice for "we can't hold a real turn hostage", but it means the exact failure this PR is fixing (reasoning-only turn that completes with no output) still slips through silently when the reasoning phase takes longer than 30s or exceeds 1 MiB. The hold-budget guards exist to avoid unbounded buffering, not to make the failure mode conditional on duration.

Suggestion: instead of dropping the verdict entirely on budget exhaustion, mark the guard as "indeterminate" and have `handleResponses` record a distinguishable usage-event marker (e.g. `emptyCompletionGuardReleased: true`) on that turn. The operator keeps the telemetry that says "this turn may have been an empty completion we chose not to retry", and the retry behaviour stays as-is. Without a marker, a 40-second empty completion looks identical to a healthy 40-second turn in `usage-events.jsonl`.

src/empty-completion-guard.mjs
```diff
@@ -0,0 +1,303 @@
...
+      // The SSE algorithm joins repeated data fields with a line feed before
+      // dispatch. Overwriting them truncates valid multiline JSON and can turn
+      // a content event into a false empty completion.
+      dataText: dataLines.length ? dataLines.join("\n") : undefined,
+    };
+  }
+
+  #contentOf(eventType, dataText) {
+    if (!dataText || dataText === "[DONE]") return false;
+    try {
```
Bruce-Yii, line 315: Non-SSE data blocks are classified as "no content" and can turn a healthy turn into a retried one. In `#contentOf`, a block whose `data:` payload is not parseable JSON (or whose event type is unknown) falls through to `return false`. For a gateway that mixes a non-JSON sentinel line (e.g. an `: keepalive` comment is fine since it has no `data:`, but a `data: pong` from a custom heartbeat) with the real SSE events, the guard would hold the stream and, once `response.completed` arrives, mark the whole turn empty — even though a genuine `output_text.delta` earlier in the stream was already classified as content and `#sawContent` is true.

Wait — checking `#classifyBlock` again: content detection happens first, and `#release()` fires as soon as any content event is seen, so a genuine delta releases the stream before a later non-JSON block can matter. The risk is narrower: only streams where the *first* meaningful event is the terminal one (a "whole turn rides on `response.completed`" gateway) and the completed payload itself fails JSON.parse. For those, the turn is treated as empty and retried even though it carried output.

Suggestion: when the terminal event's data fails to parse, prefer releasing the stream (indeterminate) over declaring it empty — a parse failure at the terminal event is not evidence of "nothing was produced".

Bruce-Yii (commented):
Solid approach overall — holding the pre-content stream so a retry contributes zero bytes to the client's view is the right design, and the meter/doc work (double-billing, `emptyCompletionRetried` markers, env kill-switch) is thorough. Three issues to consider, all in the guard's classification edge cases:

1. **Sniff disagreement between the two attempts** (empty-completion-guard.mjs:4 vs router.mjs:394): the guard's `SSE_FIELD_LINE` requires the first byte to be `event|data|id|retry:`, while `HEADERLESS_SSE_PREFIX` tolerates BOM/comment/blank leading lines. A stream that starts with `: keepalive\n\n` is passed through unguarded on the first attempt but guarded on the retry — the detection silently never runs for the very turn that needs it. Shared predicate would fix both.

2. **Budget-exhaustion turns lose all telemetry** (empty-completion-guard.mjs:189/271): after the 30s or 1 MiB release, `isEmpty()` stays false, so a 40-second reasoning-only empty completion records as a healthy 200 in usage-events. A `emptyCompletionGuardReleased` marker would keep the failure observable without changing the conservative no-retry choice.

3. **Terminal-event parse failure flips to empty** (empty-completion-guard.mjs:296): a gateway that rides the whole turn on `response.completed` and sends non-JSON data there gets retried as empty despite carrying output. Prefer releasing (indeterminate) on parse failure at the terminal event.

None of these block the feature; they are classification robustness gaps. Happy to expand any of them with a failing fixture.

PR #146, opened 2026-08-10
Flatten every namespace tool, not just collaboration

duolahypercho:
## The bug

Codex ships MCP servers, bundled plugins (browser, computer-use, sites), and its collaboration toolset to the model as Responses-API tools of `type: "namespace"`. LiteLLM's Responses -> Chat Completions bridge drops that type outright:

```python
elif _tool_type in ("computer_use", "image_generation", "namespace", "shell"):
    # Drop unsupported Responses-API-only tool types...
    continue
```

So on every provider that is not on the Responses protocol — which is all of them except `meta` and opencode's Responses surface — the model was offered none of those tools and reported them missing. Only the `collaboration` namespace had been rescued, and only because routed subagents broke first.

Reported as "the browser and Chrome don't work on DeepSeek, `mcp__node_repl__js` is missing". It was never DeepSeek-specific.

## The fix

`src/namespace-tools.mjs` (replacing the collaboration-only module) flattens every namespace into ordinary functions the bridge carries, and completes the round trip:

- **Tool list** — `mcp__node_repl` + `js` becomes `mcp__node_repl__js`
- **Stored history** — namespaced `function_call` items are renamed to match, so the model does not copy a name it was never offered out of its own transcript
- **Response stream** — the call is restored to `{name, namespace}` before it reaches Codex

Restoration goes through a per-request map rather than splitting on the delimiter: a namespace named `mcp__node_repl` makes `__` ambiguous, and a name shortened to fit a provider's 64-character function-name limit cannot be split at all. A bare name is recovered only when exactly one namespace claims it — two namespaces sharing a tool name are left for Codex to reject rather than dispatched to the wrong server. Responses-protocol providers are untouched.

## Verification

Live against DeepSeek V4 Pro through the real gateway, with the gateway hop proxied so the forwarded bodies were inspected rather than inferred.

**Before** — the model cannot see the tools and falls back to shelling out:

```
repl:    call name=exec_command  arguments={"command": "node -e \"console.log(6*7)\""}
browser: call name=exec_command  arguments={"command": "open https://example.com"}
```

**After**:

```
repl:    call name=js        namespace=mcp__node_repl  arguments={"code": "6*7"}
browser: call name=navigate  namespace=browser         arguments={"url": "https://example.com"}
```

**Two-turn test with a real browser executing the call.** Turn 1 forwarded `browser__navigate` as a plain function and the model called it; the call was executed for real; turn 2 forwarded the history as `function_call name=browser__navigate namespace=(none)` matching the tool list, and the model answered `The page's main heading is "Example Domain".` from the real tool output. That second turn is the leg that matters — an un-renamed history is what taught the model to emit bare names in the collaboration regression.

673 tests pass. New coverage in `test/namespace-tools.test.mjs` plus an end-to-end case in `test/routing.test.mjs` that drives a real router process and asserts the flattened tools, the renamed history, and the restored call together.

## Not yet measured

Codex's own dispatch of the restored `{name, namespace}` pair — that needs a real Codex turn against an installed build. It is the identical mechanism the collaboration path already uses in production and the exact shape Codex sent, but it is inference until someone runs it.

## Notes

`AGENTS.md` gains a section on the rules that keep this correct, since the narrow version of this fix is what caused the bug.


duolahypercho:
Closing as redundant.

`main` already fixes this in `src/namespace-relay.mjs`, which flattens every namespace, renames the stored history, and restores the calls. This PR was written against a local `main` that was 46 commits behind and reimplements that work more narrowly.

Merging it would be a regression: it rewires `src/router.mjs` onto the narrower module and drops `rewriteNamespaceResponsePayload` (so non-streaming routed responses would stop having namespaces restored), plus `injectSessionModelForSpawnCalls` and `SPAWN_MODEL_TOOLS` from 33b6c62 and d90c1fd.

Verified live against current `main` (2941895) on DeepSeek V4 Pro: the model calls `js` in `mcp__node_repl` and `navigate` in `browser`. The reported issue is already fixed for anyone who runs `bin/update`.

One gap does remain on `main`: `flattenNamespaceTools` concatenates `namespace__tool` with no length cap, and OpenAI-compatible providers reject function names over 64 characters. Worth a separate focused change.

PR #151, opened 2026-08-10
feat: diagnose kimi oauth credential health without breaking refresh

Bruce-Yii:
## What

Diagnose the Kimi OAuth credential file with the refresh path in mind, instead of treating a past `expires_at` as a failure.

Follow-up to the discussion on #131: an expired access token is a **recoverable** state — `ensureFreshKimiOAuthToken()` refreshes it with the still-valid refresh token before forwarding — so status reporting must not tell the operator to re-login for that. The genuinely broken states are a **revoked tombstone** (the CLI wrote all-empty tokens with zero expiry metadata after a rejected refresh) and a structurally unusable file.

## How

`src/oauth-status.mjs` — new read-only `kimiOAuthHealth()` (return shape untouched for `kimiOAuthStatus()`, no behavior change for its callers):

| status | meaning |
|---|---|
| `ok` | live session |
| `stale` | access token past expiry, refresh token still valid → auto-refreshes on next request, **no action needed** |
| `revoked` | all-empty tokens + zero expiry metadata (the session reader's own tombstone shape) → `kimi login` genuinely required |
| `incomplete` / `invalid` / `missing` | no usable session on disk |

The metadata validation mirrors what the session reader (`validateToken` in `kimi-oauth-session.mjs`) accepts at request time — finite second-based `expires_at` and `expires_in` — so a status verdict can never diverge from runtime usability (the divergence #131's reviewer flagged).

`src/doctor.mjs` — the Kimi OAuth check now reports `fail` only for a revoked session; a stale credential reports `ok` with detail "access token expired; it refreshes automatically on the next request" instead of a misleading re-login instruction.

## Tests

New `test/oauth-status-health.test.mjs`, 7 cases: live → ok; expired access token → stale (not fail, fix says "No action needed"); revoked tombstone → revoked; missing file → missing; missing tokens / non-finite or absent expiry metadata → incomplete; invalid JSON → invalid.

`npm test` on this branch: **794 tests, 782 pass, 0 fail, 12 skipped**. `npm run check` passes.

### DIFF
--- a/src/doctor.mjs
+++ b/src/doctor.mjs
@@ -10,7 +10,7 @@ import { grokCliPreflight } from "./grok-cli.mjs";
 import { detectLegacyInstallations } from "./legacy-migration.mjs";
 import { PROVIDERS } from "./model-registry.mjs";
 import { grokOAuthStatus } from "./grok-oauth-status.mjs";
-import { kimiOAuthStatus } from "./oauth-status.mjs";
+import { kimiOAuthHealth, kimiOAuthStatus } from "./oauth-status.mjs";
 import {
   readMultiAgentSettings,
   subagentEligibleModels,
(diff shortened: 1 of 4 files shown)


Bruce-Yii:
Quick follow-up on the Windows CI failure: `test/doctor-kimi-oauth-health.test.mjs` cleaned up with `rmdirSync`, which only removes empty directories. On Windows the doctor subprocess leaves files under the temp state dir (launch-agents etc.), so both tests threw `ENOTEMPTY` in the `finally` cleanup — `npm test` was red on windows-latest.

Fixed in e4c5339: the temp-dir cleanup now uses `rmSync(..., { recursive: true, force: true })`. Test logic and assertions are untouched. `npm test` on this branch: **796 tests, 784 pass, 0 fail, 12 skipped**.

PR #169, opened 2026-08-12
fix: preserve account native model catalog

wousp112:
## What changed

- capture the account-aware native model catalog from Codex's local `models_cache.json`
- merge it with the bundled catalog so account visibility and account-only models win while bundled schema fields and bundled-only models remain available
- derive `base_instructions` from the account catalog's instruction template when the strict custom-catalog schema requires it
- fingerprint the account model list so server-side catalog changes invalidate the saved native snapshot even when the Codex binary version is unchanged
- cover account visibility, bundled fallback, account-only models, and cache invalidation with tests

## Why

The install and refresh paths currently request `--bundled-native`. The bundled catalog can lag the signed-in account catalog: in practice it marked `gpt-5.4-mini` hidden and omitted `gpt-5.3-codex-spark`, so enabling the router made native models disappear from the picker.

This is not picker-slot competition. `model_catalog_json` replaces the catalog Codex reads, so the router must preserve the signed-in account's native model list and use the bundled catalog only as a schema and availability fallback.

## Impact

Native account models remain visible alongside routed external models. Existing provider routing, native GPT authentication, vision bridging, and provider credential isolation are unchanged.

## Validation

- `npm run check`
- `node --test test/catalog.test.mjs test/refresh-catalog.test.mjs` — 38 passed
- live catalog refresh and `doctor` on macOS — all core checks passed
- verified the resulting picker catalog lists `gpt-5.4-mini`, `gpt-5.3-codex-spark`, and routed models together

The complete `npm test` run reported 897 passed, 2 failed, and 6 skipped. The two failures are unrelated/environmental: a Kimi OAuth temporary-directory cleanup race that passes in isolation, and a POSIX hidden-prompt test that requires a real interactive TTY and also fails on an untouched checkout.

### DIFF
--- a/src/catalog.mjs
+++ b/src/catalog.mjs
@@ -1,3 +1,4 @@
+import { createHash } from "node:crypto";
 import {
   existsSync,
   mkdirSync,
@@ -15,6 +16,7 @@ import {
   ANNOUNCED_MODELS_PATH,
   CONFIG_PATH,
   MERGED_CATALOG_PATH,
+  MODELS_CACHE_PATH,
   NATIVE_ALIAS_PATH,
   NATIVE_CATALOG_PATH,
 } from "./paths.mjs";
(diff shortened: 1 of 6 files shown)


duolahypercho (changes requested):
Thanks — this is a real bug and it reproduces on my install today (bundled catalog hides `gpt-5.4`/`gpt-5.4-mini` and omits `gpt-5.3-codex-spark` while the account cache lists all three). The read-through + fingerprint-invalidation design is right. Three asks before undrafting:

1. **Blocker — `{{ personality }}` placeholder leaks into `base_instructions` for account-only models.** The bundled `base_instructions` is the instruction template *with* `{{ personality }}` substituted out; the raw `instructions_template` in `models_cache.json` still contains the literal placeholder. Shared slugs are safe (bundled `base_instructions` survives the merge), but account-only models — including `gpt-5.3-codex-spark`, the model this PR exists for — get the literal `{{ personality }}` token injected into their system prompt. Please substitute the `instructions_variables` defaults (or strip the placeholder), add a regression test asserting no `{{ ... }}` tokens survive into `base_instructions`, and include evidence of how Codex consumes `base_instructions` vs `model_messages` from a custom catalog — the "strict parser requires base_instructions" claim is currently asserted in a comment but never tested.

2. **`--bundled-native` becomes a silent no-op.** The PR removes the flag's effect in `src/catalog.mjs` but `src/refresh-catalog.mjs` still passes it and `test/refresh-catalog.test.mjs` / `test/doctor-routing-mode.test.mjs` still assert it. Remove it from all callers and tests in this PR — no dead flags.

3. **Minor:** compute the fingerprint from the already-parsed cache object instead of re-reading `models_cache.json` (double read is a TOCTOU window and a wasted ~215KB parse).

Nice-to-haves: dedupe duplicate slugs in the account list; note that `captured_with` stamps the installed binary version even though the cache may have been fetched by a different client version (fingerprint invalidation mostly covers this).

Branch merges cleanly onto current main and the new tests pass here (39/39, 40/40 post-merge). Happy to merge once item 1 lands.

duolahypercho:
Pushed `1131825` to this branch (maintainer edit) addressing all three review items so this can move:

1. **Placeholder blocker** — `base_instructions` is now derived by substituting `instructions_variables` defaults into the template, stripping any placeholder that has no default. Evidence for the substitution rule: for every shared slug that carries variables (`gpt-5.4`, `gpt-5.4-mini`, `gpt-5.5`), the bundled catalog's `base_instructions` is byte-identical to the account template with `{{ personality }}` → `personality_default` applied — so this mirrors exactly what Codex ships. Regression test asserts no `{{ ... }}` token survives into any merged `base_instructions` (including via a default that itself contains a placeholder). Verified against a live account cache: all 9 models derive clean, and account-only `gpt-5.3-codex-spark` gets 11,949 chars of placeholder-free instructions.
2. **`--bundled-native`** — removed from `src/refresh-catalog.mjs`, `test/refresh-catalog.test.mjs`, and `test/doctor-routing-mode.test.mjs`. No dead flags remain.
3. **Double read** — `readModelsCache()` reads and parses `models_cache.json` once; the fingerprint is computed from the same parsed snapshot that gets merged.

Also took the dedupe nice-to-have (first occurrence wins). Full suite: 903/903 pass on the branch, and it merges cleanly onto current main with catalog tests green post-merge.

@wousp112 thanks for the report and the fix — the merge design was right, we just hardened the instruction derivation.

PR #170, opened 2026-08-12
feat(tray): add custom model orchestrator mode control

calumcodes96:
## Summary
- adds a safe Custom Model / Orchestrator Mode switch to the native macOS tray
- assigns native Sol/Luna/Terra and external Flash/V4 Pro/GLM/Kimi roles without routine Kimi validation
- adds explicit ChatGPT weekly and OpenCode Go rolling/weekly/monthly usage cards with honest unavailable states
- packages the tray as **Open Router Model Router**

## Verification
- `swift build -c release`
- tray rebuild tests: 15/15
- `plutil -lint` and `codesign --verify --deep --strict` on the installed app

## Limitation
- Swift XCTest cannot run in this host toolchain because the XCTest module is unavailable.

### DIFF
--- a/apps/macos/ModelRouterTray/Package.swift
+++ b/apps/macos/ModelRouterTray/Package.swift
@@ -13,5 +13,10 @@ let package = Package(
       path: "Sources",
       resources: [.process("Resources")]
     ),
+    .testTarget(
+      name: "ModelRouterTrayTests",
+      dependencies: ["ModelRouterTray"],
+      path: "Tests"
+    ),
   ],
 )
--- a/apps/macos/ModelRouterTray/Resources/Info.plist
+++ b/apps/macos/ModelRouterTray/Resources/Info.plist
@@ -3,13 +3,13 @@
 <plist version="1.0">
 <dict>
   <key>CFBundleDisplayName</key>
-  <string>Model Router</string>
+  <string>Open Router Model Router</string>
   <key>CFBundleExecutable</key>
   <string>ModelRouterTray</string>
   <key>CFBundleIdentifier</key>
   <string>io.github.codex-router.tray</string>
   <key>CFBundleName</key>
-  <string>Model Router</string>
+  <string>Open Router Model Router</string>
   <key>CFBundlePackageType</key>
   <string>APPL</string>
   <key>CFBundleShortVersionString</key>
--- a/apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
+++ b/apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
@@ -56,6 +56,152 @@ enum RouterActivityState: String, Decodable {
   }
 }
 
+
+enum CodexMode: String {
+  case customModel
+  case chatGPT
+}
+
+struct CodexModeController {
+  private static let routerProviderLine = "model_provider = \"codex-router\""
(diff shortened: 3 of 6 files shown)


duolahypercho:
Evaluated this against current main before deciding what to do with it. The mode-switch implementation here is genuinely careful (timestamped 0600 backups, atomic replace, rollback + post-write verification), and a one-click flip between routed custom models and native ChatGPT is the right instinct. But main has moved a long way since this branched (46 commits, 13 touching the tray), and each of the PR's goals has since shipped through a different mechanism:

- **Mode switching**: the tray's "Use Router with ChatGPT" (`control signed-routing on/off`) and "Use without OpenAI login" toggles cover the custom-model/ChatGPT flip, routed through `src/config-manager.mjs`, which owns all `config.toml` mutation (backups, mode guards, and the recent provider-grouping/GLM-5.3 changes). Rewriting the config directly from Swift bypasses those guards — all mutations need to keep going through `control`.
- **Subagent roles**: hardcoded `default_subagent_model` swaps were superseded by the subagent modes in `src/multi-agent-state.mjs` (`all`/`selected`/`proven`) with per-model and per-provider tray toggles, plus the v2-proof requirement.
- **Usage cards**: OpenCode Go Rolling/Weekly/Monthly windows are now emitted by the router (`opencodeGoUsageMetrics` in `src/provider-account-usage.mjs`) and rendered by the existing cards with unavailable/reconnect states. The ChatGPT card kept both the 5-hour and weekly windows, so the weekly-only reduction would be an information regression.
- **Swift tests**: a Swift Testing target now exists at `apps/macos/ModelRouterTray/Tests/`, which the XCTest target here would collide with.

Two smaller notes for the record: the health-poll port change (4202→4102) points at the legacy port — 4202 is the current default in `src/paths.mjs` — and the "Open Router Model Router" rename collides with the app identity that shipped with #204 (and reads as an OpenRouter affiliation the project doesn't have).

A rebase would effectively be a rewrite against a restructured, localized tray file, so this one isn't salvageable as-is. The piece that would travel well is the mode-transition **verification** idea — confirming `control --json` + `service status` agree with the selected mode after a toggle — rebuilt on top of the existing signed-routing flow, through `control` rather than direct config writes. A small PR doing just that would be very welcome. Thanks for the thorough work here.

duolahypercho:
Closing as superseded — the full analysis is in the earlier comment, but the short version: everything this PR set out to do has since shipped through different mechanisms, and main has moved too far for a rebase to be anything but a rewrite.

- The custom-model/ChatGPT mode flip lives in the tray's signed-routing and login-free toggles, routed through `config-manager` so every config.toml mutation keeps its guards and backups.
- Subagent role assignment became the automatic verify-then-promote system (#225): toggling any model on researches it live and promotes it per machine, which goes further than the hardcoded `default_subagent_model` swaps here could.
- The OpenCode Go rolling/weekly/monthly usage windows now come from the router itself (`provider-account-usage`) and render in the existing tray cards.
- A Swift Testing target exists where this PR's XCTest target would collide.

The one idea from here still worth building — post-toggle mode verification, confirming `control --json` and `service status` agree with the selected mode after a flip — would be very welcome as a small fresh PR on top of the current signed-routing flow. Thanks for the careful engineering on the backup/rollback path; that part was genuinely well done.

PR #181, opened 2026-08-13
Add Simplified Chinese desktop i18n and selectable token units

Yuxin-Qiao:
## Summary

- add a persistent `Full` / `M` selector to the macOS tray's daily token usage section
- format the chart hover text, primary metric, and range summary with the selected unit
- add Simplified Chinese and accessibility labels, and document the new control

## Why

Large daily token counts are difficult to scan in the usage chart. The selector keeps the exact grouped value available while offering a compact millions view such as `174.6M`.

## Validation

- `swift build` in `apps/macos/ModelRouterTray`
- `npm run check`
- `node --test test/desktop-ui.test.mjs test/tray-rebuild.test.mjs`
- `git diff --check`

The full `npm test` run completed with 969 passing tests; 8 unrelated tests remain blocked by the checkout's missing `proper-lockfile` dependency.


duolahypercho:
Reviewed this in depth — it's the closest to mergeable of the open PRs, and the i18n key parity, placeholder handling, and token-unit arithmetic all check out. `swift build` clean, `node --test test/desktop-ui.test.mjs` 8/8. The `apps/desktop/ui/i18n.mjs` catalogs are at **perfect parity**: 200 keys each, no missing, no extra, no used-but-undefined. Nice work.

Three things before this lands.

### 1. One missing translation key (HIGH)

`ModelRouterTrayApp.swift:4327` — `.accessibilityLabel(routerLocalized("Fix Codex Router installation"))` has no entry in `RouterChineseText.values`. Its sibling on 4317 does, so a zh-CN VoiceOver user hears every control in Chinese except this one.

```diff
--- a/apps/macos/ModelRouterTray/Sources/Localization.swift
+++ b/apps/macos/ModelRouterTray/Sources/Localization.swift
@@
     "Update and verify Codex Router": "更新并验证 Codex 路由",
+    "Fix Codex Router installation": "修复 Codex 路由安装",
     "Running Codex Router maintenance": "正在维护 Codex 路由",
```

Matches the surrounding terse accessibility-label style. `修复 Codex 路由安装问题` if you prefer it more explicit.

Second gap, cosmetic: `ModelRouterTrayApp.swift:3055` is `Text(routerLocalized("CODEX"))`, the first column header of the local-models table. Its neighbours are translated, so zh-CN renders `CODEX | 模型 | 大小`. Defensible as a product name — your call, and fine to leave.

Those are the **only** two gaps across all 232 `routerLocalized`/`routerFormat` literals. Printf-specifier parity across all 292 dict entries is clean.

### 2. `zh-Hans.lproj/Localizable.strings` is unreachable (HIGH)

It ships but nothing can read it. Three separate reasons:

- Zero `NSLocalizedString` / `LocalizedStringKey` / `Bundle.localizedString` / `String(localized:)` calls anywhere in the four Swift sources. Translation goes entirely through your `routerLocalized` dict.
- Every `Text(routerLocalized("…"))` resolves to `Text.init<S: StringProtocol>(_:)`, which does **not** localize. The 16 genuine `Text("…")` literals that exist (`Text("·")`, `Text("\(percent)%")`) have no keys in the file anyway.
- Build output shows it landing in `ModelRouterTray_ModelRouterTray.bundle/zh-hans.lproj/` — a nested SPM resource bundle, not `Bundle.main`. So even a future `NSLocalizedString` with the default bundle would miss it.

All 87 entries are a byte-identical subset of the 292 in `RouterChineseText.values` — already 205 entries stale on arrival, with nothing to keep them in sync.

I'd **delete rather than wire it**. Wiring means rewriting ~232 call sites, and it would lose capability: `routerLocalized` is called with non-literal arguments at 16 sites (ternaries, `model.fit`, `operation.kind.label`) that `.strings` tooling can't extract.

```
git rm -r apps/macos/ModelRouterTray/Sources/Resources/zh-Hans.lproj
```

`defaultLocalization: "en"` in `Package.swift` exists only to satisfy that resource — SwiftPM hard-errors without it when localized resources are present. Optional to revert, but it's meaningless once the `.lproj` is gone.

Flagging rather than pushing this myself, since you deliberately authored the file and the reasoning isn't obvious from reading it.

### 3. Worth adding here: a parity test for the Swift dict

Nothing in `test/` references `Localization.swift`, `routerLocalized`, or `RouterChineseText` — which is exactly how item 1 shipped. A ~30-line Node test that regex-extracts every `router(Localized|Format)\(\s*"…"` literal and asserts a dict entry exists would have caught it, and would keep the 292-entry dict from rotting on every tray change. `test/tray-rebuild.test.mjs` already establishes the source-assertion convention for Swift, since there's no Swift test target.

If you add a placeholder-parity check for the JS side too, assert **zh ⊆ en** rather than exact equality. Exact equality fails on `models.subagentSummary`, where dropping `{plural}` is correct — Chinese has no plural marker. Subset still catches a real typo like `{cont}`.

### Heads-up on merge order

`git merge-tree` reports clean, but that's misleading. PR #185 adds ~183 lines to `ModelRouterTrayApp.swift` including ~12 new user-facing strings (`"Context savings"`, `"Old tool results replaced with receipts"`, `"24H"/"7D"/"30D"`, and an `.accessibilityLabel`) that are neither wrapped in `routerLocalized` nor in the dict. Different regions, so git merges them silently — and the zh-CN panel ships partly in English. Whichever lands second needs a follow-up pass. That's on me to fix in #185, not you; noting it so it doesn't surprise either of us.

duolahypercho:
This shipped in #194 — thank you, the translation work here is the bulk of what landed.

Everything in both of your commits is on `main`, verified feature by feature:

| your work | on main |
|---|---|
| `RouterChineseText` / `routerLocalized` — the 292-entry tray + Island catalog | ✅ |
| `apps/desktop/ui/i18n.mjs` — desktop panel i18n, 200 keys at exact parity | ✅ |
| `TokenDisplayUnit` + `TokenDisplayUnitPicker` — selectable token units | ✅ |
| `tray_text` — the Rust tray menu | ✅ |

The selectable token units came across intact: the picker sits beside the range picker in the usage header, persists under `ModelRouterTray.tokenDisplayUnit`, and feeds both the chart and the totals.

## What changed on the way in

**A language setting.** Your comment said you deliberately followed macOS's preferred language "without adding a second language setting." That's a defensible call, and I went the other way after using it: on a Mac reporting `["en-US", "zh-Hans-US"]`, a Chinese speaker has no way to ask for Chinese, and the reverse case is just as common. There's now a dropdown — `System · English` / `English` / `中文` — with `System` as the default, so your behavior is still what an untouched install gets. Each option is labelled in the language it selects, so someone who lands in the wrong one can still read their way out.

**One missing key.** `routerLocalized("Fix Codex Router installation")` (`ModelRouterTrayApp.swift:4327`) had no catalog entry, so a zh-CN VoiceOver user heard every control in Chinese except that one. Added, plus a test that now checks every `routerLocalized`/`routerFormat` literal in the sources has a translation — nothing checked that before, which is how it slipped through. Only two literals were missing across 232 call sites, which is a good hit rate for a catalog that size.

**Removed `Sources/Resources/zh-Hans.lproj/Localizable.strings`.** This one's non-obvious, so it's worth explaining rather than just doing: SwiftPM compiles it into a *nested* resource bundle (`ModelRouterTray_ModelRouterTray.bundle/zh-hans.lproj/`), not `Bundle.main`, and the tray makes zero `NSLocalizedString` calls — every string goes through your dictionary. So it shipped but was unreachable, duplicating 87 of the 292 entries with nothing keeping the copies in sync. `defaultLocalization` went with it, since it existed only to satisfy that resource.

**Merge conflicts resolved by combining.** `main` had moved four commits. Both conflicts kept both sides: #185's live savings summary is preserved with its static copy localized, and #188's island-mode migration was kept over yours — both make the overlay opt-in, but #188's also keeps it for an install that already had it rather than retiring it silently on update. Nice convergence, incidentally: you'd independently reached the same opt-in conclusion.

Closing as superseded. The catalog is yours and it's what people will actually read.

PR #183, opened 2026-08-13
Enable DeepSeek V4 Flash as a v2 subagent model

Yuxin-Qiao:
## Summary

- mark `deepseek/deepseek-v4-flash` as a proven native v2 subagent model
- add a registry regression assertion so the marker cannot silently disappear

## Compatibility evidence

Authorized live probe through the installed router (`./bin/test-model 'deepseek/deepseek-v4-flash' --live --yes --json`):\n\n- basic response: HTTP 200, marker received\n- streaming: stream text and completion event verified\n- tool calling: function call and JSON arguments verified\n- compaction: compaction response verified\n\nThe PR registry generated `router-model-deepseek-deepseek-v4-flash.toml` with `model = "deepseek/deepseek-v4-flash"`; a direct Codex CLI request using that temporary catalog returned `DEEPSEEK_NATIVE_CHILD_OK`. The existing routed payload relay and follow-up regression coverage is green.\n\n## Checks\n\n- `git diff --check`\n- `npm run check`\n- focused registry/catalog/routing suite: 117 passed\n- full suite: 967 passed, 8 environment failures because this shallow checkout lacks `proper-lockfile` (the failures are in existing Kimi OAuth/account/lock/quota/startup tests)\n\nThe full native desktop spawn/follow-up probe remains to be confirmed in a Codex build that exposes the receiver thread to the CLI; no model behavior failure was observed.

### DIFF
--- a/config/deepseek/deepseek-v4-flash.json
+++ b/config/deepseek/deepseek-v4-flash.json
@@ -7,6 +7,7 @@
       "upstreamModel": "deepseek-v4-flash",
       "provider": "deepseek",
       "listed": true,
+      "multiAgentVersion": "v2",
       "displayName": "DeepSeek V4 Flash (API)",
       "description": "DeepSeek V4 Flash via the DeepSeek API; fast reasoning, tool calls, and a 1M context window.",
       "priority": 6,
--- a/test/registry.test.mjs
+++ b/test/registry.test.mjs
@@ -241,6 +241,7 @@ test("provider registry exposes configured API and OAuth model families", () => 
     assert.equal(model.multiAgentVersion, "v2", model.slug);
   }
   assert.equal(MODEL_BY_SLUG.get("grok-oauth/grok-4.6").multiAgentVersion, undefined);
+  assert.equal(MODEL_BY_SLUG.get("deepseek/deepseek-v4-flash").multiAgentVersion, "v2");
   assert.equal(MODEL_BY_SLUG.get("deepseek/deepseek-v4-pro").multiAgentVersion, undefined);
   for (const slug of [
     "kimi-oauth/kimi-for-coding-highspeed",


CCanxue:
Data point: DeepSeek v2 subagent delivery works reliably once the task arrives as a plain user message. For non-OpenAI parents the value stored in `encrypted_content` is plaintext, so either mapping it to text at the router or delivering plain user input in the client works (we verified the latter end-to-end). OpenAI-parent -> DeepSeek-child still carries real ciphertext and is not recoverable at a router boundary (#36376 / #37197). Patch and evidence: https://github.com/CCanxue/codex-deepseek-subagent-fix

duolahypercho:
Took a close look at bringing this up to date. The config change itself is the right shape, and main has not shipped a DeepSeek v2 promotion, so the idea is still novel — but this can't land as-is, and the blocker isn't the merge conflict.

**The conflict is trivial**: only `test/registry.test.mjs` conflicts — main refactored the v1-holdout assertions into an `unprovenForV2` set (see the "Deliberate v1 holdouts" comment there). The new assertion would move next to the existing `deepseek/deepseek-v4-pro` check.

**The v2 proof is incomplete per AGENTS.md**, which requires all four of: tool calls, encrypted subagent payload relay without disclosure, a marker-return spawn, and a same-thread follow-up. The live `test-model` probe covers tool calls, and `DEEPSEEK_NATIVE_CHILD_OK` via a direct CLI request is encouraging — but the PR itself notes the native desktop spawn/follow-up probe is unconfirmed, and @CCanxue's data point suggests the OpenAI-parent → DeepSeek-child encrypted relay does not survive the router boundary without a client-side patch that isn't part of this repo (openai/codex#36376 / #37197). Main recently tightened exactly this gate (`fix: require proof for routed v2 subagents`), so the full native probe transcript — spawn from a native Codex parent, marker return, same-thread follow-up — needs to be attached before the flag flips. If the encrypted-relay path genuinely can't work for OpenAI parents yet, the model should stay v1 until the upstream issues resolve.

One more thing to state explicitly when re-probing: at `priority: 6` this model enters Codex's small advertised spawn-override subset ahead of `grok-api/grok-4.5` (priority 8) — per AGENTS.md that displacement should be intentional, not incidental.

Leaving this as a draft until the probe evidence exists.

duolahypercho:
Closing this as superseded rather than rejected — the goal it was after has since shipped through a different mechanism, and the specific change here still lacks the proof it needs.

Since this was opened, the router gained automatic v2 capability research (#225): enabling any model as a subagent probes it live (forced tool call + streaming) on that user's own provider, advertises it as an experimental v2 child on a pass, and the first real spawn settles a durable machine-local proof. DeepSeek V4 Flash and Pro both verify and run as v2 subagents through this path today (after #227 taught the router that DeepSeek's thinking mode rejects a forced tool_choice — the same class of evidence this PR's probe work pointed at, so the investigation here genuinely fed the fix).

What this PR changes — a repo-wide `multiAgentVersion: "v2"` on the direct DeepSeek API route — still requires the full native-collaboration proof per AGENTS.md (marker-return spawn from a native parent, encrypted payload relay, same-thread follow-up), which the draft notes were never completed, plus the openai/codex#36376 relay question. If anyone completes that probe, the change itself is two lines and a fresh PR with the transcript attached will sail through.

Thanks for the careful writeup — the probe evidence and the priority-displacement question it surfaced both shaped how the automatic verifier ended up working.

PR #184, opened 2026-08-13
feat: add anonymous free model gateways

Yuxin-Qiao:
## Summary

- add `opencode-free` for OpenCode Console’s documented keyless free chat models
- add `kilo-free` for Kilo Gateway’s documented anonymous `:free` models
- keep both providers catalog-only: live `/models` discovery is filtered to the provider’s free subset, then users curate the models locally
- enforce fixed official endpoints, reject paid model IDs, strip credentials from upstream requests, expose anonymous status in setup/doctor/usage, and add desktop/macOS tray icons and notes

## Why

OpenCode Console and Kilo Gateway both document a free-model exception that does not require an API key. Existing `opencode-go`/Zen support is separate: OpenCode documents Zen as a signed-in/billing service, so it is not an anonymous route.

The providers’ free catalogs and limits are controlled upstream and may change. Kilo documents anonymous traffic as limited to 200 requests per hour per IP. The router therefore does not check in unstable model metadata, does not permit endpoint overrides, and shows router traffic when provider quota headers are unavailable.

## Validation

- `npm run check`
- `npm test` — 1022 passed, 7 skipped, 0 failed
- focused provider/registry/discovery/credential/onboarding/usage tests — 86 passed
- `git diff --check`
- `node --check apps/desktop/ui/app.js`
- `env HOME="$PWD/.tmp-home" CLANG_MODULE_CACHE_PATH="$PWD/.tmp-clang-module-cache" swift build --package-path apps/macos/ModelRouterTray`

No live model request was made, so this PR does not consume provider quota.


duolahypercho:
Reviewed this closely. The security-critical parts are genuinely well built, and I want to say what I verified rather than just what's wrong:

- No credential leakage: `resolveProviderCredential` returns `value: undefined` for anonymous providers, and `upstreamHeaders` explicitly skips `Authorization`/`x-api-key`, so the gateway's internal bearer never goes upstream.
- `ANONYMOUS_ENDPOINTS` is real enforcement, not documentation — hardcoded allowlist, `baseUrlEnv` override forbidden, `credential`/`keyless` metadata forbidden, `anonymousModelPolicy`/`anonymousNote` required.
- The free-model policy is enforced at **both** registry load and the user-curation overlay, closing the "curate a paid model id through the free route" loophole end to end.
- Support-bundle redaction correctly skips anonymous providers.

One blocking issue, and it isn't where it looks.

### Anonymous providers are enabled without consent

`src/provider-selection.mjs:90` and `src/setup.mjs:172` treat `authMode === "anonymous"` identically to `keyless`. But `keyless` is loopback-enforced — nothing leaves the machine. `anonymous` means the prompt, and whatever was pasted into it, reaches `console.opencode.ai` / `api.kilo.ai`. Result: both gateways arrive **pre-checked** in the guided picker, and a bare `./install.sh --auto` writes them into `enabled-providers.json` with no prompt at all. Your own test locks this in — zero credentials gives `["kilo-free", "local", "opencode-free"]`.

**The obvious fix doesn't work.** Reverting the `|| provider.authMode === "anonymous"` clause leaves the behavior unchanged, because this PR also changed `resolveProviderCredential()` to return a persistent marker for anonymous providers — so the `credentialStatus(provider, { persistent: true }).configured` fallback branch returns `true` anyway. The exclusion has to be explicit.

The underlying problem is that `configuredProviderIds()` answers two different questions that diverge for exactly one class:

| Question | Callers | Anonymous included? |
|---|---|---|
| "can this authenticate?" | `ensure-configured` gate, `catalog.mjs`, `selectedConfiguredListedModels()` | **yes** — otherwise someone who explicitly chose one is told to run setup and their curated models vanish |
| "may this be on when nobody asked?" | `bin/install` → `ensure-configured` no-file branch, `--providers configured`, non-guided setup | **no** |

So: add a second export rather than changing the first.

```js
// Which providers a run that was told nothing may switch on by itself.
// `configuredProviderIds()` answers "can this authenticate"; this answers "may
// this be turned on when nobody asked for it". They diverge for exactly one
// class. An anonymous provider needs no credential, so the first question says
// yes on a machine that has never heard of it — but unlike `local`, which
// serves from this machine, enabling it points the operator's prompts at a
// third party over the internet.
export function defaultProviderIds() {
  return configuredProviderIds().filter(
    (id) => PROVIDERS.get(id)?.authMode !== "anonymous",
  );
}
```

Then swap `configuredProviderIds()` → `defaultProviderIds()` at the three write sites (`enableProvider`, `disableProvider`, `ensure-configured`) and in `setup.mjs` / `setup-shared.mjs` `selectProviders()` — both the `"configured"` branch and the no-flag default. Leave `providerConfigured()` and `providers.mjs` `configured()` alone, or an explicitly-selected anonymous provider fails the install with a credential prompt.

Separately, the pre-checked box has its own cause — and **your own code already shows the intended design**. You added `anonymous: "no API key"` to `ACTION_LABELS` in `src/setup-ui.mjs` and then never emit it. `provider-onboarding.mjs` sets `entry.action = "ready"`, and the guided picker pre-checks every `ready` row:

```diff
         entry.kind = "anonymous";
-        entry.action = "ready";
+        // Not "ready": the guided picker pre-checks every `ready` row, and an
+        // internet endpoint must be chosen rather than arrive checked.
+        entry.action = "anonymous";
```

Nothing else consumes `action` — the desktop UI and Swift tray key off `kind`/`configured`/`anonymousNote` — so this is contained.

Tests: `test/provider-selection.test.mjs:58,61` keep their current values (they're the right answer to "can authenticate") and gain `defaultProviderIds()` assertions; `test/setup.test.mjs:50,52` revert to `["deepseek", "local"]`; `test/provider-onboarding.test.mjs:83` becomes `"anonymous"`. Worth adding the regression that would have caught this: `ensure-configured` with no selection file — exactly what `bin/install` runs — must write `["deepseek", "local"]`.

I applied this whole patch to a scratch copy of your tree and ran the suite: 1023 pass, plus `scripts-check`. The one failure (`test/skills-install.test.mjs:55`) fails identically on your unpatched branch — pre-existing, unrelated.

**Severity context, in fairness:** because both gateways ship zero preselected models, an auto-enabled provider produces no traffic until someone runs `bin/curate-models`. So this is not live exfiltration today. It's still an endpoint the operator never chose being written into their config and shown as connected — and the first curated model turns it into traffic.

### AGENTS.md

`AGENTS.md` is the canonical installation contract for both agent families and this PR doesn't touch it, so an installing agent has no way to learn these providers exist — and per the above, omitting them from `--providers` doesn't actually withhold them. Needs: the two ids added to the step-4 selectable list with an explicit "never select one on the user's behalf", and a short section defining `authMode: "anonymous"` against `keyless` (off-box vs on-box, configured-but-never-default, catalog-only). I'd rather write that wording myself since it's the shared contract — I'll follow up, you don't need to.

Everything else here I'd merge happily. Fix the consent path and this is good to un-draft.

duolahypercho:
Pulled this locally and ran both gateways anonymously against the real endpoints. Kilo behaves exactly as documented — every `:free` model tried answered 200 with no key.

**OpenCode did not, and the cause turned out to be the endpoint rather than the filter.** Five of the seven ids the filter admitted returned `401 Paid inference requests require an Authorization bearer token`:

```
big-pickle                200
mimo-v2.5-free            200
deepseek-v4-flash-free    401
nemotron-3-ultra-free     401
north-mini-code-free      401
laguna-s-2.1-free         401
ling-3.0-flash-free       401
```

`console.opencode.ai/inference/openai/v1` is the paid Console inference route. It lists the free-suffixed models in `/models` and then refuses them without a bearer token — so the ids were right and the host was wrong. The documented Zen base URL is `https://opencode.ai/zen/v1`, and there all seven answer 200 with no Authorization header:

```
big-pickle  deepseek-v4-flash-free  mimo-v2.5-free  hy3-free
nemotron-3-ultra-free  nemotron-3.5-lightning-free  laguna-s-2.1-free
```

The catalogs differ too: Zen carries `hy3-free` and `nemotron-3.5-lightning-free` where the Console listed `north-mini-code-free` and `ling-3.0-flash-free`. The Zen list is the one matching the published free-model docs.

No OAuth or API key is involved — anonymous is the documented path for this subset, and the key is what buys the *paid* Zen catalog. That is why `opencode-zen` already points at this same base URL with a credential: two providers, one endpoint, two auth postures.

I pushed two commits to this branch:

1. **`docs:`** a use-at-your-own-risk note in the README and in both `anonymousNote` strings, which are what the tray and guided setup actually display. Neither catalog carries a pricing field, so the suffix rule stays a heuristic the router cannot verify before a request — worth saying plainly. Also removes a paragraph this branch's merge commit duplicated in `AGENTS.md`.
2. **`fix:`** the endpoint change, with the live evidence in the commit message.

`npm run check` and `npm test` are green after both: **1185 passed, 0 failed, 8 skipped**.

Thanks for this — the isolation work (endpoint pinning that fails the registry load, `variantOf`-free credentials, anonymous providers excluded from `defaultProviderIds`) is careful, and it all held up under review.

PR #186, opened 2026-08-13
fix: resolve and spawn commands the way Windows requires

duolahypercho:
On Windows, a command you can *find* and a command you can *spawn* are two different things. `where.exe` lists the extensionless npm shim before the `.cmd` shim, and Node has refused to spawn a `.cmd` without a shell since the CVE-2024-27980 fix. Four copies of the same lookup helper took the first line regardless, and every spawn failure that followed was attributed to something other than the spawn.

`src/codex-binary.mjs` already carried the fix for Codex (#46). It never propagated to the other four copies.

## The most visible symptom

A healthy npm-installed **Grok CLI could not be launched**, which raises `spawn UNKNOWN` — the same code Smart App Control raises. `UNKNOWN` is in `WINDOWS_BLOCK_CODES`, so the preflight concluded that Windows application control had blocked the CLI and told the operator to abandon OAuth for an API key. The token refresh behind that preflight spawned the shim the same unusable way, so passing the preflight would not have helped either.

`docs/TROUBLESHOOTING.md` documents this as a Windows policy problem. It was substantially self-inflicted; that section is updated.

## Same root cause, other call sites

| Where | Effect on Windows |
|---|---|
| `agent-check.mjs` | The routed-subagent proof from `c15e3a8` graded **every** routed model as not agent-capable |
| `doctor.mjs` config probe | Hit the error path and returned "no complaint" — a check that silently always passed |
| `grok-oauth-session.mjs` | Token refresh failed, expiring a session the preflight had just called healthy |
| `setup-shared.mjs` / `provider-onboarding.mjs` | `npm install -g` and CLI sign-in resolved `npm`/`kimi` to the shim and failed |

## Three standalone bugs

- **`codex-account-usage.mjs`** kept a private search for Codex — an undocumented `CODEX_BINARY`, a macOS-only path, then the bare name `"codex"`. None of the three exists on Windows, so the usage panel reported *"the Codex app-server could not be started"* on every Windows machine. It also killed only the shell on timeout, leaking a Codex process per poll.
- **`control.mjs` `apply`** ran the POSIX `bin/enable` shell script. `doctor --fix` and `curate-models` already branch to `install.ps1`; this now does too.
- **The vision-download worker** was the one detached child without `windowsHide`, opening a console window for the length of a multi-gigabyte pull.

Separately, `vision-host.mjs` probed for Ollama through the runtime's known install locations but then spawned the bare name — so a Windows install under `%LOCALAPPDATA%` reported as available and failed at the very next call.

## Approach

Resolution and launching now live together in `src/spawnable-command.mjs`, because a caller that gets the path right also needs to get the launch right. Arguments are escaped for `cmd.exe` (cross-spawn's algorithm) rather than joined on spaces — Node's `shell: true` would have split the agent-check prompt into eighteen arguments, and a path under `C:\Program Files` into two.

## Testing

Suite is green: **1038 pass, 0 fail** (1019 before; +19 new tests).

`test/windows-parity.test.mjs` guards all three classes from the source — no file but `spawnable-command.mjs` may call `where.exe`, no `bin/` script is spawned without a win32 branch, every detached spawn sets `windowsHide`. I verified each guard fails against the original code.

**What is still unverified:** I could not execute `cmd.exe`, so the escaping is checked by unit-testing the rendered command line against the reference implementation's output, not by running it. The Windows CI leg won't exercise a `.cmd` shim either. The outstanding check is a real npm-installed Codex or Grok CLI on Windows:

```powershell
npm install -g @xai-official/grok
./codex-router.ps1 doctor          # Grok CLI should read ready, not blocked
./model-router.ps1 codex control account-usage   # should return usage, not an app-server error
```

https://claude.ai/code/session_01Ay8brnrizV8xfbB322f1w8

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ay8brnrizV8xfbB322f1w8)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,43 @@
 
 ## Unreleased
 
+- **Windows stops mistaking its own spawn failures for provider problems.** A
+  command resolved on Windows and a command Windows can spawn are two different
+  things: `where.exe` lists the extensionless npm shim first, and Node has
+  refused to run a `.cmd` shim without a shell since CVE-2024-27980. Four
+  copies of the same lookup helper took line one anyway, and the spawn errors
+  that followed were each read as something else. The official Grok CLI was the
+  worst of it — a healthy npm install failed to launch, raising the same
+  `spawn UNKNOWN` that Smart App Control raises, so the router announced that
+  Windows application control had blocked it and told the operator to give up
+  on OAuth and use an API key. Even had the preflight passed, the token refresh
+  behind it spawned the shim the same unusable way.
+
+  Resolution and launching now live in one module and are used everywhere:
+  the routed-subagent proof run (which hands Codex a whole sentence, so it
+  cannot go through a shell that joins arguments on spaces), the Codex account
+  usage panel, the doctor's Codex configuration probe, the `npm install -g` and
+  sign-in paths for provider CLIs, and the precedence probe. Arguments are
+  escaped for `cmd.exe` rather than concatenated, so a path under
+  `C:\Program Files` and a prompt containing spaces both survive.
+
+- **Three more Windows-only breakages in the same family.** The Codex account
+  usage panel kept a private two-line search for the CLI — an undocumented
+  environment variable, a macOS-only path, then the bare name — which finds
+  nothing on Windows, so the panel reported "the Codex app-server could not be
+  started" on every machine; it now uses the shared discovery and kills the
+  process tree rather than leaking a Codex process per poll. `control apply`
+  ran the POSIX `bin/enable` script, which Windows cannot execute, and now
+  takes the PowerShell installer that `doctor --fix` already uses. The
+  vision-model download worker was the one detached child without
+  `windowsHide`, so it opened a console window for the length of a
+  multi-gigabyte pull.
+
+- **A local Ollama the router could find is one it can also run.** The vision
+  host probed for Ollama through the runtime's known install locations but then
+  spawned the bare name, so a Windows install under `%LOCALAPPDATA%` reported
+  as available and failed at the next call. Both now use the same resolver.
+
 - **Routed models that emit integer tool arguments as JSON floats no longer
   get those calls rejected by Codex.** Grok 4.6 was sending
   `timeout_ms: 20000.0`; Codex's native `shell_command` schema wants a `u64`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+- **Windows installs the tray companion, and keeps it.** Nothing on Windows
+  ever built or started it: `install.ps1` had no tray option at all, the
+  installer's own decision helper excluded the platform outright, and
+  `control tray enable` answered `{"supported":false}` and exited 0 — a silent
+  no-op that reads as success while no tray was ever going to appear. The only
+  route was knowing to run `scripts/build-desktop-tray.ps1` by hand, and even
+  then the companion vanished at the next reboot. `install.ps1 -WithTray` (and
+  `-NoTray`, matching `install.sh`) now builds it and registers a `Codex Router
+  Tray` logon task, kept separate from the router's own task so stopping one
+  never takes the other down. Quitting from the tray menu stays quit: the
+  restart setting covers a crash, not a clean exit. A platform with no
+  supervisor now says so on stderr rather than reporting success, and the
+  guidance names the `^` overflow that hides new tray icons on Windows 11.
+
 - **Windows stops mistaking its own spawn failures for provider problems.** A
   command resolved on Windows and a command Windows can spawn are two different
   things: `where.exe` lists the extensionless npm shim first, and Node has
--- a/README.md
+++ b/README.md
@@ -1020,10 +1020,16 @@ surface.
 ```
 
 ```powershell
-# Windows PowerShell
+# Windows PowerShell -- build, launch, and start at logon
+.\install.ps1 -CheckoutInstall -WithTray
+
+# or build and launch it by hand
 .\scripts\build-desktop-tray.ps1 -BinaryOnly
 Start-Process .\apps\desktop\src-tauri\target\release\codex-router-desktop.exe
 ```
+
+Windows 11 hides new tray icons in the `^` overflow next to the clock; drag the
+icon onto the taskbar to pin it.
 
 Windows and Linux on X11 receive the floating top-center activity pill. Linux
 on Wayland uses the tray panel without the pill because the compositor owns
--- a/docs/DESKTOP-TRAY.md
+++ b/docs/DESKTOP-TRAY.md
@@ -91,6 +91,28 @@ Start-Process .\apps\desktop\src-tauri\target\release\codex-router-desktop.exe
 Pass `-BinaryOnly` for an unbundled executable. Installer artifacts are written
 under `apps\desktop\src-tauri\target\release\bundle` by a full build.
 
+## Starting at logon
+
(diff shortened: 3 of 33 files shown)


duolahypercho:
Pushed `635e1c9`. Two things resolved, one needs your call.

## Windows CI failure — fixed, and it was a real bug

The Windows leg rejected the command line outright:

```
cmd.exe /d /s /c "...\codex-doctor.cmd ^^^"debug^^^" ^^^"models^^^""
The syntax of the command is incorrect.
```

I was double-escaping `^` for every batch file. Only one shim shape needs that second layer — the cmd-shim npm generates under `node_modules/.bin`, which re-enters `cmd.exe` to reach the real program, so the first interpreter consumes a layer before the second sees it. The global npm shim this router actually meets (in `%APPDATA%\npm`) reaches its program without a nested interpreter and takes one layer.

**This is exactly the gap I flagged as unverified when I opened the PR, and the reason was structural:** every escaping test asserted the *shape of a command line nobody runs*, so a plausible-looking string sailed through. Two Windows-only tests now build a real shim — a plain batch file and npm's actual global template — push arguments containing spaces, quotes, `&`, `%` and `^` through it, and read back what the child received. Those would have failed on the first push.

## CodeQL — two of the three alerts are mine

`js/shell-command-injection-from-environment` at `codex-binary.mjs:76`, `doctor.mjs:103`, `agent-check.mjs:129`.

- **`codex-binary.mjs` predates this PR.** `main` already built a shell command from the same env-derived path — `codexSpawnTarget` returned `` command: `"${binary}"` `` with `shell: true`.
- **`doctor.mjs` and `agent-check.mjs` are new here.** Both previously spawned directly with no shell; routing them through `cmd.exe` is what introduced the flow.

CodeQL is factually right: a shell command line is built from a path sourced from `CODEX_BIN` / `LOCALAPPDATA` / `PATH`. My assessment is that it is low severity, because anyone who can set those already chooses which program runs — injection buys an attacker nothing they don't already have.

That said, escaping is the wrong last line of defence for a character that cannot occur in a real path, so `spawnableCommand` now **refuses** a command path containing `"`, `<`, `>`, `|`, `?`, `*`, or a control character — no Windows file name may hold any of them, and they are precisely the ones that could end the quoted span or start a second command. Legal-but-hostile characters (`C:\Program Files (x86)`, `Foo & Bar`) keep being escaped, since rejecting those would break ordinary installs.

**What I can't do:** that guard is unlikely to satisfy CodeQL, which won't recognise a custom validator as a sanitiser. Clearing the three alerts needs a dismissal in the Security tab, and whether that's the right call is yours — I'd rather leave them visible than paper over them. Say the word if you'd prefer a different approach.

---

duolahypercho:
Correction to my previous comment: it's **four of five**, not two of three. The CodeQL run on `635e1c9` surfaced two more.

| Alert | File | New here? |
|---|---|---|
| 13 | `src/codex-binary.mjs:76` | **No** — `main` already did `` command: `"${binary}"` `` with `shell: true` |
| 16 | `src/agent-check.mjs:129` | Yes |
| 20 | `src/doctor.mjs:103` | Yes |
| 28 | `src/grok-cli.mjs:106` | Yes |
| 29 | `test/spawnable-command.test.mjs:162` | Yes — but see below |

Alert 29 is a different animal: it's the Windows end-to-end test I just added, and its "environment value" is `os.tmpdir()` — a directory the test creates with `mkdtempSync` two lines earlier. I'm not going to contort that test to satisfy the scanner; using `os.tmpdir()` is what every other test in this repo does, and writing shim fixtures into the working tree instead would be worse.

The wider point, now that the shape is clear: **this rule will fire on every site that runs a `.cmd` shim**, because building a command line for `cmd.exe` from a discovered path is what running one *is*. There's no code arrangement that both supports npm-installed CLIs on Windows and avoids the rule. So this isn't something I can drive to zero by editing — it's a policy call:

1. **Dismiss the five** with a "used safely" rationale, leaving the rule live for genuinely new sites.
2. **Filter the rule** for these paths via `.github/codeql/codeql-config.yml`. I can write that, but narrowing a security rule on your repo is your decision, not mine to make quietly.
3. **Leave them standing** as a permanent visible reminder of where the shell boundary is.

I'd lean 1 — it keeps the rule sharp for future code while recording the reasoning per site. Happy to do 2 if you'd rather, but I'm not going to suppress a security rule without you saying so.

---

duolahypercho:
Reviewed line by line, including checking out the branch and reading the Windows CI logs rather than trusting the diff. **Verdict: merge.**

### CodeQL triage — the five `js/shell-command-injection-from-environment` alerts

Decision: **accept, not defects.** Evidence, since "the scanner is wrong" deserves more than an assertion:

1. **This PR removes shell usage; it adds none.** The diff deletes `return { command: `"${binary}"`, options: { shell: true } };` from the old resolution path. Grepping the branch for `shell: true` in the touched files returns exactly one hit — a *comment* at `src/agent-check.mjs:107` explaining why the code avoids it. Net change to shell surface is negative.
2. **Every flagged call site is a plain argv spawn.** `agent-check.mjs:123`, `doctor.mjs:98`, `grok-cli.mjs:100`, `codex-binary.mjs:72` all pass `(target.command, target.args, options)` with no `shell` key. On macOS and Linux `spawnableCommand` is an identity pass-through, so those three platforms have no shell in the path at all.
3. **The one genuine shell hop is deliberate and bounded.** A `.cmd`/`.bat` shim on win32 must go through `cmd.exe` — Node has refused to spawn one directly since CVE-2024-27980. That path uses an explicit `cmd.exe /d /s /c` with `windowsVerbatimArguments: true` and the repo's own escaping, not `shell: true`, precisely because Node's shell mode joins argv on spaces and would split a prompt sentence into eighteen arguments.
4. **The tainted source is guarded before it reaches escaping.** `assertSpawnablePath` throws on `/["<>|?*\u0000-\u001f]/` — the characters no Windows filename may contain and the ones that could close the quoted span. Rejecting rather than escaping is the right call for a path.
5. **Proven by execution, not string shape.** `test/spawnable-command.test.mjs` builds real `.cmd` shims (plain batch and npm's actual global-shim template) inside a directory named `shim dir (x86) & co`, spawns them with `C:\Program Files (x86)\thing`, `a&b`, `100%`, `caret^value`, and asserts on what the child received via its own `argv`. Confirmed in the `windows-latest` log that both ran for real and the suite reported 1031 pass / 0 fail.
6. The taint sources are `CODEX_BIN`, `GROK_CLI`, `LOCALAPPDATA` — set by the same principal who already controls what gets spawned. Someone who can set `CODEX_BIN` can simply point it at their own binary; there is no privilege boundary being crossed.

Alert #29 (`test/spawnable-command.test.mjs:172`) is a test-fixture false positive on `os.tmpdir()`.

**Mechanical note:** these five sit on `refs/pull/186/head` with `state: null`, so they are PR-scoped and the dismissal API cannot act on them yet. They become dismissible branch alerts after merge. I'll dismiss them with this rationale then rather than leaving them to accumulate.

### Also verified

- No credential reaches argv — `XAI_API_KEY` is stripped from the environment before every Grok CLI spawn.
- The escaping is a faithful port of `cross-spawn`'s algorithm: same character class, same backslash-doubling before quotes, same conditional bare-passthrough for shims that compare `%1` textually, same double-escape for `node_modules/.bin` shims re-entering `cmd.exe`.
- POSIX behavior unchanged; ubuntu and macOS legs green.
- All five CI jobs green, including a real Windows test run.

### Follow-up, not blocking

`src/control.mjs` `runApply()`'s new Windows PowerShell branch has no test. The arguments match the pattern already proven in `doctor.mjs --fix`, so confidence is moderate — but it is literally unverified. Worth a unit test after this lands.

PR #191, opened 2026-08-14
ci: ship the desktop companion as a download, and manage it on Windows

duolahypercho:
The desktop companion could be obtained exactly one way: **install a Rust toolchain and compile it.** Releases shipped source archives only, and the CI job that already built the Windows and Linux binaries on every run threw them away. So there was no prebuilt tray anywhere, for anyone.

That is the reason "just run the tray" was never a real instruction.

## Ship the binary that already gets built

- **Releases** build the companion on Windows and Linux and attach `codex-router-tray-<version>-windows-x64.exe` and `codex-router-tray-<version>-linux-x64`. They are staged into `dist/` **before** the checksum step, so they land in `SHA256SUMS` and under the existing build-provenance attestation rather than arriving unverifiable. The Homebrew step reads its tarball line by name, so the extra entries do not disturb it.
- **CI** publishes the same binaries as artifacts, so unreleased changes can be tried without a toolchain. `if-no-files-found: error`, so a build that stops producing a binary fails the job instead of publishing an empty zip.

Windows 10/11 already ship the WebView2 runtime the companion needs, so a downloaded binary just runs.

## Manage it on Windows the way macOS is managed

#186 made the Windows tray installable but nothing else. `bin/model-router-tray` answered Windows with *"use scripts/build-desktop-tray.ps1"* and `codex-router.ps1` had no `tray` verb at all — so where macOS and Linux each have one command that builds the companion and hands it to a supervisor, Windows had two incantations and no supported way to check, restart, or remove it.

```powershell
.\codex-router.ps1 tray [install|status|start|stop|restart|uninstall]
```

Install consults the rebuild plan first and stamps the build afterwards. That stamp had nowhere to go: `TRAY_PLATFORMS` covered `darwin` and `linux` only, so `recordTrayBuild()` **threw** on win32 and `trayRebuildPlan()` answered `"unsupported"`. The one platform whose tray must be built deliberately was the one that never recorded having been built, so every update would have rebuilt it from scratch. Windows now mirrors the Linux definition, and the two deliberately share a fingerprint — one edit to the UI or the Rust does make both stale.

Guided setup runs the same command rather than repeating the build/stamp/register sequence, so the two cannot drift.

## Review follow-up from #186

`control.mjs`'s Windows apply branch was a hand-rolled PowerShell argument list with no test — flagged as non-blocking in the review. It now reuses `currentCheckoutInstaller()`, which `update` already uses and which is unit tested. The helper gained a `posixScript` option because Windows runs one installer for both entry points while POSIX has separate `bin/install` and `bin/enable`.

## Testing

**1079 pass, 0 fail.**

Two existing tests asserted the old premise — that Windows has no companion to keep in sync, and that its fingerprint is empty. Both are now false, so they were rewritten to the full lifecycle the other platforms already assert (absent → rebuild → skip → rebuild) rather than deleted, with an `aix` case added so "genuinely unsupported" stays covered.

**Unverified:** the Task Scheduler registration itself. No test or CI job runs `tray install` on Windows, because it would register a real logon task on the runner. The workflow changes are likewise only proven by running them — if `actions/upload-artifact@v4` is behind this repo's other pins (`checkout@v7`, `cache@v5`), CI will say so immediately.

https://claude.ai/code/session_01Ay8brnrizV8xfbB322f1w8

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ay8brnrizV8xfbB322f1w8)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,3 +141,17 @@ jobs:
       - run: npm ci --prefix apps/desktop
       - run: npm run check --prefix apps/desktop
       - run: npm run build:binary --prefix apps/desktop
+      # The companion needs a Rust toolchain and a multi-minute compile, which
+      # is a hard prerequisite for anyone who only wants to run it. This job
+      # already produces exactly the binary they need and was discarding it, so
+      # there was no way to obtain a prebuilt tray anywhere -- every user had
+      # to install Rust and build it themselves, or go without.
+      - name: Publish the tray binary
+        uses: actions/upload-artifact@v4
+        with:
+          name: codex-router-tray-${{ runner.os }}
+          path: |
+            apps/desktop/src-tauri/target/release/codex-router-desktop
+            apps/desktop/src-tauri/target/release/codex-router-desktop.exe
+          if-no-files-found: error
+          retention-days: 14
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,55 @@ permissions:
   attestations: write
 
 jobs:
+  # Releases shipped source archives only, so the desktop companion could be
+  # obtained exactly one way: install a Rust toolchain and compile it. Building
+  # it here turns that into a download.
+  desktop:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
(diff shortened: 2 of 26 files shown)


duolahypercho:
Superseded by #202, which already contains this PR's full commit history and is now the consolidated desktop companion + Windows tray parity change. Continuing review in #202.

PR #200, opened 2026-08-14
fix: release the empty-completion hold when the upstream proves it is generating

duolahypercho:
## The problem

`EmptyCompletionGuard` buffers every routed streaming response until it sees content, and reasoning deltas deliberately do not count as content. On a reasoning model the gap between the first reasoning delta and the first output token is seconds to minutes, so the guard held that entire gap. The client saw nothing until the turn closed or the 30s budget expired.

Measured against `opencode-go/deepseek-v4-pro`, two router processes sharing the same gateway, forwarder and provider, with only this behaviour varying and arms alternating so provider drift cancels:

| Model | Held | Released on liveness |
|---|---|---|
| `deepseek-v4-pro` | 30,638 / 30,531 ms | **517 / 526 ms** |
| `deepseek-v4-flash` | 29,409 / 29,577 ms | **536 / 503 ms** |

Both held arms park at `MAX_PRECONTENT_MS` — the hold budget, not the model, decided when the caller saw anything. `gpt-5.6-luna` showed no difference, because that upstream emits nothing until it emits text, so there was never a gap to hold.

Throughput is unaffected, which is why this never read as a slow turn — only a frozen one. It could not read as anything in telemetry either: `responseStartMs` stops at the response headers and `firstTokenMs` fires on reasoning deltas, and the guard kept holding past both. There was no field for the hold.

## What it cost, from 77,566 recorded turns

| | Count |
|---|---|
| Routed turns (guard applies) | 19,043 |
| Empty-completion retries fired | 168 |
| Retries that succeeded | **17** |

Every reasoning turn paid up to 30 seconds of dead air for a silent rescue on roughly one routed turn in a thousand. In the other 90% of empty turns the user got an error anyway, just 30 seconds later and after a second billed request.

## The change

The hold exists only to make the retry invisible. That is worth paying for when the upstream is silent — a silent upstream has no prologue worth waiting for — and not worth paying for when it is visibly generating.

- **`isLivenessEvent()`** — reasoning deltas, reasoning `.added`/`.done`, `output_item.added` carrying a reasoning item, and chat-completions `reasoning_content` / `reasoning`.
- Liveness **ends the hold without settling the verdict**. The guard relays and keeps parsing from behind the relay, so a turn that streams reasoning and then produces nothing is still classified empty.
- **`suppressedPrologue()`** — true only when every byte was still held at the moment the turn was declared empty. That is the precondition for an invisible retry, and it is what the router now branches on.
- A silent turn still holds every byte and still retries silently. Unchanged.
- When the attempt was already relayed, the router states the failure into the open stream via `writeStreamErrorEvent` rather than grafting a second response onto it.
- New `emptyCompletionUnrepairable` usage-event field. `recordUsageEvent` has an explicit field allowlist, so this had to be added there too or it is silently dropped.

One subtlety worth review: `#release()` no longer clears `#parseBuffer` on a liveness release. That buffer holds the partial SSE block straddling the release, and dropping it would corrupt the block the verdict may depend on.

## Why the visible error is not a regression

Verified against `codex-cli 0.145.0` with a stub upstream and an isolated `CODEX_HOME`, 3/3 reproducible:

| Stub response | Upstream turns | Answered | Codex retried |
|---|---|---|---|
| reasoning + text (control) | 1 | yes | — |
| reasoning + empty completion | 1 | **no** | **no** |
| reasoning + `event: error` | 2 | yes | yes |

The silent arm reproduces the original bug exactly: Codex prints the reasoning, reports tokens used, exits 0, and shows no answer and no error. The error arm makes Codex reissue the turn on its own ladder (`Reconnecting... 1/5`), so recovery is stronger than the single silent retry it replaces — up to five attempts instead of one — and visible instead of hidden.

Cosmetic wart: the reasoning summary renders twice in the error arm, once from the failed attempt and once from the retry. That is the visible seam that buys the 25 seconds.

## Test plan

- [x] `test/empty-completion-guard.test.mjs` + `test/empty-completion-router.test.mjs` — 52/52
- [x] `npm test` — 1161 pass, 0 fail, 8 skipped
- [x] Live A/B against `opencode-go/deepseek-v4-pro` through the real provider: first byte 30,638 ms -> 1,159 ms
- [x] Live telemetry confirms the verdict survives the release: patched runs record `emptyCompletion` + `emptyCompletionUnrepairable` where the unpatched run recorded only `emptyCompletionGuardReleased` with no verdict at all
- [x] Codex client behaviour verified with a stub upstream, 3 modes, 3 reps

Fixture split for review: `SILENT_TURN` (no liveness — suppressed and retried) vs `EMPTY_TURN` (reasoning — relayed and stated). The byte/time-budget fixtures were repointed off reasoning, since liveness now releases long before the byte cap and they were about to exercise the wrong path.

## Base branch

Stacked on `feat/harness-lifecycle-and-native-session` (#198) because that branch also modifies `src/router.mjs`. Retarget to `main` once #198 merges.

### DIFF
--- a/src/empty-completion-guard.mjs
+++ b/src/empty-completion-guard.mjs
@@ -60,10 +60,53 @@ function chunkHasChatContent(data) {
   });
 }
 
+// Liveness is proof the upstream is generating without being output the client
+// can act on. Reasoning is the canonical case, and it is why liveness exists as
+// a separate verdict: on a reasoning model the gap between the first reasoning
+// delta and the first output token is seconds to minutes, and holding that gap
+// is what turns a healthy turn into a frozen screen. Measured against
+// deepseek-v4-pro, the hold moved the client's first byte from 517 ms to
+// 30,638 ms — the byte/time budget, not the model, decided when the caller saw
+// anything. Liveness ends the hold. It deliberately does not count as content,
+// so the emptiness verdict below is unchanged: a turn that streams only
+// reasoning and then completes with nothing is still classified empty, it just
+// can no longer be repaired invisibly.
+function isLivenessEvent(eventType, data) {
+  if (typeof eventType === "string") {
+    if (/reasoning/.test(eventType)) {
+      if (/\.delta$/.test(eventType)) {
+        return typeof data?.delta === "string" && data.delta.length > 0;
+      }
+      // `.added`/`.done` on a reasoning part carry no delta of their own but
+      // still only appear once the model has started producing.
+      if (/\.(?:added|done)$/.test(eventType)) return true;
+    }
+    // Responses streams without summaries announce reasoning as an output item
+    // rather than as a typed reasoning event.
+    if (
+      eventType === "response.output_item.added" &&
+      data?.item?.type === "reasoning"
+    ) {
+      return true;
+    }
+  }
+  const choices = data?.choices;
+  if (!Array.isArray(choices)) return false;
+  return choices.some((choice) => {
+    const delta = choice?.delta ?? choice?.message;
+    if (!delta || typeof delta !== "object") return false;
+    // `reasoning_content` is DeepSeek's field; `reasoning` is what several
+    // OpenAI-compatible resellers relay instead.
+    const reasoning = delta.reasoning_content ?? delta.reasoning;
+    return typeof reasoning === "string" && reasoning.length > 0;
+  });
+}
+
 // Content means something the client can act on: output text or a tool call.
 // Reasoning deltas are deliberately not content — a turn that streams only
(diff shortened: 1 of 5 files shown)


duolahypercho:
Superseded by #205. This PR was auto-closed at 10:15:13 when its base branch `feat/harness-lifecycle-and-native-session` was deleted by the squash-merge of #198 — two seconds after that merge landed. It was not reviewed down.

The same commit (`9d5e87a`) is cherry-picked onto current `main` in #205 with no conflicts, and the suite is green there: 1166 passed, 0 failed, 8 skipped.

PR #209, opened 2026-08-14
fix(desktop): make Windows tray control spawns and task status work

### LINKED ISSUE #208: [Bug]: Windows tray: every refresh fails with ESDIR on stat C, console windows flash, and tray status reports stopped
> **About this report:** This bug was found and verified with the help of **DeepSeek V4 Flash** (the current official release, codename **DeepSeek-V4-Flash-0731**), which diagnosed the root cause and drafted the proposed fix below. The reporter is not a developer and is not familiar with the languages involved (Rust, Node.js, PowerShell); the local reproduction, system checks, and logs were performed by the reporter, while the technical analysis and fix were prepared by the model. Please review the technical claims with that context in mind.

## What happened

On Windows, every **Refresh** in the desktop companion fails. A refresh fires several `control` commands (snapshot, account, provider-usage, provider-setup, local-models), each spawning `node.exe <checkout>\src\control.mjs`. Every spawn dies at node startup before the script runs:

```
node:fs:2745
const out = binding.stat(base, false, undefined, true /* throwIfNoEntry */);
^
Error: ESDIR: illegal operation on a directory, stat 'C'
    at Object.realpathSync (node:fs:2745:25)
    at toRealPath (node:internal/modules/helpers:83:13)
    at Module._findPath (node:internal/modules/cjs/loader:80:22)
    at resolveMainPath (node:internal/modules/run_main:35:21)
    ...
```

## Root cause

`resolve_source_root()` in `apps/desktop/src-tauri/src/main.rs` finishes with `candidate.canonicalize()`. On Windows, `std::fs::canonicalize` returns an extended-length `\\?\` path, e.g.:

```
\\?\C:\Users\<user>\AppData\Local\codex-router
```

`run_control()` then spawns `node \\?\C:\...\src\control.mjs`. Node cannot execute a `\\?\`-prefixed path as a script entry point — it strips the prefix, resolves the bare drive letter, and fails at startup. Reproduced with `node "\\?\C:\Users\<user>\AppData\Local\codex-router\src\control.mjs" --json`.

## Secondary issues

- **Console windows flash on every control call.** The tray is a GUI process with no console, but the node.exe spawns use a plain `std::process::Command` without `CREATE_NO_WINDOW` (`0x08000000`). Each control call flashes a new terminal window; with every spawn failing instantly, this shows up as a constant stream of popups.
- **Tray status always reports "stopped" on non-English Windows.** `taskRunning()` in `src/tray-service-windows.mjs` matches `\bRunning\b` against localized `schtasks` output (Spanish shows "En ejecución", for example), so `tray status` never sees a running task even while it runs.

## Expected behavior

Clicking Refresh loads the panel data; no console windows appear; tray status reflects the real task state.

## Steps to reproduce

1. Windows 10/11, install the router and the tray (`install.ps1 -WithTray`).
2. Open the tray panel and click **Refresh** (or wait for the 60-second auto-refresh).

## Fix (local patch; happy to open a PR)

- Strip the `\\?\` prefix before handing the script path to node.
- Set `creation_flags(0x08000000)` (CREATE_NO_WINDOW) on the spawn.
- Read the task state through `Get-ScheduledTask`'s locale-independent `State` enum in `taskRunning()`.

## Doctor output

All doctor checks pass (exit 0). This is desktop-companion specific and does not affect the headless router.

richardotomislav:
Fixes #208

## What this fixes

On Windows, every refresh in the desktop companion app failed. A refresh fires several `control` commands, each spawning `node.exe <checkout>\src\control.mjs`, and every spawn died at node startup with:

```
node:fs:2745
const out = binding.stat(base, false, undefined, true /* throwIfNoEntry */);
^
Error: ESDIR: illegal operation on a directory, stat 'C'
```

This PR addresses three related Windows issues:

1. **`\\?\` extended-length paths break node spawns.** `resolve_source_root()` ends with `candidate.canonicalize()`, which on Windows returns a `\\?\C:\...` path. `run_control()` then spawns `node \\?\C:\...\src\control.mjs` — node cannot execute a `\\?\`-prefixed path as a script entry point and fails at startup (reproduced with `node "\\?\C:\...\src\control.mjs" --json`). The prefix is now stripped before the path is handed to node.

2. **Console windows flash on every control call.** The tray is a GUI process with no console, but the `node.exe` spawns did not set `CREATE_NO_WINDOW` (`0x08000000`), so every control call flashed a new terminal window.

3. **Tray task status always reported "stopped" on non-English Windows.** `taskRunning()` in `src/tray-service-windows.mjs` matched `\bRunning\b` against localized `schtasks` output (e.g. Spanish renders it "En ejecución"). It now reads the task state through `Get-ScheduledTask`'s locale-independent `State` enum, matching how `service-windows.mjs` already does it.

## Changes

- `apps/desktop/src-tauri/src/main.rs` — strip the `\\?\` prefix (`windows_readable_path`), set `CREATE_NO_WINDOW` on the spawn, add a regression test.
- `src/tray-service-windows.mjs` — locale-independent `taskRunning()`.

## Verification

- `cargo fmt --check`, `cargo check`, and `cargo test` pass (7 tests, including the new path-stripping test).
- `npm run check` and `npm test` pass; the 7 local failures are environmental (4 POSIX installer tests require `sh`, which is not on this session's PATH; 2 pre-existing `doctor-kimi` cleanup failures reproduce on clean `main`; 1 tray test expects a checkout with no compiled binary — the local checkout has one). None touch these changes.
- Local reproduction confirmed the fix: the tray's control command now runs, and no console windows appear.

## Transparency

The reproduction, system checks, and logs were performed by the reporter (who is not a developer). The technical analysis and patch were prepared with the assistance of DeepSeek V4 Flash (current official release, codename **DeepSeek-V4-Flash-0731**).


duolahypercho:
The Windows fixes themselves check out — the spawn flags and `\\?\\` stripping are correctly gated, `tray-service-windows.mjs` is unreachable on macOS, and the port/Codex-termination safety rules are untouched. One problem: the new `source_root_path_is_executable_by_node` test had no `#[cfg(windows)]` gate, and since the stripping only exists under that cfg, the assertion fails on every non-Windows target — reproduced on macOS, and it would also fail the `ubuntu-latest` desktop CI job.

Pushed 6c1f155 gating the test and keeping a cross-platform passthrough assertion, plus a merge of current main. `cargo test` passes on macOS (7 tests), `npm run check` and `npm test` pass on the merged branch (1189/0). Merging.

PR #211, opened 2026-08-14
perf: speed up desktop snapshot

jepgambardella:
**Faster desktop snapshot**

The desktop panel no longer starts a second local-model process for the same data.

The control probe reuses local reads, and the optional registry size refresh no longer blocks the probe.

Validation:
- `npm test`
- `npm run check`
- `npm --prefix apps/desktop run check:js`

### DIFF
--- a/apps/desktop/ui/app.js
+++ b/apps/desktop/ui/app.js
@@ -204,7 +204,6 @@ function startPanel() {
       ["account", "account_usage"],
       ["providerUsage", "provider_usage"],
       ["providerSetup", "provider_setup"],
-      ["localModels", "local_models"],
       ["health", "router_health"],
       ["platform", "platform_info"],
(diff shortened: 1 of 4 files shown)


duolahypercho:
Review found one regression in the consolidation: `emitProbe` lost its call to `refreshVisionModelSizesIfStale()`, and the only remaining caller was the manual `control vision-bridge models` CLI action — nothing the desktop app runs automatically. That would have left the weekly vision-model size cache permanently stale in normal desktop use.

Pushed 94ede08 restoring the stale-checked refresh in the probe path (right before the reused local reads, so `annotateLocalModels` sees a fresh cache). The dedup itself checks out — 3 spawns instead of 4-plus per probe, no credential checks skipped. `npm run check` and `npm test` pass (1185/0). Merging.

PR #217, opened 2026-08-14
force a stable HTTP/1.1 upstream transport

IndelibleVivi:
## What changed

- install one process-wide Undici dispatcher before the router begins serving requests
- disable HTTP/2 while preserving the existing `fetch` call sites
- add a focused regression test and an explicit runtime dependency

## Root cause

A long-lived Node process could retain a destroyed HTTP/2 session. Once that happened, later upstream requests repeatedly failed with `ERR_HTTP2_INVALID_SESSION` and surfaced as local 502 responses until the service restarted. Forcing HTTP/1.1 avoids reusing the poisoned HTTP/2 session while keeping the existing request pipeline intact.

## Validation

- `node --test test/fetch-transport.test.mjs` — 1/1 passed
- the full local router suite also passed before publication
- live routed native and Kimi smoke requests succeeded after restart, with no new invalid-session errors after the latest ready boundary

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Volumes/OWC 2TB EXternal SSD/Code/codex-router/node_modules
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,8 @@
       "name": "codex-model-router",
       "version": "0.4.0-beta.3",
       "dependencies": {
-        "proper-lockfile": "4.1.2"
+        "proper-lockfile": "4.1.2",
+        "undici": "8.10.0"
       },
       "engines": {
         "node": ">=22.19.0"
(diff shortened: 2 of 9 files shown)


duolahypercho:
Review confirmed the root cause is real (api.openai.com negotiates h2, and `setGlobalDispatcher` from undici 8 does reach Node 22's built-in fetch). But the fix as submitted only covered 1 of the 4 long-lived processes: `start.mjs` spawns the router, api-forwarder, Kimi OAuth forwarder, and Grok OAuth forwarder as separate node processes, and a global dispatcher installed in one has no effect on the others. The api-forwarder — which carries most provider traffic — was still on the h2-capable default, so the incident class this PR targets stayed reproducible there. (The Kimi smoke test in the PR description ran through the unpatched oauth-forwarder process, so it wasn't actually exercising the fix.)

Pushed two commits:
1. Install `installStableFetchTransport()` in all four server entry points, plus a source-scan regression test: any `src/*.mjs` that calls `createServer(` must install the transport, so a future forwarder can't silently regress this.
2. Found while testing: the HTTP/1.1 dispatcher rejects requests carrying an explicit `Content-Length` beside a body (`UND_ERR_INVALID_ARG` → every api-forwarder upstream call 502'd). Node's default transport silently recomputed the header, so the explicit copies in `upstreamHeaders` (api-forwarder, oauth-forwarder) were already redundant — removed them and let fetch compute it, which is what reached the wire before.

Merged current main into the branch; `npm run check` and `npm test` pass (1187/0). Merging.

PR #220, opened 2026-08-14
feat: group a vendor's providers, and let the tray restart itself

duolahypercho:
Two tray changes.

## A vendor's providers are drawn together

The Providers list showed 13 rows for 5 vendors. `variantOf` already collapses the rows that share a credential — that is why Command Code and opencode Go each appear once despite carrying three protocol surfaces between them — but it cannot help the vendors whose providers are separate *accounts*:

| Vendor | Providers | Why they cannot merge |
|---|---|---|
| Z.ai | `zai-coding`, `zai-api` | different endpoints; a Coding Plan key is not billable on the metered one and vice versa |
| Kimi | `kimi-oauth`, `kimi-api`, `kimi-api-cn` | different regions and auth kinds |
| xAI | `grok-oauth`, `grok-api` | API key vs OAuth |

So they stay separately connectable and are only drawn together:

```
Z.ai                    2 accounts
    API
    GLM Coding Plan
```

The heading is derived from the display names rather than a new config field: `"Z.ai GLM Coding Plan"` and `"Z.ai API"` share their leading word, so the heading is `Z.ai` and the rows read `GLM Coding Plan` and `API`. Whole words only, so a group sharing a few letters cannot produce a heading that is half a word. A vendor with one provider renders exactly as before — a heading above a single row is noise, not structure.

`ownedBy` already existed on every provider and was already correct for all five vendors; it was only ever used for API metadata and error text. The only addition is passing it through `control`'s provider payload, and it is decoded as optional so an older router degrades to ungrouped rather than failing to decode.

## The grouped list is bound once per render

Reading the computed property inside its own `ForEach` regrouped and re-sorted the whole registry for every row. Measured at **89us per pass**, that is ~2.4ms of a render spent on nothing, and `body` re-runs on every store publish.

Bound once, the whole grouping costs **49us per render**, measured over 332 real renders with temporary instrumentation:

| | calls/render | cost/render |
|---|---|---|
| `providerVendorGroups` | 0.5 | 49us |
| `providerUsage(for:)` | 29.6 | 54us |
| `providerGroups(models:67)` | 0.5 | 8us |
| `providerName` | 4.8 | 1us |

~115us per render in total, so the remaining linear scans are deliberately left alone — hoisting them into the store would save a tenth of a millisecond and touch working code for it. The instrumentation was removed before this commit.

## Restarting the tray

Rebuilding the app replaces the bundle but leaves the running agent on the old binary, and quitting from the panel does not bring it back — `KeepAlive` is conditional on `SuccessfulExit: false`. A button beside **Refresh** closes that loop through the existing `control tray restart`.

It kills the process that asked for it, which is safe twice over: `launchctl kickstart -k` makes launchd commit to the relaunch *before* it signals the job, and an abnormal exit is exactly what `KeepAlive` covers. Only the failure path sets a message, because a success takes the window that would have shown it.

## Test plan

- [x] `swift test --package-path apps/macos/ModelRouterTray` — 18 passed, including the suite asserting every Chinese translation keeps the format specifiers its English source declares (covers the new `%d accounts`)
- [x] `npm test` — 1185 passed, 0 failed
- [x] Grouping verified against the live registry: Z.ai, Kimi and xAI group; the other 22 providers render unchanged
- [x] `control tray restart` verified end to end — tray PID changes and the agent comes back

Not verified: how the heading weight and 10pt indent *look*. Both are one-line changes if they want adjusting.

## Note for review

Chinese strings added for all four new user-facing strings, matching the repo's convention.


duolahypercho:
Review confirmed the tray restart path is safe (it kickstarts only the tray's own launchd label, never the router service or Codex) and the grouping works for Z.ai/Kimi/xAI — but it found one reachable defect: `commonWordPrefix` compares words case-sensitively, so `opencode Go/Zen` + `OpenCode Free` (both `ownedBy: opencode`) shared no prefix and hit the fallback, shipping a group headed by the raw registry id `opencode` claiming **2 accounts** — one of which is the anonymous keyless gateway, not an account at all. Every install would have rendered it.

Pushed a fix: the probe now ships each provider's `authMode`; anonymous providers always stand alone outside any accounts group; and a multi-account vendor whose display names share no leading words stays flat instead of falling back to an internal id as a brand. Older routers without the field decode as before (`authMode` optional).

Verified: `swift build` + `swift test` (18/18), `npm run check`, `npm test` 1189/0 on the branch merged with current main. Merging.

PR #221, opened 2026-08-15
fix: close finished subagents instead of leaving them Working

duolahypercho:
Work that was stranded on `feat/glm-5.3-and-zai-platform` — that branch's only PR (#201) merged for the GLM/Z.ai change, leaving these three commits pushed with nothing pointing at them.

## The problem

Codex 0.147 records a child's `FINAL_ANSWER` as `subAgentActivity: interacted` and maps that to **Working** until one of three things happens: the parent turn ends, the user clicks into the child, or the parent calls `interrupt_agent`. `close_agent` is not in the v2 toolset.

So a finished child sits there looking busy for the rest of a long parent turn, and a multi-agent run ends showing a row of children that all claim to still be working.

## The fix, in three parts

**1. Tell the parent to close them.** The managed `multi_agent_v2` block now ships a root usage hint instructing the parent to interrupt finished children, so new tasks settle the badge without a click.

**2. Close them anyway when it doesn't.** The hint alone is not enough — long San Francisco multi-agent parents often never call `interrupt_agent` despite it. The router scans request input for unfinished `FINAL_ANSWER` children and injects the missing `collaboration.interrupt_agent` calls into the parent response, on both the stream and non-stream paths, before `response.completed`.

**3. Do it for native parents too.** San Francisco multi-agent parents run through the router as **native GPT**, so a routed-only inject never fired for exactly the builds that needed it most. The close now queues on the native response path as well.

Model-authored interrupts are left alone; only missing closes are added.

## Test plan

- [x] `npm test` — 1192 passed, 0 failed
- [x] `test/subagent-completion.test.mjs` — 7 passed (new file, 236 lines)
- [x] `test/config-manager.test.mjs` — covers the usage-hint block

## Note on the history

The three original commits each rewrote the same `## Unreleased` entry, so replaying them individually re-conflicted at every step against a `main` that had moved. They are applied here as their net effect in one commit. The code applied cleanly via 3-way merge; only `CHANGELOG.md` needed resolving, and both the existing entries and the two new ones are preserved.


duolahypercho:
Adversarial review confirmed the core mechanism (injected `interrupt_agent` closes, sequence-preserving splice, Fernet-aware envelope reading) but demonstrated two reachable defects end-to-end:

1. **Detection was not scoped to real envelopes.** Any `message`-typed item's text was scanned for the FINAL_ANSWER shape, so an ordinary turn that merely *quoted* an envelope — docs, the changelog, a conversation about this very feature — got a fabricated `interrupt_agent` call spliced into its response, including on turns that never advertised the collaboration toolset (the empty-inventory bypass made the tool gate moot there). The `Task name:` fallback could even queue `/root` itself — an instruction to interrupt the parent. Fixed: evidence now comes from `agent_message` items only (the envelope type Codex uses on both native and routed paths), and a target resolving to the root is never queued.

2. **An injection switched the whole native stream onto the routed rewrite path.** Attaching the transform for `pendingInterrupts` also ran every native event through the routed-provider rewrites (re-serialization, spawn-model sanitizer). Fixed: native attachments run `injectOnly` — events the injection doesn't touch relay in their original bytes.

Pushed 6a6f79f with both fixes plus regression tests: quoted envelopes queue nothing, the root is never a target, an unrelated native event survives byte-identical beside an injection, and a stream with nothing to inject passes through untouched. Full suite on the branch merged with current main: 1201 pass / 0 fail. Merging.

PR #223, opened 2026-08-15
fix: address reviewer findings from the Aug 14 PR sweep

duolahypercho:
Three small items the adversarial reviews flagged while merging #206–#221, none blocking then, all real:

## Keyless baseUrl overrides must stay on loopback (from the #210 review)
The registry loader proves a keyless provider's checked-in baseUrl is loopback, but `MODEL_ROUTER_LOCAL_BASE_URL` / `MODEL_ROUTER_LMSTUDIO_BASE_URL` arrive at request time and skipped that proof — a non-loopback override sent an unauthenticated `Bearer local` request off-box. Overrides now resolve through `resolveProviderBaseUrl`, which refuses a non-loopback override on a keyless provider (falling back to the checked-in URL, logged once) while credentialed providers keep theirs. Covered in `test/registry.test.mjs`.

## Table hoisting uses the real TOML scanner (from the #216 review)
`foreignTableSegments` split parked content on any `[`-prefixed line, so a parked table holding a multiline string whose content looked like a header was cut mid-string. It now segments on `scanTomlDocument`'s headers — the same fail-closed scanner signed routing uses — and an ambiguous block aborts the rewrite instead of guessing. Regression test: a header-looking line inside a parked `"""` string survives the hoist intact.

## Stale comment (from the #212 review)
`test/startup-cleanup.test.mjs` still described startup as five sequential spawns; three run in parallel since #212.

## Test plan
- [x] `npm run check`
- [x] `npm test` — 1203 pass / 0 fail (two new tests)

https://claude.ai/code/session_0113FjuH7bU6xRMsUf8Aytio


duolahypercho:
Adversarial review of this PR itself found one more call site of the unguarded pattern: `providerPayload` in `src/model-discovery.mjs` read `process.env[provider.baseUrlEnv]` directly, and the keyless placeholder credential passes its credential check — so `discover-models local` / `curate-models` could still probe an off-box host with `Bearer local`. Now routed through the same `resolveProviderBaseUrl` guard.

Also noted by the review, intentional and worth a release-note line: the TOML-scanner hoisting validates the whole parked block, so a malformed foreign block that the old regex silently tolerated now fails the enable/disable closed (config left untouched) instead of being guessed at.

`npm run check` + `npm test`: 1203 pass / 0 fail. Merging.

PR #225, opened 2026-08-15
feat: research any enabled model for the v2 subagent role automatically

duolahypercho:
Closes the gap #183 kept hitting: people want arbitrary models — DeepSeek, Kimi, GLM, whatever their provider serves — as native v2 subagents, but the only promotion path was a per-model registry change gated on a manual four-part probe. Six models ever made it through.

**Selection is now the assignment.** The operator toggles a model on as a subagent (tray or `control subagents set`); the router's job is answering whether that model *under that provider* can hold the child role:

1. **Auto probe on enable** — a detached worker (same shape as the vision-download worker, so the tray toggle never blocks) spends 2 live requests proving streaming + a forced tool call through the installed router. Snapshot shows `checking` → verdict → catalog republish.
2. **Experimental advertisement** — a passing model is offered to Codex as a v2 subagent, recorded in protected `multi-agent-proofs.json`.
3. **Proof by observation** — Codex marks child turns with `x-openai-subagent`; the router's request path records the first clean completion as the durable proof (`proven`), and a structural rejection (400/422) demotes back to v1 with the reason kept in the subagent snapshot. Transient failures (429/5xx/disconnects) prove nothing and leave the window open.
4. `control subagents verify [slug ...]` re-researches explicitly; no-arg sweeps the enabled list. Select-all and mode changes never probe — a sweep across every provider is quota the operator must ask for.

The AGENTS.md invariant survives intact: local settings still never manufacture a v2 claim. The only writers of promotion evidence are the probe worker and the router's own observation of real traffic; unreadable proofs promote nothing; hidden/disabled slugs stay v1 whatever evidence they carry. Registry v2 remains the path for repo-shipped defaults.

## Test plan
- [x] `npm run check`
- [x] `npm test` — 1210 pass / 0 fail
- [x] New unit suite `test/subagent-proofs.test.mjs` (lifecycle, corrupt-state fail-closed, promotion overlay vs demotions/registry, candidate filtering, probe-crash handling)
- [x] New e2e in `test/routing.test.mjs`: marked child turn 200 → `proven`; marked child turn 400 → `failed` with reason; unmarked 200 → untouched
- [ ] Live acceptance on a real install: enable a non-v2 model, watch `checking → experimental`, spawn it once in Codex, confirm `proven`

https://claude.ai/code/session_0113FjuH7bU6xRMsUf8Aytio

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -342,7 +342,10 @@ so the unit of evidence is always the slug, never the model name.
 1. Enabling a non-registry-v2 model hands it to a detached capability probe
    (`src/subagent-verify.mjs`): two live requests through the installed router
    proving streaming and a forced tool call. The proofs snapshot shows
-   `checking` until the verdict lands and the catalog republishes.
+   `checking` until the verdict lands and the catalog republishes. A worker
+   that dies without a verdict marks its slugs `failed` on the way out, and a
+   `checking` older than the probe ceiling is treated as abandoned — so
+   switching a wedged model off and on always re-researches it.
 2. A passing model is advertised to Codex as an **experimental** v2 subagent,
    recorded in the protected `multi-agent-proofs.json`. The first real child
    turn settles it: Codex marks child turns with `x-openai-subagent`, and the
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -331,6 +331,38 @@ to ship tested support to every installer.
 If the provider itself is unknown to the registry, stop treating the request as
 installation. It is repository development and requires the process below.
 
+### Subagent capability is researched, not asserted
+
+Switching a model on as a subagent (tray toggle, `control subagents set`) is
+the operator's whole job; deciding whether that model **under that provider**
+can hold the v2 child role is the router's. The same model answers differently
+per provider — tool support, request profiles, and payload handling all vary —
+so the unit of evidence is always the slug, never the model name.
+
+1. Enabling a non-registry-v2 model hands it to a detached capability probe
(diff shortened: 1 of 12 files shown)


duolahypercho:
Adversarial review of the PR found one HIGH and two MEDIUMs; two are fixed in the follow-up commit, one is noted:

- **HIGH (fixed):** a probe worker dying mid-`checking` wedged the slug forever, and toggle-off/on didn't recover it (the candidate filter skipped `checking` without force). Now the worker marks unresolved slugs `failed` on exit, and `checking` older than the probe ceiling counts as abandoned — the ordinary toggle path re-researches it. Covered by a new staleness test.
- **MEDIUM (fixed):** `dsh-config-manager` built its subagent preset without the proofs overlay, so locally-verified models existed in the Codex picker but not the harness composition. The overlay now applies in `dshRoutedModels` with the same hidden/disabled demotions.
- **MEDIUM (follow-up, not blocking):** `multi-agent-proofs.json` is an unlocked read-modify-write with three writers; a concurrent write can drop another slug's update for one beat. It cannot produce a false promotion, self-heals on the next observation, and matches the existing pattern for `multi-agent-settings.json` — if flapping shows up in practice, `service-operation-lock` is the fix.
- One design nuance the review flagged as intentional, confirming: `proven` records that a marked child turn completed over the real provider — it is status-based, deliberately, because the dedicated tool/stream probe already ran at the `experimental` gate.

`npm run check` + `npm test`: 1211 pass / 0 fail. Merging.

PR #233, opened 2026-08-15
feat: credential-free install mode via --no-provider and --no-discovery

### LINKED ISSUE #224: Feature request: zero-provider / no-credential-discovery install mode
Codex Router currently expects at least one configured provider, and automatic setup may inspect existing macOS Keychain entries, CLI/OAuth sessions, and provider credential files. This makes it difficult to safely validate installation, service lifecycle, rollback, and uninstall behavior before trusting the router with real credentials. I would like to request an explicit credential-free installation mode, for example with flags such as `--no-provider` and `--no-discovery`.

When both flags are enabled, Codex Router should be installable without selecting or configuring any provider. It should not scan macOS Keychain, read existing CLI or OAuth sessions, inspect or import provider credential files, require provider credentials, create provider credential files, auto-discover providers, or automatically fall back to an existing provider. The existing native Codex configuration and authentication should remain unchanged. The router should be able to start in a deterministic idle state, bind only to loopback, preferably `127.0.0.1`, and should not route Codex traffic to any external provider or retry provider discovery in the background. Normal lifecycle operations such as `status`, `start`, `stop`, `doctor`, `uninstall`, and `rollback` should still work. If Codex traffic reaches the router while no provider is configured, it should return a clear local error instead of attempting discovery or external routing.

For network behavior, provider-related outbound connections should be `0`, credential-discovery-related outbound connections should be `0`, and the listener should be restricted to loopback only. If installation or update checks require GitHub, npm, PyPI, or another package registry, those connections should be explicitly documented and ideally suppressible for offline validation.

A possible validation flow would be:

```text
install --no-provider --no-discovery
service start
service status
doctor
service stop
uninstall
```

The complete lifecycle should work without configuring or reading any provider credentials.

## Acceptance criteria

- Keychain reads: `0`
- CLI/OAuth session reads: `0`
- Provider credential reads: `0`
- Provider credential writes: `0`
- Provider discovery attempts: `0`
- Provider external connections: `0`
- Listener: loopback only
- Original Codex config hash changed: `NO`
- Original Codex auth hash changed: `NO`
- Background provider retry/discovery: `NO`

After uninstall or rollback:

- Orphan router process: `0`
- LaunchAgent remaining: `NO`
- Proxy remaining: `NO`
- Router listening port: `NO`
- Routing override remaining: `NO`
- Native Codex launch: `PASS`
- Original config restored: `PASS`

This would allow users to verify Codex Router installation, lifecycle behavior, network behavior, rollback, and uninstall safety before granting it access to API keys, OAuth sessions, Keychain credentials, provider configuration, or production Codex traffic. The goal is not to change the normal user-friendly setup flow, but to provide an explicit, deterministic, credential-free lifecycle-validation mode for users who require stricter security controls. If an equivalent mode already exists through an undocumented flag, environment variable, or configuration option, documentation for that path would also solve this use case.

duolahypercho:
Closes #224.

Adds an explicit, deterministic, credential-free install mode so the router's install, lifecycle, network, and uninstall behavior can be validated before it is trusted with API keys, OAuth sessions, Keychain credentials, or production Codex traffic.

## What's in here (7 commits, in dependency order)

1. **`fix:` retire the background service when the last managed config block is gone** — `installedTargets()` keyed "codex is installed" on the cached native catalog, a file uninstall deliberately retains, so the service, LaunchAgent, and listening ports survived every codex uninstall. The witness is now the managed block in `config.toml`. `bin/disable` of the last client retires the service too (matching the Windows wrapper), and `bin/enable` reinstalls it on the way back.
2. **`feat:` a `stop` subcommand** — `bin/stop`, `bin/model-router … stop`, and the Windows dispatcher, completing the install → start → status → doctor → stop → uninstall loop.
3. **`feat:` a persisted discovery kill-switch** — `src/discovery-mode.mjs` (`discovery-mode.json`, overridable with `CODEX_ROUTER_NO_DISCOVERY=1|0`), consulted by every credential reader: the credential resolver (env/file/Keychain/CLI-session), `configuredProviderIds()`, the Kimi/Grok OAuth status readers, the CLI-session reader, Codex's `auth.json` reader, and the `codex login status` probe (skipped with a distinct reason the catalog treats as deliberate signed-out).
4. **`feat:` local idle error instead of native forwarding** — with discovery disabled, `/responses` turns without a route and the native image/web-search endpoints return `503 router_idle_no_provider` before any upstream fetch. Hiding every provider *without* the flag keeps native passthrough exactly as today (regression-tested).
5. **`feat:` doctor idle mode** — an explicit empty selection plus the marker reports at warn, the merged catalog is held to the nothing-routable standard, the per-provider credential sweep (itself discovery) collapses to one row, and the config-load probe / session-refresh nudge — which make Codex read its real session file — are skipped. Doctor exits 0 on a healthy idle install; an empty selection without the marker still fails loudly.
6. **`feat:` the install flags** — `--no-provider` (empty explicit selection, no credential steps; conflicts with `--guided`/`--providers`/key flags) and `--no-discovery` (requires `--no-provider`), plumbed through `install.sh`, `install.ps1`, and `src/setup.mjs`. Setup writes the discovery marker on **every** run, so re-running without the flags is the exit path. `ensure-configured` now accepts an explicitly empty selection as idle — which also un-breaks `bin/update` for anyone who hid their last provider by hand.
7. **`docs:`** INSTALL.md section (flags, guarantees, lifecycle, uninstall-vs-rollback), SECURITY.md network-boundary claims (including the one remaining `codex debug models --bundled` static-list spawn), AGENTS.md invariant, README pointer, CHANGELOG.

## Acceptance criteria mapping

- Keychain / CLI-OAuth-session / credential reads & writes: **0** — kill-switch at every reader; `test/provider-credentials.test.mjs` asserts `keychainProbeCount()` stays flat with a real credential file on disk.
- Provider discovery / external connections: **0** — `configuredProviderIds()` returns `[]`; `test/router-idle.test.mjs` points `CODEX_NATIVE_BASE_URL` at a recording server and asserts zero hits.
- Listener loopback-only: already the default everywhere; doctor now warns if `CODEX_ROUTER_HOST` is non-loopback while discovery is off.
- Codex config/auth unchanged: only the standard managed block is written; `auth.json` is never opened (reader guarded).
- Background retry/discovery: none exists (no timers in any server process); subagent verification has zero candidates with zero providers.
- Lifecycle: `status`/`start`/`stop`/`doctor`/`uninstall` all work idle; uninstall now actually removes the LaunchAgent/service (commit 1). `rollback` reverts the source checkout, not the install — documented.

## Test plan

- [x] `npm run check`, `sh -n install.sh`, `npm audit --omit=dev` (0 vulnerabilities)
- [x] `npm test`: 1244 pass / 0 fail (8 pre-existing skips)
- [x] New suites: `discovery-mode`, `router-idle` (mocked native base, zero hits), `doctor-idle` (spawn sentinel proves no `codex login` against the real home), `target-integration`; extensions to `setup`, `setup-exit-codes`, `provider-credentials`, `installer-scripts`, `model-router-cli`
- [ ] Live install → … → uninstall loop on a scratch machine (not run here: this machine has a production router install on the real service label and ports)

https://claude.ai/code/session_01FcQxQuxyLnPhr4ARF5mfb3

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -852,6 +852,28 @@ minutes later. Do not quietly drop the label because a check happened to pass.
   directories.
 - Do not restart or quit the Codex App from the installation task.
 
+## Discovery-disabled means no credential reader touches anything
+
+An install made with `--no-provider --no-discovery` persists a discovery
+kill-switch (`discovery-mode.json`, read through
+`src/discovery-mode.mjs` `discoveryDisabled()`, overridable with
+`CODEX_ROUTER_NO_DISCOVERY=1|0`). While it is set, the promise is absolute:
+no provider credential file, macOS Keychain item, other CLI's OAuth or
+session file, or Codex `auth.json` is read, no `codex login status` probe
+runs against the real `CODEX_HOME`, and traffic gets a local
+`503 router_idle_no_provider` instead of provider or native forwarding.
+
+1. Every new credential reader, sign-in probe, or session consumer must
+   consult `discoveryDisabled()` before its first read or spawn and report
+   "nothing found" rather than throwing. The guard belongs at the reader, not
+   only at its current callers — call graphs move.
+2. An explicitly written empty provider selection is a deliberate state, not
+   an error: `ensure-configured` reports it as idle, the doctor warns instead
+   of failing, and installing or updating on top of it must keep working.
+3. Never select a provider, re-enable discovery, or clear the marker on the
+   user's behalf. Re-running setup without the flags is the only exit path,
+   and it is the operator's to take.
+
 ## The `codex` shim is opt-in and must never break `codex`
 
 `src/codex-shim.mjs` can put a wrapper named `codex` on the user's PATH so the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+- **A credential-free install mode for lifecycle validation.** (#224)
+  `install.sh --no-provider --no-discovery` (PowerShell: `-NoProvider
+  -NoDiscovery`) installs the router idle: an explicit empty provider
+  selection, no credential prompts, and a persisted discovery kill-switch
+  honored by every credential reader — provider key files, the macOS
+  Keychain, other CLIs' OAuth and session files, Codex's `auth.json`, and the
+  `codex login status` probe all stay untouched. Codex traffic gets a local
+  `503 router_idle_no_provider` instead of provider or native forwarding, the
+  doctor reports the idle state at warn and exits 0, and a new `stop`
+  subcommand completes the install → start → status → doctor → stop →
+  uninstall loop. Re-running setup without the flags leaves idle mode. As
+  part of this, an explicitly empty provider selection now passes
+  `ensure-configured` as idle, which also un-breaks `bin/update` for anyone
+  who had hidden their last provider by hand.
 - **Uninstalling the last client integration now removes the background
   service.** Whether Codex still counted as installed was keyed on the cached
   native catalog, a file uninstall deliberately retains — so the service, its
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- **Uninstalling the last client integration now removes the background
+  service.** Whether Codex still counted as installed was keyed on the cached
+  native catalog, a file uninstall deliberately retains — so the service, its
+  LaunchAgent, and its listening ports survived every codex uninstall. The
+  installed-state witness is now the managed block in `config.toml`, which
+  enable writes and disable removes; `bin/disable` of the last client retires
+  the service too, matching what the Windows wrapper always did, and
+  `bin/enable` reinstalls it on the way back.
+
 - **Switching a model on as a subagent now researches it instead of ignoring
   it.** Only six registry-proven models could ever be spawned as native v2
   children; everything else the operator enabled was a silent no-show, and
--- a/README.md
+++ b/README.md
@@ -1270,6 +1270,8 @@ specific bytes. Browser and computer-use execution remains live-only.
 ./bin/model-router codex setup --guided
 ./bin/model-router codex doctor
 ./bin/model-router codex status
+./bin/model-router codex start
+./bin/model-router codex stop
 ./bin/model-router codex disable
 ./bin/model-router codex enable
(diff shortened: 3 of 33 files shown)


duolahypercho:
Adversarial review of this branch (run read-only in a scratch worktree while it was still being written — full suite was green, 1244/0, and the credential-reader chokepoints, idle-preserving `ensure-configured` logic, and native-relay gate all check out). Three findings before this merges:

**CRITICAL — the `--no-discovery` promise is broken on the very first install.** `src/catalog.mjs:172` (`captureNative()`) and `src/doctor.mjs:970` both spawn `codex debug models` — the **account-aware** variant, not `--bundled` — with no `discoveryDisabled()` guard, and both are reached during a real `--no-provider --no-discovery` install: `bin/install` refreshes the native catalog whenever `native-models.json` is absent (i.e. every fresh install), and `setup.mjs` runs doctor unconditionally as the last step. Reproduced empirically with a logging `codex` stub: the unguarded `debug models` fires twice. This contradicts the PR's own AGENTS.md section, and `SECURITY.md`'s new claim (*"the one Codex spawn that remains is `codex debug models --bundled`, which reads the binary's static model list, not credentials"*) is factually false for the shipped code. Also: `bin/install` passes `--bundled-native`, which `catalog.mjs` doesn't recognize — silent no-op. Fix: gate both call sites on `discoveryDisabled()`, teach the `test/doctor-idle.test.mjs` stub to distinguish `debug models` from `debug models --bundled` and assert only the bundled form runs, and only then keep the SECURITY/INSTALL claims.

**HIGH — service teardown fails open on an unreadable config.** `codexIntegrationInstalled()` in `src/target-integration.mjs` returns `false` when `readFileSync` throws on an *existing* `config.toml` (Codex mid-write, AV lock). `installedTargets()` feeds the retire-service-if-unused path, so a transient read failure can tear down the shared service for a user whose integration is plainly installed — the exact hazard the vision-bridge precedent fails safe against ("somebody was here and we cannot tell what they chose"). Fix: file present + unreadable → treat as installed; add the unreadable-file regression test.

**MEDIUM — document the deliberate divergence.** `discoveryDisabled()` treats a corrupt `discovery-mode.json` as *enabled* (resume credential reads), the opposite direction of the vision-bridge fail-closed precedent. It's tested and looks intentional, but worth one sentence in the AGENTS.md section saying so, or the next reader files it as a bug.

Happy to push the fixes here if you'd rather hand it off.

PR #235, opened 2026-08-15
fix(catalog): preserve native GPT metadata

jepgambardella:
## What changed

Preserve non-empty metadata from Codex's bundled native catalog when the account catalog has an empty or missing field.

## Why

The account cache can use an older schema and return empty values. The merge then removed native capabilities such as the Fast speed tier from GPT models.

## Result

Native GPT models keep their bundled capabilities. Non-empty account values still control account-specific fields such as visibility.

## Validation

- `npm test` (1186 passed, 9 skipped)
- `npm run check`
- Local catalog regeneration verified Fast for GPT-5.6 Sol, Terra, Luna, and GPT-5.5.


duolahypercho:
Review confirmed the fix addresses the real symptom (a stale account schema wiping the Fast speed tier — the new test pins it), merges cleanly onto current main, and the suite is green. One change pushed before merge: the fallback iterated **every** bundled field, restoring anything the account left empty — which would also resurrect values an account can legitimately narrow, e.g. an effort ladder the account lost (`supported_reasoning_levels: []` feeds the DSH effort choices directly). The backfill is now an explicit allowlist — the observed tier fields, `input_modalities` (empty would describe a model nothing can call), and the binary-schema tool/instruction fields — with `visibility` and `supported_reasoning_levels` deliberately account-authoritative, and unknown future fields defaulting the same way. A new test pins the allowlist boundary and the both-empty case. `npm run check` + `npm test` on the branch merged with current main: 1247 pass / 0 fail. Merging — thanks for the catch and the clean repro test.

PR #243, opened 2026-08-15
feat: add official Xiaomi MiMo API provider

sirenexcelsior:
# feat: add official Xiaomi MiMo API provider

## Problem

codex-router already routes MiMo through ClinePass, opencode Go, and Command Code, but there is no direct route for Xiaomi's official MiMo API (`https://api.xiaomimimo.com/v1`).

When a direct MiMo route was tried over the Responses gateway, Xiaomi rejected the standard Codex tool declarations:

```text
HTTP 400 responses_feature_not_supported
tool type 'web_search' is not supported by this gateway phase
tool type 'custom' is not supported by this gateway phase
```

Codex sends hosted-style tool declarations (`tools: [{type: "web_search"}]`, plus custom tools from the native template). Forwarding those untouched to `api.xiaomimimo.com/v1/responses` is rejected by Xiaomi's gateway. MiMo's official OpenAI-compatible API accepts chat completions with `web_search_options` instead.

## Changes

- Add `xiaomi-mimo` provider with `protocol: "openai"`, so LiteLLM generates `openai/<gatewayModel>` with `use_chat_completions_api: true` and requests never touch the Responses gateway.
- Add `xiaomi-mimo/mimo-v2.5` (text + image, Codex standalone web search, 1M context) and `xiaomi-mimo/mimo-v2.5-pro` (text only, 1M context).
- Allow routed model metadata to narrow tool capabilities inherited from the native template: `supports_parallel_tool_calls: false`, `experimental_supported_tools: []`, and `supports_apply_patch_tool: false` keep Codex from offering custom tools MiMo rejects.
- Validate the new `supportsParallelToolCalls` and `experimentalSupportedTools` registry fields.
- Add registry, catalog, and API-forwarder regression tests, including a chat-completions request that keeps `web_search_options` and strips `client_metadata`.
- Document the provider and credential setup in the README.

## Verification

- `npm run check` passes.
- `npm test`: 1294 passed, 0 failed, 13 skipped.
- Local end-to-end check: Xiaomi `/v1/chat/completions` accepts both a plain request and a request with `web_search_options` (HTTP 200); a router-gateway MiMo request carrying a `web_search` tool also returns 200 after the protocol change.

## Notes

- Credential environment variable: `MIMO_API_KEY`; protected key file: `xiaomi-mimo-api-key.secret`; keychain service: `codex-router-xiaomi-mimo-api`.
- Base URL override: `XIAOMI_MIMO_API_BASE_URL`.
- No API keys or credentials are included in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,8 @@ Linux installations support the Codex CLI.
 | DeepSeek V4 Pro (Ollama Cloud) | `ollama-cloud/deepseek-v4-pro` | Ollama Cloud API key |
 | DeepSeek V4 Flash (Ollama Cloud) | `ollama-cloud/deepseek-v4-flash` | Ollama Cloud API key |
 | MiniMax M3 | `minimax-token-plan/minimax-m3` | MiniMax Token Plan API key |
+| MiMo-V2.5 (Xiaomi API) | `xiaomi-mimo/mimo-v2.5` | Xiaomi MiMo API key |
+| MiMo-V2.5-Pro (Xiaomi API) | `xiaomi-mimo/mimo-v2.5-pro` | Xiaomi MiMo API key |
 | Qwen3.8 Max (Plan) | `qwen-plan/qwen3.8-max` | Alibaba Model Studio plan API key |
 | Qwen3.8 Max Preview (Plan) | `qwen-plan/qwen3.8-max-preview` | Alibaba Model Studio plan API key |
 | Qwen3.7 Max (Plan) | `qwen-plan/qwen3.7-max` | Alibaba Model Studio plan API key |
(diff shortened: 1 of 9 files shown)


duolahypercho:
Reviewed and validated locally (CI hadn't run — first-time contributor approval was pending, now approved).

**What I checked and found clean:** the two new registry fields are validated correctly and only emitted when explicitly set, and the added catalog test proves an existing model with neither field set is unchanged, so there's no regression for anything already shipped. Gateway ids don't collide with the existing `clinepass-mimo-v2-5(-pro)` routes, the credential file/keychain naming matches convention exactly, the endpoint is TLS, and the routing test genuinely exercises the path against a local mock upstream rather than stubbing it. Full suite on a trial merge with main: 1298 passing, 0 failing.

**One change I pushed to this branch** (`e55ecff`): both models compacted at `996147` — 95% of their window, the highest ratio in the registry. The routed catalog already tells Codex to treat the window as 95% effective, so at that setting a turn can grow past the usable limit before anything compacts it. Every other million-token route here compacts at `900000`, including the three other MiMo routes (`clinepass/mimo-v2.5`, `commandcode/mimo-v2.5-pro`, `opencode-go/mimo-v2.5`). Updated the corresponding assertion in `test/registry.test.mjs`.

**Two follow-ups, not blocking this merge:**

1. No tray icon. AGENTS.md's "Ship a new provider" checklist asks for an asset under `apps/macos/ModelRouterTray/Sources/Resources/ProviderIcons/` and a sourced row in `PROVIDER-ICON-SOURCES.md`. I didn't add one rather than commit a brand asset without recording its provenance. Everything else on that checklist is satisfied automatically here, since the tray setup card, `install.sh --providers`, and `curate-models` are all registry-driven.
2. Not live-verified. `./bin/test-model 'xiaomi-mimo/mimo-v2.5' --live --yes` hasn't been run because there's no MiMo credential on this machine, so `listed: true` ships these to every installer's picker on documentation alone. Worth one live request before the next release if you have a key.

Also worth confirming at some point: this declares `contextWindow: 1048576` while the already-shipped `clinepass/mimo-v2.5` says `1050000` for the same model. The ~1.4K difference is immaterial to routing, but the two routes disagreeing is the kind of thing that's easier to reconcile now than later.

PR #245, opened 2026-08-15
Add smart default-on tool result aging

HaileyStorm:
## Summary

- default routed tool-result aging on while preserving explicit Off, the environment kill switch, and native aging as a separate default-off choice
- keep exact or ambiguous outputs intact; summarize only safely parsed, highly repetitive observational logs after the model has acted and outside the newest-four frontier
- retain exact bytes owner-privately before rewriting, with bounded storage, route/call/digest binding, hardened POSIX and Windows file checks, and fail-original behavior
- expose truthful sidecar state, savings/degradation telemetry, default-on copy, and read-only browser behavior
- remove the old rerun-the-tool recovery instruction; receipts contain a deterministic digest, summary, and bounded head/tail without a model-visible path or capability

## Verification

Linux:
- focused: 131 tests, 128 passed, 3 expected platform skips
- full npm test: 1,328 tests, 1,314 passed, 14 expected platform/integration skips
- npm run check and git diff --check
- independent security and regression review: clean

Native Windows, exact final source hashes:
- deterministic two-process torn-key repair: passed
- unprotected-key ACL refusal: passed
- tool-result aging: 20 passed, 1 POSIX-only skip
- exact owner-only ACL and extra-grant rejection: 2/2 passed
- exact owner-local retrieval: passed

## Boundaries

- no provider calls were made
- real Chromium UI and macOS tray acceptance remain host-specific follow-ups
- clickable owner Save/Open UI is intentionally separate; this PR provides exact owner-local retention/retrieval without exposing a retrieval capability to the model

Source base: 83d666a5ac2c421b4b1a18c813380bafd6fe1f8a
Initial 24-file manifest SHA-256: fe97cddbdf16d0ca2be7878e0de6a1fb64680ddbb0ece95f1fb878eb13bca09f
Final Windows follow-up manifest SHA-256: 3df291d5262a9e93c2d76fed23080fabc2f5536effb588cf659e0c1bfaa67b3b


duolahypercho:
Reviewed this from three angles (security of the retention store, correctness/regression against the aging work that landed in #239, and model-facing behavior). CI hadn't run — first-time-contributor approval was pending — so I also ran everything locally against a merge with current `main`.

## Fixed on this branch

**`npm test` was failing 9 tests on a stock macOS checkout**, which CI could not see. `ensureDirectoryPathWithoutLinks` validated the store by walking from the filesystem root and refusing any component that was a symlink. `/var` is a symlink to `/private/var` on every stock macOS install, so every path under `os.tmpdir()` traverses one — as do container bind mounts and symlinked home directories. GitHub's macOS runners hand out a temp path that happens not to cross a link, which is why CI stayed green while the same commit failed locally.

This mattered beyond the tests: in production the same check runs against the real `STATE_DIR`, so any install pointing state at `/tmp`, a bind mount, or a symlinked home would have had retention fail 100% of the time. Because retention correctly fails to the original result, it would have failed *quietly* — the feature simply never working, with no data loss to signal it.

Pushed `fix(retention)`: both that walk and the realpath-identity check in `assertPlainOwnedDirectory` now canonicalize first and compare against that. The guarantee that matters is unchanged — the directory written to is the one that was validated, and we own it — but a link the operator didn't plant is no longer treated as an attack. Your `retention rejects a symlinked state path before touching its target` test passes unchanged, and I added its counterpart covering the OS's own case. Suite is now **1319 passing, 0 failing, 13 skips** on the merged tree.

## Verified, no change needed

The `file-security.mjs` rework is a genuine fix, not just refactoring: the old `icacls /grant:r` only replaced the ACE for the named identity and left any pre-existing `Everyone`/`Users` grant in place, so this was a latent Windows over-permissioning bug affecting **every** private file the router writes — credentials and OAuth tokens included, not just retention. The replacement breaks inheritance, strips explicit ACEs, and `privateFileIsProtected` now fails closed unless exactly one owner rule matches.

Also confirmed by constructing the state files by hand rather than reading the code alone: an explicit **Off survives the upgrade** (only `defaultSettings()` changed), the env kill switch still wins, native aging stays default-off, #239's counters are intact and still written with `!== undefined` so a `0` survives, and a retention failure cannot propagate into the request path (guarded at both context creation and the `retain()` call).

## Two things worth your judgement before this ships

**1. The receipt no longer tells the model how to recover.** Today's receipt says *"Repeat the preceding &lt;tool&gt; call with the same arguments if exact or omitted content is needed."* That instruction is what commit `38b49fb` measured when it graduated aging out of experimental — the finding was specifically that models "re-run the tool instead of guessing when a compacted fact is requested." This PR removes that line and replaces it with four lines describing an owner-private retrieval the model explicitly cannot invoke ("No path, command, or retrieval capability is sent to the model"). So the validated recovery path is gone, and in its place is text inviting a model to believe exact bytes are obtainable when they aren't. The probe that measured the old behavior is rewritten in the same PR, and the description notes no provider calls were made, so the new design hasn't been re-validated the way the old one was.

The receipt header also roughly doubles, 331 → 653 bytes (~81 tokens per compacted result, ~4K tokens across 50 in a long session), on a feature whose purpose is saving context. About 110 of those bytes are a second copy of the same sha256 already printed on line one.

Suggestion: keep the re-run instruction and drop the three lines about the unreachable capability. That preserves the validated behavior and most of the token cost disappears.

**2. Retention is permanent by design.** There is no TTL, no purge command, and no `doctor`/uninstall integration — a silently growing (to 512 MiB) local archive of raw tool output, which in this product routinely includes file contents and command output. It is `0600`, owner-only, correctly excluded from support bundles, and the README is upfront about it. Still, this is the first time the router persists model-visible *content* to disk, where the established stance has been that telemetry records "counts and bytes... never the results." Turning that on by default is a product call I don't think a reviewer should make silently, so I'm flagging it rather than deciding it.

Everything technical is now green; these two are judgement, not defects.

duolahypercho:
Pushed the second of the two judgement items as `fix(aging)`, so this is now green on both.

The receipt tells the model to re-run the tool again — naming the tool from the paired call — and the three lines about owner-private retention are gone. They were all true and none were actionable: the model has no way to reach those bytes, so the text was inviting it to claim a recovery it cannot perform. Re-running the tool is the recovery that works, and it is what `38b49fb` measured when it graduated compaction out of experimental. The digest, deterministic summary, and omitted-middle accounting stay — those are what the smart summary genuinely adds over the old receipt.

It is also cheaper: the header drops from 610 to 421 bytes, roughly 47 tokens back per compacted result.

Suite on the merged tree is **1320 passing, 0 failing, 13 skips**.

The retention store itself is untouched by this change — bytes are still retained, still `0600`, still owner-only. The only thing that changed is what the model is told about them, which is now nothing.

That leaves one open item, which is yours rather than a reviewer's: retention has no TTL, no purge command, and no `doctor`/uninstall integration, so it is a permanent local archive of raw tool output that grows to 512 MiB and is on by default. Worth a follow-up issue for a `control tool-result-aging purge` and a doctor line that surfaces the directory's size, so it is at least visible and clearable.

duolahypercho:
Two notes, then I'm going to stop pushing to this branch — we're both editing the same files within minutes of each other and I'd rather not trample your work.

## Windows is the only red job left

`test (windows-latest)` fails in `assertPlainOwnedDirectory`:

```
Error: Retained-result storage may not traverse a link.
    at assertPlainOwnedDirectory (src/tool-result-retention.mjs:175)
    at ensurePrivateDirectory (src/tool-result-retention.mjs:230)
```

macOS and Ubuntu are green, so this is Windows-specific to the identity comparison:

```js
normalizedRealPath(realpathSync.native(directory)) !==
normalizedRealPath(canonicalExistingPath(directory))
```

`realpathSync.native` normalizes 8.3 short names, and `canonicalExistingPath` walks components lexically, so it preserves whatever it was given. GitHub's Windows runners hand out a `TEMP` under `C:\Users\RUNNER~1\...`; realpath returns `C:\Users\runneradmin\...` and the lexical walk returns `RUNNER~1`, so the two sides differ and the guard fires even though nothing is wrong. I couldn't confirm the exact path from the log (it isn't printed), so treat that as the likely cause rather than a certainty — but the shape of the failure fits, and it's Windows-only.

If it is that, normalizing both sides fixes it without weakening anything, since `canonicalExistingPath` already rejects untrusted links before this point:

```js
normalizedRealPath(realpathSync.native(directory)) !==
normalizedRealPath(realpathSync.native(canonicalExistingPath(directory)))
```

## The receipt wording is worth a maintainer decision, not a tug-of-war

I see `Keep aged-result recovery non-executing` reverted my change and went further — the receipt now tells the model *"Do not repeat the tool call to recover omitted content. If exact content is needed, ask the owner."*

I understand the goal, and I'm not going to re-push against it. But it's worth flagging plainly: commit `38b49fb` graduated this feature out of experimental on the strength of an A/B whose finding was that models **re-run the tool** instead of guessing when a compacted fact is requested. That was the evidence that compaction was safe to enable. Instructing the model not to re-run removes that path, and "ask the owner" isn't available to an agent running unattended — which is the case aging matters most in. The likely outcome is the model guessing from head/tail, which is the failure the original wording existed to prevent.

There's a real tension here: you're protecting against a model treating a digest as a retrieval capability, which is a legitimate worry. But the two aren't in conflict — "repeat the call" and "the digest is not a retrieval handle" can both be true in the same receipt.

Since this changes behavior the owner validated, I've left it to @duolahypercho to call rather than deciding it in review. Everything else here I'm satisfied with: the `file-security.mjs` ACL fix is a genuine improvement for every credential file on Windows, the default flip correctly preserves an explicit Off, and #239's counters survive the merge.

HaileyStorm:
Closed the Windows 8.3-path finding in d57dffa13d347952c331949d9c5e9205eecf3f77. The directory check now runs canonicalExistingPath first (so untrusted links/reparse points still fail closed), then applies realpathSync.native to both lexical spellings before identity comparison. Added regressions that reproduce the short/long alias as (short, short), require both operands to be realpathed, and prove link validation throws before any realpath call on POSIX, drive-rooted Windows, and UNC-shaped roots. Linux focused: 122 pass / 1 platform skip; Linux full: 1,324 pass / 14 skip / 0 fail; npm run check and diff-check pass. Waiting for upstream Windows CI as the acceptance gate.

The new CI run 31915532888 is currently `action_required` with no jobs because this external-branch revision needs maintainer workflow approval; no test result is being claimed from it yet.

PR #253, opened 2026-08-16
Relay deferred codex_apps connector tools to routed models

### LINKED ISSUE #252: Routed models reach only the deferred subset of codex_apps connector tools
Routed models reach only the deferred subset of `mcp__codex_apps__*` connector tools, so app connectors are effectively unusable off the native route.

## What happens

On a routed model (grok-oauth/grok-4.6, deepseek/deepseek-v4-*), a task that needs an app connector never finds the tools. In one session the model looped for ~30 turns trying to read Outlook mail, reporting each time that the connector was present but its tools were not mounted:

> Outlook 插件在，但工具还没挂进当前会话。 (“The Outlook plugin is there, but its tools are not mounted in this session.”)

It then went digging on its own — read the plugin's `SKILL.md` out of `~/.codex/plugins/cache/`, called `list_mcp_resources` (which *did* return the `codex_apps` resources, so the MCP server was connected and reachable in that very session), and finally gave up on the connector and shelled out to a local `himalaya` IMAP binary instead.

The same account on the native route calls those tools without trouble. A pre-router session on this machine shows 135 successful `microsoft_outlook_email.list_messages` calls plus `gmail.search_emails`, `fetch_messages_batch`, and friends, so the connectors themselves are fine.

## Why — the deferred subset is all that survives

The captured fixture in `test/namespace-relay-routing.test.mjs:141` is itself the clearest statement of the problem. What the client sends on a routed request is:

```js
tools: [
  { type: "tool_search" },
  ...
  { type: "namespace", name: "mcp__codex_apps__github", tools: [ { name: "fetch_issue", ... } ] },
]
```

One tool. The real GitHub connector on this machine exposes **89**. Outlook exposes **46**, Notion **18** — 171 across all connected apps. Codex ships a small subset inline and registers the rest with deferred loading, leaving `tool_search` as the only way to reach them.

`flattenNamespaceTools` then drops the search control (`src/namespace-relay.mjs:107`):

```js
// Codex adds this control tool when one or more namespace tools use
// deferred loading. By this boundary every namespace is expanded below,
// so the search control is redundant; ...
if (tool?.type === "tool_search") { changed = true; continue; }
```

Dropping it is right — chat-completions providers reject `type: "tool_search"`. But "every namespace is expanded below" holds only for namespaces whose tools the client actually shipped. For `mcp__codex_apps__*` it does not: the namespace arrives deliberately incomplete, the search control that was supposed to fill it in is gone, and nothing restores the remainder. The routed model ends up with 1 of 89 tools and no way to ask for the other 88.

`codex_app` (singular, the app runtime) already has exactly this fix. `mergeCodexAppTools` at `src/codex-app-tools.mjs:1190` merges a captured snapshot back in, and its comment describes the identical situation:

> The client sends a reduced codex_app namespace (three app tools) on routed requests; the rest are deferred client-side. Merge the full app toolset in so routed providers see the same tools a native model sees.

`mcp__codex_apps__*` (plural, the connectors) has no equivalent.

## The definitions are already on disk

Unlike `codex_app`, this would not need a hardcoded snapshot. Codex maintains the full set locally at `$CODEX_HOME/cache/codex_apps_tools/<hash>.json` (`schema_version: 4`), refreshed as connectors change — the copy here was rewritten during the most recent session. Each entry carries everything the relay needs:

```json
{
  "server_name": "codex_apps",
  "tool_namespace": "codex_apps__microsoft_outlook_email",
  "namespace_description": "Search and reference your Microsoft Outlook email.",
  "connector_id": "connector_...",
  "connector_name": "Microsoft Outlook Email",
  "tool": {
    "name": "microsoft_outlook_email.list_messages",
    "description": "...",
    "inputSchema": { ... }
  }
}
```

`tool.name` is already the exact name Codex dispatches (`{ namespace: "mcp__codex_apps__microsoft_outlook_email", name: "list_messages" }` after the existing restore path), so filling the namespace in needs no new response-side handling at all — `buildNamespaceLookups` and `rewriteNamespaceFunctionCall` map it back as they do today.

## What makes it non-trivial

Merging everything unconditionally is not an option. Serialized schema weight from this machine's cache:

| namespace | tools | schema size | approx. tokens |
|---|---:|---:|---:|
| `microsoft_outlook_email` | 46 | 259 KB | ~66k |
| `github` | 89 | 201 KB | ~51k |
| `notion` | 18 | 81 KB | ~21k |
| others (6) | 18 | 108 KB | ~28k |
| **total** | **171** | **649 KB** | **~166k** |

That is the whole point of the deferred loading this bypasses.

Options, roughly in order of how much they cost to build:

1. **Opt-in per connector.** A config key (`codex_apps_connectors = ["microsoft_outlook_email"]`) merges only those namespaces in full. Smallest change, reuses `mergeCodexAppTools` almost verbatim with the cache as source instead of a constant. Operator pays a known token cost for the connector they actually want.
2. **Opt-in per tool**, same mechanism at finer granularity, for people who want three Outlook tools rather than 46.
3. **A router-side stand-in for `tool_search`.** Closest to native behavior and the only option that scales to all 171, but it needs a discovery tool the router can answer without executing anything, which is a real design question rather than a merge.

A name-plus-one-line index of all 171 tools serializes to ~16 KB (~4k tokens), if a cheap always-on discovery hint is useful as part of (3).

## Environment

- codex-router 0.4.0-beta.2 (`agent/preserve-native-account-catalog`), macOS 27.0, target Codex, codex-cli 0.146.0
- Providers: deepseek, grok-oauth, local
- Observed on grok-oauth/grok-4.6; the relay path is provider-independent, so deepseek routes the same way

Happy to put up a PR for option 1 if that is the direction you would take.

wousp112:
Fixes #252.

## The gap

Codex presents each connected app connector as its own namespace — `mcp__codex_apps__github`, `mcp__codex_apps__microsoft_outlook_email` — and loads them lazily. On a routed request the client ships a handful of the tools inline and leaves `type: "tool_search"` as the way to reach the rest. `flattenNamespaceTools` drops that control, correctly: chat-completions providers reject the type outright. But the comment above the drop —

```js
// By this boundary every namespace is expanded below,
// so the search control is redundant;
```

— only holds for namespaces whose tools the client actually shipped. Connector namespaces arrive deliberately incomplete. The fixture in `test/namespace-relay-routing.test.mjs` captures it exactly: one `fetch_issue` under `mcp__codex_apps__github`, where the connector really exposes 89 tools. The remainder is unreachable on every routed model.

What that looks like in practice: a Grok 4.6 session spent about thirty turns trying to read Outlook mail, reporting each time that the connector was present but its tools were not mounted, then gave up and shelled out to a local IMAP binary. The same account calls those tools without trouble on the native route.

## The fix

`mergeCodexAppTools` already solves this for `codex_app`, the app's own runtime, by merging a captured snapshot back in. Connectors can't use a snapshot — which ones exist is per-account and changes whenever someone connects one — and don't need to: Codex maintains the full set at `$CODEX_HOME/cache/codex_apps_tools/<hash>.json` and refreshes it as connectors change. This reads that cache and fills the deferred remainder back into the request.

Nothing changes on the response path. The merged tools flatten to `mcp__codex_apps__<connector>__<method>` and restore through the existing relay to `{ namespace, name }`, which is the shape Codex dispatches:

```
restored -> {"type":"function_call","name":"list_messages",
             "namespace":"mcp__codex_apps__microsoft_outlook_email", ...}
```

## Opt-in, because schemas are big

Merging every connector would send ~166 KB of schema (the whole cache is 649 KB including output schemas and metadata this strips). That size is why Codex defers them at all, so the operator names the ones worth the context:

```
$ codex-router connector-tools list
  relayed  connector                tools   ~tokens
       no  github                      89     20892  GitHub
       no  microsoft_outlook_email     46     18916  Microsoft Outlook Email
       no  notion                      18     17664  Notion
       ...
$ codex-router connector-tools enable microsoft_outlook_email
```

With no selection recorded, `mergeConnectorTools` returns the original array by identity and the request is untouched — an install that never opts in behaves exactly as it does today. A missing or corrupt cache falls back the same way.

The client's own definition always wins; the cache only supplies what deferred loading held back. A selected connector the client omitted entirely is appended whole, on the same reasoning `mergeCodexAppTools` relies on — the client dispatches those calls natively whether or not the definition was in the request.

## Notes for review

- Applied in the non-responses branch next to `mergeCodexAppTools`, since that is where the observed failure is (Grok and DeepSeek both route as chat completions). Responses-native providers keep `tool_search` today, so they are left alone.
- The cache is parsed once and re-read only when its mtime changes, so the 1.4 MB file is not touched per request.
- `MODEL_ROUTER_CODEX_APPS_CACHE_DIR` and `MODEL_ROUTER_CONNECTOR_TOOLS_STATE` override the paths for tests.

## Tests

`test/codex-apps-connectors.test.mjs` covers cache parsing and the namespace/method naming, a missing and a corrupt cache, the untouched-by-default path, the merge, client-definition precedence, an entirely omitted connector, and the round trip through `flattenNamespaceTools` + `rewriteNamespaceFunctionCall`.

### DIFF
--- a/bin/connector-tools
+++ b/bin/connector-tools
@@ -1,0 +1,5 @@
+#!/bin/sh
+set -eu
+
+repo_dir=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)
+exec node "$repo_dir/src/connector-tools.mjs" "$@"
--- a/src/codex-apps-connectors.mjs
+++ b/src/codex-apps-connectors.mjs
@@ -1,0 +1,215 @@
+// Codex connects app connectors (Outlook, GitHub, Notion, ...) through its
+// `codex_apps` MCP server and presents each one to the model as its own
+// namespace: `mcp__codex_apps__github`, holding `fetch_issue`, `list_issues`,
+// and the rest.
+//
+// Those namespaces use deferred loading. On a routed request the client ships
+// only a handful of the tools inline and leaves `type: "tool_search"` as the
+// way to reach the others. `flattenNamespaceTools` drops the search control --
+// it has to, chat-completions providers reject the type outright -- so on a
+// routed model the deferred remainder is simply unreachable. A live capture
+// here sent one GitHub tool out of the 89 the connector actually exposes.
+//
+// `mergeCodexAppTools` already solves the same problem for `codex_app`, the
+// app's own runtime, by merging a captured snapshot back in. Connectors cannot
+// use a snapshot: which ones exist is per-account and changes whenever the
+// operator connects one. They do not need it either, because Codex maintains
+// the full set on disk at `$CODEX_HOME/cache/codex_apps_tools/<hash>.json` and
+// refreshes it as connectors change. This module reads that cache and fills
+// the deferred remainder back into the request.
+//
+// Merging is opt-in per connector because the schemas are big -- 649 KB, about
+// 166k tokens, for all 171 tools on the capture machine. That size is why
+// Codex defers them in the first place, so the operator names the connectors
+// worth the context and the rest stay deferred exactly as they are today.
+
+import { existsSync, readdirSync, readFileSync, statSync } from "node:fs";
+import path from "node:path";
+
+import { CODEX_HOME } from "./paths.mjs";
+
+export const CONNECTOR_TOOL_CACHE_DIR =
+  process.env.MODEL_ROUTER_CODEX_APPS_CACHE_DIR ||
+  path.join(CODEX_HOME, "cache", "codex_apps_tools");
+
+// The client prefixes every MCP server namespace with `mcp__`, so the cache's
+// `codex_apps__github` is `mcp__codex_apps__github` in a request.
+const CLIENT_NAMESPACE_PREFIX = "mcp__";
+export const CONNECTOR_SERVER = "codex_apps";
+
+// `<connector>.<method>` is the MCP tool name; the namespace child is the
+// method alone. Splitting on the first dot keeps a method containing one from
+// losing its tail.
+function methodName(toolName, connector) {
+  if (typeof toolName !== "string") return undefined;
+  const prefix = `${connector}.`;
+  if (toolName.startsWith(prefix)) return toolName.slice(prefix.length) || undefined;
+  const dot = toolName.indexOf(".");
+  return dot === -1 ? toolName : toolName.slice(dot + 1) || undefined;
+}
+
+// Codex writes one cache file per connector set and leaves older ones behind
+// when the set changes, so the newest file is the live one.
+function newestCacheFile(dir) {
+  let entries;
+  try {
+    entries = readdirSync(dir).filter((name) => name.endsWith(".json"));
+  } catch {
+    return undefined;
+  }
+  let newest;
+  for (const name of entries) {
+    const full = path.join(dir, name);
+    let mtime;
+    try {
+      mtime = statSync(full).mtimeMs;
+    } catch {
+      continue;
+    }
+    if (!newest || mtime > newest.mtime) newest = { path: full, mtime };
+  }
+  return newest?.path;
+}
+
+// Reads the cache into `namespace -> { name, description, tools: Map }`, keyed
+// by the namespace name a request carries. Returns an empty Map when the cache
+// is absent or unreadable: no connectors merged is exactly today's behavior,
+// which is the right thing to fall back to.
+export function readConnectorToolCatalog(dir = CONNECTOR_TOOL_CACHE_DIR) {
+  const catalog = new Map();
+  if (!dir || !existsSync(dir)) return catalog;
+  const file = newestCacheFile(dir);
+  if (!file) return catalog;
+  let parsed;
+  try {
+    parsed = JSON.parse(readFileSync(file, "utf8"));
+  } catch {
+    return catalog;
+  }
+  const entries = Array.isArray(parsed?.tools) ? parsed.tools : [];
+  for (const entry of entries) {
+    if (entry?.server_name && entry.server_name !== CONNECTOR_SERVER) continue;
+    const toolNamespace = entry?.tool_namespace;
+    const tool = entry?.tool;
+    if (typeof toolNamespace !== "string" || !tool?.name) continue;
+    const connector = toolNamespace.slice(toolNamespace.lastIndexOf("__") + 2);
+    const name = methodName(tool.name, connector);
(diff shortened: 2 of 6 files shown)


wousp112:
Closing this: I tested the approach end to end against a live install and it does not work. Merging the definitions is necessary but not sufficient, and I'd rather not leave a proposal in your queue that I've since disproved.

## What the test showed

With the branch installed and `microsoft_outlook_email` selected, the model does see the tools — asked to name them, Grok 4.6 listed `list_messages`, `draft_email`, and the rest. The relay also does its job: the call reaches Codex in the native shape.

```
CALL name='list_messages' namespace='microsoft_outlook_email'
   OUT: unsupported call: microsoft_outlook_emaillist_messages
```

Codex rejects it. Its tool router only dispatches tools it has actually loaded into the session, so a definition that arrives only in the outbound request is callable by the model and unknown to the client.

I checked this wasn't just a naming mistake on my side. A live capture of what the client sends showed connector namespaces are named for the connector alone, not `mcp__codex_apps__<connector>`:

```
namespaces=["collaboration=>[...]", "codex_app=>[...]",
            "plugin_management=>[\"uninstall_plugin\"]"] tool_search=true
```

(Worth knowing separately: the `mcp__codex_apps__github` fixture in `test/namespace-relay-routing.test.mjs` does not match what codex-cli 0.146.0 sends.)

I corrected the naming to match and it still failed. The decisive test was `plugin_management`, the one connector namespace that *is* loaded — its `uninstall_plugin` arrives inline in every request. Merging in its three deferred siblings and calling one:

```
unsupported call: plugin_managementget_app_permissions
```

Same rejection, inside a namespace the client already has. So this is not about the namespace being absent or misnamed. Codex registers connector tools per loaded tool, and `tool_search` is what loads them.

## What this means for #252

The bug is real and this rules out the cheap fix. Of the three options I listed there, 1 and 2 are dead — no amount of definition merging makes these dispatchable. Only a router-side stand-in for `tool_search` can work, because the client has to be the one that loads the tool.

That direction now looks more promising than when I wrote it up, for a reason I had backwards: the search control isn't something the router needs to *answer*, only to *translate*. Codex executes `tool_search` itself, the same way it executes any other client-side tool. Relaying it to the provider as an ordinary function and mapping the model's call back to the native shape would let the model load a connector on demand and dispatch it on the next turn — which is exactly the native flow. The open question is the native call shape, which you're far better placed to know than I am.

Happy to try that if it's a direction you'd take. Leaving #252 open with these findings added.

PR #255, opened 2026-08-16
Bridge native tool_search for routed chat-completions models

### LINKED ISSUE #252: Routed models reach only the deferred subset of codex_apps connector tools

labmimors:
## Summary

- Lower Codex client-executed native tool_search into a collision-safe ordinary function for routed chat-completions providers.
- Restore streamed and non-streamed provider calls to native tool_search_call items so the Codex client performs discovery and registration.
- Replay only adjacent, matching client call/output history; expose accepted discovered function schemas on the following routed request and extend the exact request-local namespace lookup.
- Keep Responses-native routes unchanged and fail closed on malformed history, server execution, custom/freeform tools, and discovered-name collisions.

## Why

Codex 0.146 registers deferred connector tools only after its native tool_search flow. Copying cached definitions into the request can make a model see a tool that the client still rejects as unsupported. This change preserves the client-owned discovery and execution boundary while adapting only the provider-facing wire shape.

Stable Codex contract references:

- Native tool declaration: https://github.com/openai/codex/blob/rust-v0.146.0/codex-rs/tools/src/tool_spec.rs#L18-L29
- Client tool_search handler: https://github.com/openai/codex/blob/rust-v0.146.0/codex-rs/core/src/tools/handlers/tool_search_spec.rs#L18-L76
- Native dispatch path: https://github.com/openai/codex/blob/rust-v0.146.0/codex-rs/core/src/tools/router.rs#L144-L162
- Follow-up history contract: https://github.com/openai/codex/blob/rust-v0.146.0/codex-rs/core/tests/suite/search_tool.rs#L745-L815

## Verification

- npm ci --ignore-scripts: 0 vulnerabilities
- node --test test/namespace-relay.test.mjs test/namespace-relay-routing.test.mjs: 35/35 passed
- npm run check: passed
- npm test on current main: 1,338 total; 1,326 passed; 12 skipped; 0 failed
- Independent read-only review: PASS after tightening replay to adjacent call/output pairs

The repository-gated real LiteLLM adapter integration and a quota-consuming live provider probe were not run. Mock-gateway coverage exercises routed SSE, non-streaming JSON, next-turn replay, discovered namespace dispatch, collision handling, and the Responses-native control path.

Disclosure: I maintain dsh-mcp-lens and am contributing here because the same deferred tool-discovery boundary affects routed DeepSeek Harness and Codex workflows. This patch does not add or depend on MCP Lens.

Fixes #252.

### DIFF
--- a/src/namespace-relay.mjs
+++ b/src/namespace-relay.mjs
@@ -42,6 +42,47 @@ export const NAMESPACE_DELIMITER = "__";
 // Map in a WeakMap preserves the Map's public shape for existing callers while
 // letting the response path validate model-generated overrides.
 const SPAWN_AGENT_MODELS = new WeakMap();
+const TOOL_SEARCH_RELAYS = new WeakMap();
+
+const TOOL_SEARCH_FUNCTION_NAME = "tool_search";
+
+function providerFunctionName(tool) {
+  return tool?.name ?? tool?.function?.name;
+}
+
+function providerVisibleToolNames(tools) {
+  const names = new Set();
+  if (!Array.isArray(tools)) return names;
+  for (const tool of tools) {
+    if (tool?.type === "namespace" && Array.isArray(tool.tools)) {
+      for (const fn of tool.tools) {
+        if (fn?.name) names.add(`${tool.name}${NAMESPACE_DELIMITER}${fn.name}`);
+      }
+      continue;
+    }
+    const name = providerFunctionName(tool);
+    if (typeof name === "string" && name) names.add(name);
+  }
+  return names;
+}
+
+function availableToolSearchName(tools) {
+  const names = providerVisibleToolNames(tools);
+  if (!names.has(TOOL_SEARCH_FUNCTION_NAME)) return TOOL_SEARCH_FUNCTION_NAME;
+  let suffix = 1;
+  while (names.has(`codex_tool_search_${suffix}`)) suffix += 1;
+  return `codex_tool_search_${suffix}`;
+}
+
+function providerToolSearchDescription(description, providerName) {
+  if (typeof description !== "string") return undefined;
+  if (providerName === TOOL_SEARCH_FUNCTION_NAME) return description;
+  const rewritten = description.replaceAll(
+    `\`${TOOL_SEARCH_FUNCTION_NAME}\``,
+    `\`${providerName}\``,
+  );
+  return `${rewritten}\n\nFor this routed request, call \`${providerName}\` for deferred tool discovery; \`${TOOL_SEARCH_FUNCTION_NAME}\` is a separate ordinary function.`;
+}
 
 function schemaStringValues(schema, values = new Set()) {
   if (!schema || typeof schema !== "object") return values;
@@ -128,23 +169,51 @@ export function repairToolSchemaRoots(tools) {
   return changed ? repaired : tools;
 }
 
+function flattenNamespaceChild(namespace, fn) {
+  const clientSchema = fn.parameters ?? fn.inputSchema;
+  const parameters =
+    clientSchema === undefined ? undefined : providerToolSchema(clientSchema);
+  return {
+    ...fn,
+    name: `${namespace}${NAMESPACE_DELIMITER}${fn.name}`,
+    ...(parameters === undefined ? {} : { parameters }),
+  };
+}
+
 // Flatten every namespace entry into plain functions named
 // `<namespace>__<tool>`. Returns the set of namespaces that were flattened
 // (name -> tool names) so callers can rename history and restore calls.
-export function flattenNamespaceTools(tools) {
+export function flattenNamespaceTools(tools, { bridgeToolSearch = true } = {}) {
   if (!Array.isArray(tools)) return { tools, flattened: false, namespaces: new Map() };
   const flattened = [];
   const namespaces = new Map();
   const spawnAgentModels = new Set();
+  const toolSearchName = bridgeToolSearch ? availableToolSearchName(tools) : undefined;
+  let toolSearchRelay;
   let changed = false;
   for (const tool of tools) {
-    // Codex adds this control tool when one or more namespace tools use
-    // deferred loading. By this boundary every namespace is expanded below,
-    // so the search control is redundant; chat-completions providers accept
-    // only ordinary function tools and reject `type: "tool_search"` before
-    // the model can call any MCP function.
+    // Codex registers deferred tools client-side and exposes this native
+    // control so the model can search them on demand. Chat-completions
+    // providers reject the native type, so present the same request-local
+    // capability as an ordinary function and restore its calls on the
+    // response path. Only the native client-executed shape enables the relay;
+    // an unrelated function named `tool_search` never does. When such a plain
+    // function already exists, use a deterministic alias so neither call can
+    // hijack the other.
     if (tool?.type === "tool_search") {
       changed = true;
+      if (bridgeToolSearch && tool.execution === "client" && !toolSearchRelay) {
+        const parameters =
+          tool.parameters === undefined ? undefined : providerToolSchema(tool.parameters);
+        const description = providerToolSearchDescription(tool.description, toolSearchName);
+        flattened.push({
+          type: "function",
+          name: toolSearchName,
+          ...(description === undefined ? {} : { description }),
+          ...(parameters === undefined ? {} : { parameters }),
+        });
+        toolSearchRelay = { providerName: toolSearchName };
+      }
       continue;
     }
     if (tool?.type === "namespace" && Array.isArray(tool.tools)) {
@@ -166,14 +235,7 @@ export function flattenNamespaceTools(tools) {
         // never touches automations still dies on its first message. Normalize
         // only the provider-facing copy; `inputSchema` stays exactly as the
         // client sent it.
-        const clientSchema = fn.parameters ?? fn.inputSchema;
-        const parameters =
-          clientSchema === undefined ? undefined : providerToolSchema(clientSchema);
-        flattened.push({
-          ...fn,
-          name: `${tool.name}${NAMESPACE_DELIMITER}${fn.name}`,
-          ...(parameters === undefined ? {} : { parameters }),
-        });
+        flattened.push(flattenNamespaceChild(tool.name, fn));
         names.add(fn.name);
         if (tool.name === "collaboration" && fn.name === "spawn_agent") {
           schemaStringValues(fn.inputSchema?.properties?.model, spawnAgentModels);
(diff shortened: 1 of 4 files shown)


duolahypercho:
Reviewed this against current `main` (it merges cleanly and `npm test` is 1339/0 failed with the merge in place). The design is right, and it is the same conclusion #253's author reached independently after disproving the definition-merging approach — the client has to be the one that loads the tool, so translating the control rather than answering it is the only door left. The fail-closed instincts throughout are good: `execution: "client"` gating so an unrelated function named `tool_search` never enables the relay, the deterministic alias on collision, live schemas winning a name collision, and `allowIncompleteToolSearch` scoped to `response.output_item.added`.

One defect, and I think it is load-bearing.

## Pairing is positional, so a batched search turn silently degrades

`flattenToolSearchHistory` matches a call to its output by **adjacency** (`input[index]` / `input[index + 1]`) and additionally requires `call_id` equality. When the model issues two `tool_search` calls in one turn, Codex emits them the way it emits any parallel tool calls — all calls, then all outputs — and adjacency never holds:

```js
const parallel = [
  { type:'tool_search_call',   execution:'client', call_id:'c1', arguments:{query:'mail'} },
  { type:'tool_search_call',   execution:'client', call_id:'c2', arguments:{query:'calendar'} },
  { type:'tool_search_output', execution:'client', call_id:'c1', tools:[{type:'function',name:'list_messages'}] },
  { type:'tool_search_output', execution:'client', call_id:'c2', tools:[{type:'function',name:'list_events'}] },
];
flattenToolSearchHistory(parallel, tools, namespaces);
// types after bridge: tool_search_call, tool_search_call, tool_search_output, tool_search_output
// tools added to request: 0
// flattened flag: false
```

Two consequences, and the second is the one that matters:

1. Four native Responses item types reach a chat-completions provider verbatim. Nothing downstream strips them — I grepped; `flattenToolSearchHistory` is the only code in `src/` that touches either type.
2. **Zero discovered tools are added to the request**, while the model has already been told in its own transcript that it discovered them. Its next call names a tool that is not in the tool list — which is the same "model can see it, client rejects it" failure this PR exists to fix, reintroduced through a different door.

The same thing happens to a pair split by remote compaction, in either direction:

```js
// output survived, its call did not  -> stays tool_search_output
// call survived, its output did not  -> stays tool_search_call
```

Both were returned unchanged in my run.

## Suggested fix

Pair by `call_id` rather than by position: walk the input once, record every well-formed `tool_search_call` id, and match each `tool_search_output` to a recorded id that appears **before** it. That keeps the property the adjacency tightening was reaching for — only matched, well-formed pairs replay, orphans never add tools — without making it positional. Orphans probably also want to be dropped rather than forwarded, since a native item type in a chat-completions body is not something the provider can do anything with.

Worth a test for the batched shape either way; the current suite only exercises the adjacent one, which is why this passes 35/35 today.

## Smaller notes

- `providerVisibleToolNames` is computed once before the `map`, then mutated as outputs are accepted. That is correct for "earlier discovery wins", but it does mean two outputs discovering the same tool report it in the first `function_call_output` and omit it from the second. Probably what you want; worth a comment saying so, because the omission reads as a bug from the model's side.
- The non-streaming path (`namespace-relay.mjs:594`) correctly omits `allowIncompleteToolSearch`, so a malformed-argument call stays an ordinary `function_call` there. That degrades to an "unsupported call" at the client rather than a malformed `tool_search_call`, which is the right trade — but it is silent. A log line would make it diagnosable.

Thanks for the Codex contract links and for stating what you did not run; that made this much faster to check.

---
🤖 Automated review by [Claude Code](https://claude.com/claude-code) — the repro above was executed against this branch merged with current `main`.

labmimors:
Thanks — I reproduced the exact calls-all-then-outputs shape and fixed the pairing on the rebased branch.

`flattenToolSearchHistory` now makes one forward pass keyed by `call_id`: an output can match only one well-formed client call that appeared earlier. Parallel calls followed by parallel outputs work; duplicate, reversed, orphaned, malformed, or non-client native items are dropped fail-closed and cannot inject schemas into the provider request. A valid compacted pair with `tools: []` becomes ordinary function history but revives no tools. I also documented that the first accepted discovery wins when later outputs return the same provider-visible name.

Added regressions for the batched shape, late duplicate invalidation, orphan/split history, malformed and server-executed items, duplicate discoveries, and empty-tool compaction. Current head: `2244d3c`.

Verification on Node 25.1.0:
- targeted relay/routing: 38/38 pass
- full suite: 1,389 total, 1,377 pass, 0 fail, 12 skip
- `npm run check` and `git diff --check`: pass

Boundary: I still did not run the opt-in real LiteLLM/provider integration; those integration cases remain among the skips.

labmimors:
Thanks, this was the right blocker. The current head now pairs native tool_search history by call_id instead of adjacency, keeps only ordered/well-formed pairs, and drops orphaned/duplicate/malformed native items before the chat-completions bridge can see them. I also verified the relevant coverage locally on the latest PR head: ✔ flattenNamespaceTools flattens every namespace, including MCP ones (2.910875ms)
✔ flattenNamespaceTools keeps the full tool schema on flattened entries (0.214167ms)
✔ flattenNamespaceTools preserves a supplied provider parameter schema (0.139833ms)
✔ flattenNamespaceTools exposes client tool_search as an ordinary provider function (0.205875ms)
✔ tool_search bridge uses a collision-safe request-local name (0.661791ms)
✔ flattenNamespaceTools handles non-array and empty input (0.105792ms)
✔ full inventory survives merge + flatten with nothing dropped (1.183125ms)
✔ stored namespaced calls are renamed to match the flattened tools (0.768ms)
✔ matched tool_search history declares discovered tools and expands namespace lookup (0.61425ms)
✔ parallel tool_search calls pair by call_id when all outputs follow the calls (0.262708ms)
✔ orphaned, malformed, and no-control native tool_search history is dropped (0.201125ms)
✔ compacted tool_search history keeps an ordered pair without reviving tool schemas (0.144292ms)
✔ duplicate ids, reversed order, and malicious outputs invalidate the whole id (0.316917ms)
✔ history rename is idempotent and leaves other namespaces alone (0.125333ms)
✔ history rename recovers calls stored without a namespace field (0.076083ms)
✔ response transform restores flattened calls to the native namespace shape (3.926208ms)
✔ tool_search response bridge covers SSE added, delta, done, and completed (0.941208ms)
✔ tool_search response bridge fails closed without native control or valid arguments (0.176833ms)
✔ response transform restores namespace on unambiguous unprefixed calls (0.77275ms)
✔ response transform drops a spawn-agent model override not offered by the tool schema (0.149ms)
✔ response transform detects headerless SSE after split framing prelude (0.942ms)
✔ response transform restores declared and headerless non-streaming JSON output (1.06675ms)
✔ non-streaming rewrite covers nested output and leaves malformed JSON untouched (0.27675ms)
✔ response transform leaves ambiguous and ordinary calls alone (0.363917ms)
✔ rewriteNamespaceFunctionCall rejects non-call events (0.07125ms)
✔ response rewrite turns Grok whole-float tool arguments into integers (0.420917ms)
✔ response transform rewrites native shell_command integer floats in SSE (0.361083ms)
✔ every flattened app tool reaches the provider with an object root (0.794417ms)
✔ flattened parameters drop literals that contradict their declared type (0.196ms)
✔ a plain function tool's union root is repaired too (0.124458ms)
✔ a plain function tool's nullable root is repaired too (0.083625ms)
✔ an ordinary function tool is passed through by identity (0.057917ms)
✔ repairToolSchemaRoots fixes roots without flattening (0.089666ms)
✔ repairToolSchemaRoots returns the original array when nothing needs repair (0.045083ms)
ℹ tests 34
ℹ suites 0
ℹ pass 34
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 95.393208 and ✔ routed request flattens every namespace to the gateway and restores calls to the client (481.326333ms)
✔ non-streaming routed responses restore namespace calls before client dispatch (193.49675ms)
✔ routed tool_search history declares discovered tools and restores their calls (389.901375ms)
✔ Responses-native routed providers inherit the model on fresh local thread calls (386.579625ms)
ℹ tests 4
ℹ suites 0
ℹ pass 4
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 1535.131291 both pass, including the parallel-output shape and the compacted/orphan cases you called out.

PR #272, opened 2026-08-16
feat(gemini): publish routed models into Gemini CLI

duolahypercho:
Adds `gemini` as a target alongside `codex` and `dsh`, so the router's models
show up in [Gemini CLI](https://github.com/google-gemini/gemini-cli).

**Please test this if you have Gemini CLI installed** — see "How to try it" below.
It is verified end to end on macOS with gemini-cli 0.27.3, but only against the
providers on one machine, and `install.ps1` has never been run.

## Why it needs a new surface

Gemini CLI speaks only the Gemini API, and Google
[does not support](https://discuss.ai.google.dev/t/antigravity-add-your-own-api-keys-models/137068)
bring-your-own-provider. But with `GEMINI_API_KEY` set, `createContentGenerator`
in `@google/gemini-cli-core` builds a plain `@google/genai` client and calls
`{GOOGLE_GEMINI_BASE_URL}/v1beta/models/{model}:{method}` — so a Gemini-shaped
endpoint is all that is missing.

The router now serves one at `/_codex-router/<key>/gemini`, behind the same
caller capability as `/v1` and `/panel`.

## It is not a second request path

`gemini-surface.mjs` translates the turn into a Responses request, sends it
through the router's existing `/v1/responses` **over the loopback**, and
translates the answer back. Tool-result ageing, the vision bridge, prompt-token
substitution, upstream retry, model failover, and usage accounting all still sit
on one request path. The loopback deliberately carries no credential — the key
the CLI presents *is* the caller capability, and leaving the header off is also
what lets native GPT models reach the ChatGPT-session fallback.

## What it writes

One marker block in `~/.gemini/.env`, and nothing else:

```sh
# BEGIN codex-router-gemini
GOOGLE_GEMINI_BASE_URL=http://127.0.0.1:4202/_codex-router/<caller-key>/gemini
GEMINI_API_KEY=<caller-key>
GEMINI_MODEL=anthropic/claude-opus-4-6
# END codex-router-gemini
```

`settings.json` is **never opened for writing** — it is JSONC carrying the
user's own comments, and a `JSON.parse`/`stringify` round trip would delete
every one of them. The block is 0600 because it holds the caller key,
publishing twice is byte-identical, and removing it restores the file exactly.
A managed key assigned outside the block stops the publish with the line named:
`dotenv` lets the last assignment win, so a duplicate would silently decide
which endpoint is in force.

## How to try it

```sh
./bin/model-router gemini enable
./bin/model-router gemini doctor
cd /some/project && gemini -p "what does this repo do?"
```

If the CLI asks how to authenticate, choose **Use Gemini API key** once — the
key is the router's local caller capability, not a Google one, and it never
leaves the machine. To undo: `./bin/model-router gemini disable`.

Please report: which provider/model you used, whether tool calls worked, and
anything `gemini doctor` said.

## Verified

Driven with the real `gemini` binary against real providers:

| Check | Result |
|---|---|
| `GET /v1beta/models` | 84 models |
| `:countTokens` | answers without spending a turn |
| `:streamGenerateContent` (`qwen-plan/qwen3.6-flash`) | correct text, reasoning as `thoughtsTokenCount`, `finishReason: STOP` |
| real CLI, plain turn | passed |
| real CLI, **tool calling** (`zai-coding/glm-5.3`) | full loop: 10 tool schemas out → call in → result out → answer in |
| provider 503 / quota 403 | relayed as `UNAVAILABLE` / `PERMISSION_DENIED` with the router's own message |

That live run caught the one bug the unit tests could not: a Gemini tool
declares its schema as **`parametersJsonSchema`**, not `parameters`
(`gemini-cli-core/dist/src/tools/tools.js:194`). The first cut read only
`parameters`, so every tool went upstream with no schema, the model invented
argument names, and the CLI rejected each call with
`params must have required property 'file_path'`.

## Deliberate limits

- **Embeddings refused** with a named 501 — no routed provider exposes one, and
  a fabricated vector is worse than an error. Gemini CLI calls `embedContent`
  only from `baseLlmClient`, never from the turn loop.
- **`countTokens` is estimated** rather than answered by spending a real turn.
- **The default model is written**, unlike the harness integration's opt-in
  equivalent, because Gemini CLI's own default is a Gemini model this router
  does not route — an install that left it alone would 404 on the first turn.
  `--model` still outranks it.
- **`~/.gemini/.env` loses to a project `.env`**: `findEnvFile` walks up from
  cwd first. Inside a repo that has its own `.env`, the router's block is
  ignored. Documented rather than worked around.
- **No tray surface.** The Swift tray reads `targets["codex"]` only, so Gemini
  will not appear there. Additive, nothing breaks.

## Also in here

`routed-client-models.mjs` extracts the "what may be published to a client that
carries no ChatGPT session of its own" rule out of the harness manager. It was
always a general rule; a second client wanting it verbatim made a second copy
the wrong answer.

## Tests

77 new tests. Full suite **1443 pass / 0 fail** on this branch; base `ef7995a`
is 1366 / 0.

https://claude.ai/code/session_01KKT3Zabs9wpfa3UEWgmaVV

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,23 +4,24 @@ These instructions apply when a user asks an agent to install this repository.
 
 ## Choose the target
 
-- `codex` (the Codex CLI and desktop app) and `dsh` (DeepSeek Harness) are the
+- `codex` (the Codex CLI and desktop app), `dsh` (DeepSeek Harness), and
+  `gemini` (Gemini CLI) are the
   supported targets. If the user asks for Cursor or opencode integration,
   explain that those targets were removed and the router does not have them;
   the opencode provider (the Go subscription and the pay-per-use Zen endpoint)
   remains available as a provider inside both targets.
-- **A target is a client, not a router.** One installation serves both: one
-  background service, one gateway, one set of provider credentials, one
+- **A target is a client, not a router.** One installation serves all of them:
+  one background service, one gateway, one set of provider credentials, one
   provider selection, one set of ports. `MODEL_ROUTER_TARGET` selects which
   client's configuration a command writes. It must never fork the state
-  directory, the service, or the credential store — a user who installs both
+  directory, the service, or the credential store — a user who installs two
   would otherwise be asked for every API key twice and would run two gateways
   against one set of provider quotas. `ROUTER_PLANE_TARGET` in
   `src/paths.mjs` names that shared plane, and the environment aliases and the
   `/health` service name are keyed on it rather than on the client.
-- Installing both is normal and needs no special handling: run the install
-  once per target. Whichever one is already present is republished whenever the
-  routable set changes, so the two clients cannot drift apart.
+- Installing more than one is normal and needs no special handling: run the
+  install once per target. Whichever ones are already present are republished
+  whenever the routable set changes, so the clients cannot drift apart.
 
 ## Codex outcome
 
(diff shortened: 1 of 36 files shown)


duolahypercho:
Note on the base branch: this targets `feat/model-failover` rather than `main`, because it was branched from `ef7995a` (that branch's tip) and targeting `main` bundled four unmerged failover commits into the diff. Retarget to `main` once the failover work lands — the Gemini changes touch none of the same files apart from `README.md` / `CHANGELOG.md`.

duolahypercho:
Rebased onto `main` and retargeted. The earlier base (`feat/model-failover`) was stale — that work landed in #260/#263 as squash commits, so its branch commits never became ancestors of `main` and an ancestry check wrongly read as unmerged.

One conflict, in `CHANGELOG.md`: additive only. The Gemini entry sits on top, and the four entries `main` gained meanwhile are kept below it. No lines from `main` were dropped in `src/doctor.mjs` or `src/router.mjs`.

Re-tested on the rebase: **1495 pass / 0 fail**, against a `main` baseline of **1418 / 0** — +77 tests, nothing broken.

duolahypercho:
CI is red on `test` for **all three platforms** on `ce9b742` — five tests in `test/gemini-surface.test.mjs`, every one that relays a turn. Root cause found, and it is not a rebase artifact:

**`src/gemini-surface.mjs:183` — `signal: request.signal`**

This is the only `request.signal` in the repository. CI runs **Node v24.19.0**, where `http.IncomingMessage` exposes a `signal` that has **already aborted** by the time the handler finishes `readRequestBody` and issues the loopback fetch. `fetch` throws `AbortError` immediately, the bare `catch {}` two lines down swallows it, and every turn becomes `502 "The local model router could not be reached."` The upstream never receives a request at all.

Reproduced by exact signature: shimming an aborted `IncomingMessage.prototype.signal` on Node 22 produces all five failures in the same order with the same messages — `502 !== 200`, `Cannot read properties of undefined (reading 'headers')`, `502 !== 200`, `…(reading 'candidates')`, `502 !== 429`.

**Why local runs look fine.** Node 22.23.1 and 25.6.0 lack the property entirely, and 26.7.0 has it non-aborted at that point. Only the Node 24 LTS line is affected. A green local suite is not evidence against this.

**Why it matters beyond CI.** `package.json` declares `"node": ">=22.19.0"`, so Node 24 is a supported runtime. On Node 24 the Gemini integration 502s on every single turn — the feature is non-functional there, not merely untested.

**Fix options**, in increasing order of correctness:

1. Guard it — `signal: request.signal?.aborted ? undefined : request.signal`. Verified: restores 13/13 under the Node-24 shim.
2. Drop `request.signal` entirely.
3. Wire an explicit `AbortController` off the response's `close` event.

Option 1 is the minimal unblock, but 2 or 3 is probably the right end state, since the current line cannot do what it intends on Node 24 anyway — it aborts the loopback fetch before it is ever issued rather than cancelling a genuinely abandoned turn.

Separately, and unrelated to this: the rebase itself is verified good. `git diff origin/main...ce9b742 --stat` is 36 files, all Gemini-related; the forbidden-path check is empty; and all nine files touched by #262 are byte-identical to `origin/main`, so the revert risk that the old `feat/model-failover` base carried is gone. That base would by now have reverted #262, #276, #264, #255, #274 and #275 — 41 files, 273 insertions against 3,633 deletions — so retargeting was the right call.

duolahypercho:
CI caught a real bug that local testing could not, now fixed.

The loopback call forwarded `request.signal`. That property does not exist before Node 22, and on **Node 24 — which is what CI runs — it aborts as soon as the request body has been read**, which is before the loopback call is even made. Every routed turn failed with an instant 502 on all three runners while passing locally on Node 22, where the property is simply absent.

The turn's lifetime is now tied to the response closing, which means the same thing on every supported version, and a failed loopback names its cause chain instead of swallowing it — the original `could not be reached` with no cause is why the first CI run had nothing in it to diagnose.

Two regression tests added: one drives the handler with an already-aborted `request.signal` (fails against the old code, passes now), one asserts the 502 carries its cause.

Verified on Node 22.23.1 and Node 26.7.0: **1497 pass / 0 fail**.

PR #279, opened 2026-08-16
feat(cursor): use a Cursor subscription as a provider via a local CLI bridge

duolahypercho:
Draft: the code is complete and the suite is green, but **no live Cursor generation has ever run** — the development account has no remaining usage. See "What is not verified" below before merging.

## For a user, with no terminal

Open the tray → **Connections** → **Cursor CLI (Local)** → **Install & Sign In**. That installs `cursor-agent` if missing, then opens a Terminal window into Cursor's browser sign-in. Afterwards the **Cursor CLI** section lists the account's models; tick the ones you want and they show up in the Codex picker.

The first revision of this PR made that button raise an error telling the user to run a curl command instead. That was a dead end for anyone not already comfortable in a terminal, and inconsistent besides — every other sign-in CLI here is installed by the same press. Fixed.

## Why a bridge, and not a target

Cursor CLI cannot be a router target, and this is not a gap waiting to be filled:

- `~/.cursor/cli-config.json` has no endpoint or provider keys — only `model`, `permissions`, `sandbox`, and display settings.
- `--endpoint` exists but speaks Cursor's own protocol, not anything OpenAI-shaped.
- `bedrock` mode validates the credential against Cursor's backend.
- The IDE's "Override OpenAI Base URL" is IDE-only and refuses private-network addresses.

So Cursor runs the other way round. `cursor-cli` is a keyless provider on a loopback address, like `local` and `lmstudio`, but what serves that address is the router's own bridge: it spawns `cursor-agent -p --output-format stream-json` per request and adapts it to chat completions. No credential is stored or read — the sign-in stays in cursor-agent's own store.

AGENTS.md records the above so nobody re-derives it.

## Design constraints held

- **Read-only, always.** `turnArguments()` pins `--mode ask --sandbox enabled` and never passes `--force`, `--yolo`, `--trust`, or `--auto-review`. This process takes prompt text off a socket and hands it to an agent holding shell and file tools.
- **Empty workspace.** The agent runs in a directory the router owns, not the caller's repository. A chat-completions request is a prompt; everything the model should see was put there by the caller.
- **Stateless turns.** History is rendered into the prompt; `--resume` is never passed, so identical requests cannot answer differently based on an unrelated caller's session.
- **Service-supervised.** `start.mjs` launches and health-gates the bridge. A manually started one is a reboot away from every Cursor model 502ing while the router reports itself healthy.
- **No `curl | bash` from a button.** Cursor does not ship on npm, so the tray's one-click install path refuses and names the command instead.

## Known limitation

`cursor-agent` emits text and its own tool events, never OpenAI `tool_calls`, and Codex dispatches every turn through tool calls. **Cursor models answer questions but cannot drive an agentic Codex turn.** Documented rather than left to be debugged.

## Bug fixed along the way

The tray reported Cursor as connected whenever the provider existed — `cursor-cli` is keyless, so the generic row read "no key needed" as "ready" while cursor-agent sat signed out. The row now reflects the CLI's real state and gains a working Sign in button, reusing the existing oauth rendering with no UI code.

## Tray

The Cursor row appears in Connections for free (the row list is built from the registry). Two things were added on top:

**A working Sign in button.** The tray already renders "Sign in" / "Reconnect" for any row reporting `kind: "oauth"`, so this needed no UI code — only making Cursor report an oauth-shaped row. It also fixed a bug: `cursor-cli` is keyless, so `providerNeedsNoKey` was true and the generic row reported Cursor as **connected** while cursor-agent sat signed out. The only way to discover that was every model 502ing.

Two Cursor-specific wrinkles: it does not ship on npm, so the tray installs it from Cursor's own script — fetched to a file and then run, never piped into a shell, with the URL pinned to `cursor.com` and the response checked for a shell shebang before anything executes; and it keeps its session in the OS credential store rather than a file, so the mtime probe that detects a finished sign-in had nothing to watch — `cursor-agent status` is used instead (local, ~0.5s, safe on a refresh; `models` is a network round trip and is not).

**A model curation section**, beside LM Studio's and built the same way, writing the same user-model overlay so the panel and `bin/curate-models cursor-cli` remain two doors to one state. It reads the bridge's own `/v1/models` rather than shelling out to the CLI, so a timed panel refresh costs a loopback GET behind the bridge's 60s cache instead of a round trip to Cursor. Three states are distinguished — bridge down, signed out, and listed — because the fixes differ.

Published entries are labelled "(Cursor, answers only)", since the picker is the last place someone learns these models cannot emit tool calls.

The native macOS tray needs nothing: it has no LM Studio section either, and `RouterProviderInfo.legacyFallback` exists for routers older than this feature.

## What is verified

- Bridge serves loopback; `/health`, `/v1/models`, and the port override confirmed against a running instance.
- The router's own `discoverProviderModels("cursor-cli")` reaches it and surfaces actionable reasons.
- The stream-json event schema was read out of cursor-agent 2026.08.11-e8db854's own bundle, not guessed — including the two facts the translation rests on: under `--stream-partial-output` an `assistant` event carries the delta, and `result` carries the complete text.
- The tray sign-in path drives `cursor-agent login` through a stub terminal launcher.
- Port-drift guard verified by injecting the drift and watching it fail.
- The tray sign-in path drives `cursor-agent login` through a stub terminal launcher.
- The panel snapshot was exercised against a running bridge and a stopped one, reporting both states distinguishably.
- The Rust model-id validator was compiled and exercised standalone (4 valid forms accepted, 7 injection/malformed refused) rather than trusting a full Tauri build.
- Panel wiring is guarded structurally and each guard was mutation-tested by breaking it.
- The installer was run for real into an isolated `HOME`: 10.4s, `cursor-agent` installed, resolved by `oauthCliPath`, `--version` reporting `2026.08.11-e8db854`. The development machine's own install was untouched.
- The installer origin guard is exported and tested: `cursor.com.evil.example`, a GitHub raw URL, and an `http://` downgrade are each refused before anything executes.
- Full suite green on current main: **1457 pass, 0 fail**.

## What is NOT verified

**One successful live generation.** Everything downstream of a real, authenticated, in-quota Cursor turn is unexercised. Do not merge as proven until someone runs:

```sh
cursor-agent login
./bin/model-router codex providers enable cursor-cli
./bin/curate-models cursor-cli
```

## Test plan

- [x] `node --test test/*.test.mjs` — 1457 pass, 0 fail
- [x] Bridge health, models, and port override against a running instance
- [x] Tray sign-in path via stub terminal launcher
- [ ] One live generation through a signed-in, in-quota Cursor account
- [ ] `bin/curate-models cursor-cli` against the real `/v1/models`
- [ ] A routed turn selected from the Codex picker
- [x] `cursor-agent` installed end to end through the tray's install path (isolated HOME)
- [ ] The tray Sign in button pressed against a real Terminal
- [ ] The panel curation section rendered in a rebuilt tray

https://claude.ai/code/session_01S2jpUWGe7PyfHLuFyeP9UD

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1015,9 +1015,17 @@ bridge is the integration.
    bridge not running both surface as "every Cursor model 502s", so the doctor
    names them separately -- and probes neither unless the provider is selected,
    because `cursor-agent models` is a network round trip to Cursor.
-8. The bridge default port (4209) and the `baseUrl` checked into
-   `config/cursor/cursor.json` are one fact in two files. Change both or
-   neither.
+8. **The service owns the bridge's lifetime, not the operator.** `start.mjs`
+   launches and health-gates it beside the other forwarders. A bridge somebody
+   starts by hand is one reboot away from every Cursor model 502ing while the
+   router reports itself healthy; `bin/cursor-bridge` is for debugging one in
+   the foreground, not the supported way to run it.
+9. The bridge default port (4209) and the `baseUrl` checked into
+   `config/cursor/cursor.json` are one fact in two files, and `start.mjs`
+   republishes the resolved address through the provider's own
+   `MODEL_ROUTER_CURSOR_BASE_URL` so a moved port cannot strand the registry
+   entry on the old one. `test/cursor-cli-bridge.test.mjs` asserts all three
+   agree; do not change one and leave the others.
 
 ## Codex safety boundaries
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1020,7 +1020,17 @@ bridge is the integration.
    starts by hand is one reboot away from every Cursor model 502ing while the
    router reports itself healthy; `bin/cursor-bridge` is for debugging one in
    the foreground, not the supported way to run it.
-9. The bridge default port (4209) and the `baseUrl` checked into
+9. **The tray installs it, and fetches before it runs.** Cursor is the one
+   sign-in CLI npm does not carry, so `installScriptCli` downloads the vendor
+   script and then executes it. Do not "simplify" that back into
+   `curl | sh`: a piped install that dies halfway still runs everything that
+   arrived, and what runs is whatever landed. The URL is a constant pinned
+   against `installHost` before anything executes, and the response is checked
+   for a shell shebang because a captive portal answers HTML with a 200.
+   Refusing to install at all was tried first and was wrong -- every other
+   sign-in CLI here is installed by the same button, and handing a user a
+   shell command instead of doing the work is a dead end.
+10. The bridge default port (4209) and the `baseUrl` checked into
    `config/cursor/cursor.json` are one fact in two files, and `start.mjs`
    republishes the resolved address through the provider's own
    `MODEL_ROUTER_CURSOR_BASE_URL` so a moved port cannot strand the registry
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,6 +9,16 @@ These instructions apply when a user asks an agent to install this repository.
   explain that those targets were removed and the router does not have them;
   the opencode provider (the Go subscription and the pay-per-use Zen endpoint)
   remains available as a provider inside both targets.
+- **Cursor cannot be a target, and this is not a gap waiting to be filled.**
+  Cursor CLI (`cursor-agent`) has no BYOK and no custom base URL: its
+  `~/.cursor/cli-config.json` reference carries `model`, `permissions`,
+  `sandbox`, and display keys and nothing that names an endpoint. `--endpoint`
+  exists but speaks Cursor's own protocol rather than anything OpenAI-shaped,
+  and Bedrock mode validates the credential against Cursor's backend. The
+  IDE's "Override OpenAI Base URL" is IDE-only and refuses private-network
+  addresses besides. Nobody should spend another afternoon re-deriving this.
+  Cursor runs the other way round instead — as the `cursor-cli` provider, see
+  below.
 - **A target is a client, not a router.** One installation serves both: one
   background service, one gateway, one set of provider credentials, one
   provider selection, one set of ports. `MODEL_ROUTER_TARGET` selects which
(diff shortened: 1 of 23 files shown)


duolahypercho:
Closing unmerged. The integration works — it was verified end to end against a signed-in account — but the economics do not, and that only became visible once a real turn ran.

## Why

`cursor-agent` is an agent, not a model endpoint. It prepends its own harness to every request, and that cannot be turned off on a standard plan.

Measured on live turns:

| Prompt sent | `prompt_tokens` billed |
|---|---|
| ~20 tokens ("What is 17 times 23?") | **22,166** |
| ~12 tokens ("Reply with exactly: …") | **22,162** |

A fixed floor of roughly **22,150 input tokens per turn**, whatever you ask. The same question to a normal provider bills ~20. On a plan denominated in credits tied to API cost, that is about a thousandfold markup on short prompts.

Two hidden flags looked like they would fix it. Both are refused by Cursor's backend, not the CLI:

- `--exclude-workspace-context` → `[invalid_argument] Workspace context exclusion is not allowed for this user, team, or selected model`
- `--system-prompt <file>` → `[invalid_argument] unknown option '--system-prompt'`

They are internal or enterprise-gated. There is no other route: Cursor publishes no inference API, and `--endpoint` speaks their own protocol.

Combined with the limitation already documented here — cursor-agent never emits OpenAI `tool_calls`, so these models cannot drive an agentic Codex turn — what remains is a provider that costs ~22K tokens a turn to answer questions and cannot do the main job. Not worth shipping.

## Worth keeping from this

Verified facts, so nobody re-derives them:

1. **Cursor CLI cannot be a router target.** No BYOK, no custom base URL. `~/.cursor/cli-config.json` has no endpoint keys, `--endpoint` speaks Cursor's own protocol, Bedrock mode validates against Cursor's backend, and the IDE's "Override OpenAI Base URL" is IDE-only and refuses private addresses.
2. **Cursor CLI as a provider costs ~22,150 prompt tokens per turn**, unavoidably, and the flags that would reduce it are server-gated.
3. **`cursor-agent` never emits OpenAI tool calls**, so it cannot drive Codex turns regardless of cost.

Also learned, and applicable beyond Cursor: `--stream-partial-output` does not replace cursor-agent's message-level emitter, it runs beside it, so naive concatenation doubles the output; usage is reported as camelCase in live results but snake_case in the bundle; and `--trust` is mandatory for any directory the CLI has not seen.

The branch is kept locally. Reopen from `feat/cursor-cli-provider-main` if Cursor ever exposes a raw endpoint or ungates the harness flags.

PR #295, opened 2026-08-17
fix: retry a Grok OAuth progress-only stop once under the author contract

Zzy-min:
## Why

#264 landed the parser/mapping fixes. Genuine progress-only stops remain: short visible text, no tool calls, high `output_tokens` because Grok counts reasoning in that total. The merge note asked for a later retry PR that treats the six review points as one contract, plus a narrower `dispatchSseBlock` try/catch.

## Contract

1. **Tools gate.** Retry only when the client request offered function tools. A no-`tools` turn is never retried, even if hosted search is injected.
2. **Keep the first answer.** Prefer the retry only when it actually produces a tool call. A second short status sentence is discarded.
3. **Bounded hold, live stream afterwards.** The first byte is held only while a retry is still legal. The hold ends on a tool call, visible text above 120 characters, 1 MiB, or 30 s. Requests without client tools stream live, unchanged from #264.
4. **Logged and metered.** The retry log line is not gated on `MODEL_ROUTER_QUIET`. Emitted `usage` sums both attempts and sets `retries: 1`; `normalizeTokenUsage` forwards that into the router usage event.
5. **Kill switch.** `CODEX_ROUTER_GROK_PROGRESS_ONLY_RETRY=0` disables it. Thresholds are overridable.
6. **Drain a failed retry body.**

The nudge is a trailing **user** message, not `developer`/`system`, so `toResponsesRequest` does not fold it into `instructions` and cold-miss the #265 prefix cache. `x-grok-conv-id` is still derived from the original opening messages.

`dispatchSseBlock` now catches only `JSON.parse`. A handler exception is no longer swallowed as a malformed event.

## Liveness deviation, stated plainly

`EmptyCompletionGuard` releases on reasoning liveness because that path has something to show the client. This forwarder speaks Chat Completions to Codex and has no reasoning channel. Releasing on Grok reasoning would send the first client byte and make a silent retry illegal — which is the original repro. Reasoning therefore does not end the hold; the 30 s / 1 MiB caps do.

## Tests

Targeted suite: grok-oauth turn + forwarder + response-usage, 64 pass, 0 fail. `npm run check` passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,18 +2,18 @@
 
 ## Unreleased
 
-- **Grok OAuth retries a genuine progress-only stop once, under the same
-  contract as empty-completion retry.** When the client offered tools and the
-  first turn ends with short visible text, no tool calls, and enough output
-  tokens to be reasoning-heavy, the forwarder holds the first byte, retries
-  once with a trailing user nudge (not a developer/system prefix rewrite),
-  keeps the first answer unless the retry actually calls a tool, sums both
-  attempts into `usage` with `retries: 1`, and logs the retry even when
-  `MODEL_ROUTER_QUIET=1`. The hold is capped at 1 MiB / 30 s and ends as soon
-  as a tool call or a long visible answer appears. Set
+- **Grok OAuth retries a genuine progress-only stop once, without holding
+  the first byte.** Attempt 1 streams live. If the client offered tools and
+  the turn ends with short visible text, no tool calls, and enough output
+  tokens to be reasoning-heavy, the forwarder retries once with a trailing
+  user nudge and *appends* only the retry's tool-call deltas plus
+  `finish_reason: "tool_calls"` onto the same open stream. The first answer
+  is kept when the retry also has no tools. Both attempts are summed into
+  `usage` with `progress_only_retried: true`; that marker is what
+  `acceptedInputTokens` excludes, not a bare transport `retries` count. The
+  retry log is not gated on `MODEL_ROUTER_QUIET`. Set
   `CODEX_ROUTER_GROK_PROGRESS_ONLY_RETRY=0` to pay once and see the raw
-  first attempt. `dispatchSseBlock` now catches only `JSON.parse`, so a
-  handler bug is no longer swallowed as a malformed event.
+  first attempt. `dispatchSseBlock` now catches only `JSON.parse`.
 
 - **The Devin CLI probe no longer reports "unknown" for a Devin CLI that is
   installed and working.** `devinCliVersion` was the one call site out of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,18 +2,27 @@
 
 ## Unreleased
 
-- **Grok OAuth retries a genuine progress-only stop once, without holding
-  the first byte.** Attempt 1 streams live. If the client offered tools and
-  the turn ends with short visible text, no tool calls, and enough output
-  tokens to be reasoning-heavy, the forwarder retries once with a trailing
-  user nudge and *appends* only the retry's tool-call deltas plus
-  `finish_reason: "tool_calls"` onto the same open stream. The first answer
-  is kept when the retry also has no tools. Both attempts are summed into
-  `usage` with `progress_only_retried: true`; that marker is what
-  `acceptedInputTokens` excludes, not a bare transport `retries` count. The
-  retry log is not gated on `MODEL_ROUTER_QUIET`. Set
-  `CODEX_ROUTER_GROK_PROGRESS_ONLY_RETRY=0` to pay once and see the raw
-  first attempt. `dispatchSseBlock` now catches only `JSON.parse`.
+- **Grok OAuth retries a progress-only stop once, without holding the first
+  byte.** Attempt 1 streams live. If the client offered tools and the turn
+  ends with short visible text, no tool calls, and enough output tokens to be
+  reasoning-heavy, the forwarder retries once with a trailing user nudge and
+  *appends* only the retry's tool-call deltas plus `finish_reason:
+  "tool_calls"` onto the same open stream. The first answer is kept when the
+  retry also has no tools. Both attempts are summed into `usage` with
+  `progress_only_retried: true`; that marker is what `acceptedInputTokens`
+  excludes, not a bare transport `retries` count. The retry log is not gated
+  on `MODEL_ROUTER_QUIET`. Set `CODEX_ROUTER_GROK_PROGRESS_ONLY_RETRY=0` to
+  pay once and see the raw first attempt. `dispatchSseBlock` now catches only
+  `JSON.parse`.
+
+  The trigger is a shape and cannot be anything else: a finished task answered
+  in one line — "Yes, that is correct." after 1,500 reasoning tokens — is
+  indistinguishable from a turn that stopped early, so it is retried too. The
+  nudge therefore offers the no-tool branch first ("if that already completed
+  the task, restate the final answer and call no tool"), which routes the
+  finished case into keep-first. An imperative nudge makes such a turn invent
+  a tool call, and the forwarder would graft it onto the answer for the client
+  to run. A false positive now costs one round trip, not a wrong action.
 
 - **The Devin CLI probe no longer reports "unknown" for a Devin CLI that is
   installed and working.** `devinCliVersion` was the one call site out of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+- **Grok OAuth retries a genuine progress-only stop once, under the same
+  contract as empty-completion retry.** When the client offered tools and the
+  first turn ends with short visible text, no tool calls, and enough output
+  tokens to be reasoning-heavy, the forwarder holds the first byte, retries
+  once with a trailing user nudge (not a developer/system prefix rewrite),
+  keeps the first answer unless the retry actually calls a tool, sums both
+  attempts into `usage` with `retries: 1`, and logs the retry even when
+  `MODEL_ROUTER_QUIET=1`. The hold is capped at 1 MiB / 30 s and ends as soon
+  as a tool call or a long visible answer appears. Set
+  `CODEX_ROUTER_GROK_PROGRESS_ONLY_RETRY=0` to pay once and see the raw
+  first attempt. `dispatchSseBlock` now catches only `JSON.parse`, so a
+  handler bug is no longer swallowed as a malformed event.
+
 - **The Devin CLI probe no longer reports "unknown" for a Devin CLI that is
   installed and working.** `devinCliVersion` was the one call site out of
   twenty that took `command` and `args` from `spawnableCommand` and threw away
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -304,17 +304,16 @@ arguments without holding the full turn, so Codex sees the tool call instead of
 mistaking the preceding status text for the completed task.
 
 A remaining Grok OAuth shape is a genuine progress-only stop: the model
-reasons, emits a short status sentence, and never calls a tool. When the
-client actually offered tools, the forwarder holds the first relayed byte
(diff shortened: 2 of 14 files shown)


duolahypercho:
Thanks for taking the contract seriously — five of the six points are met cleanly, and two things in here are exactly right: the conv-id / instructions-prefix reasoning, and the `dispatchSseBlock` narrowing. I verified both rather than taking them on trust.

Point 3 is not met, though, and it is a **data-loss regression** rather than a gap. Three blockers below, then two smaller items.

## Blocker 1 — held streaming deltas are dropped

The final `emitPendingDeltas()` at `src/grok-oauth-forwarder.mjs:545` returns immediately because `holding` is still `true` — nothing clears it after `consumeResponsesStream` returns. `beginStream()` on the next line then sets `holding = false`, writes the head and role chunk, and the code proceeds straight to the finish chunk. **Every buffered delta is discarded.**

Reproduced with three probes against your head:

- **Streaming, tools offered, short answer** (`"Done."`, 5 output tokens, no retry): client receives role chunk → `finish_reason: "stop"` → usage → `[DONE]`. The answer text never arrives. The same request on the merge base `16b68f7` streams `{"content":"Done."}` correctly — so this is introduced here, on the most common shape there is, since Codex always offers tools.
- **Streaming + retry preferred**: `finish_reason: "tool_calls"` with **zero tool_call deltas** — a protocol-invalid response.
- **Streaming + keep-first**: the first answer's text is dropped.

One-line fix, verified: insert `releaseHold();` before the `emitPendingDeltas()` at line 545. All three probes then pass.

The suite stays green because **every retry test in this PR uses `stream: false`** (the non-stream path reads `turn.contentText`, not `turn.deltas`), and the two pre-existing streaming tests either offer no tools (`:451`, so the hold never arms) or emit a tool call that releases it (`:515`). Please add streaming coverage for all three shapes above.

## Blocker 2 — the liveness design, which I have to push back on harder than "stated plainly"

It is worse than the PR describes, because `beginStream()` is what calls `response.writeHead(200)`. Holding means the client gets **no HTTP response head at all** — not a silent stream, a silent socket.

And the forwarder handles no reasoning events whatsoever (`applyResponsesEvent`, `grok-oauth-turn.mjs:208-255`, has no reasoning case), so during Grok's thinking phase there are no deltas at all, `shouldReleaseProgressOnlyHold` cannot fire, and 1 MiB of raw SSE is far more than a reasoning summary produces. The only releases that matter are the first visible token / tool call, or the 30 s cap.

Measured against a backend that reasons for 3 s then answers:

```
no tools  : first byte 25 ms
with tools: first byte 3,070 ms
```

Time-to-first-byte becomes the model's full thinking latency. grok-4.6 ships `defaultEffort: "high"` with `xhigh` available, and your own fixture is 1,620 reasoning tokens against 40 visible ones — so the cap is reached routinely, not in the tail. This is the same regression `empty-completion-guard.mjs:64-73` already measured and rejected (517 ms → 30,638 ms), reintroduced on a model chosen for reasoning depth. It also inflates `firstTokenMs` on every held turn, and non-retried held turns are not excluded by `provider-usage.mjs`'s `measurable` gate, so grok-4.6 would start reporting a fabricated multi-second TTFT in the meter.

Traded against the bug being fixed — one wasted round trip when Grok says "Next I will update the deck." and stops — 30 s of dead air on the common path is the worse deal. Codex will not hard-fail (undici defaults to 300 s, `router.mjs:3253` sets `requestTimeout = 0`, and `upstream-retry.mjs` only retries pre-header transport failures), so this is a UX cost rather than a correctness one — but it is paid by every turn to buy a fix for a rare one.

**There is a hold-free design available, and I would rather have it.** Chat Completions permits `tool_calls` deltas after `content` deltas within one choice. Stream attempt 1 live; if it stops progress-only, run the retry and *append* its tool-call deltas plus `finish_reason: "tool_calls"` to the same open stream, discarding the retry's text. The client sees the status sentence followed by the tool call — the outcome you want — at zero added latency, and `finish_reason` is only written at the end so nothing is finalized prematurely.

Failing that, gate the hold on evidence: only hold a conversation that has already shown a progress-only stop on this `x-grok-conv-id`, so the dead air is paid only by sessions actually exhibiting the bug.

## Blocker 3 — a stray commit

`tracked.txt` arrives in commit `9cdb109` ("seed", author `t <t@example.com>`, content `original`). That is verbatim the fixture from `test/installer-scripts.test.mjs:376-393` — same filename, same content, same commit message, same `GIT_AUTHOR_NAME`. Running the suite leaked its throwaway git checkout onto your branch. Nothing references it. Please drop that commit.

## Should fix — the summed usage reaches capacity evidence

`src/context-window-drift.mjs:39-44` excludes `estimatedInputTokens` and `emptyCompletionRetried` — its own comment says a merged pair "is not one measurement" — but has no `retries` check. Demonstrated:

```
event { status:200, model:"grok-oauth/grok-4.6", inputTokens:525000, retries:1 }
acceptedInputTokens → 525000
contextWindowDrift  → { declared:500000, observed:525000, ratio:1.05 }
describe → "grok-oauth/grok-4.6 declares 500000 but accepted 525000"
```

grok-4.6 declares `contextWindow: 500000`, so any real prompt above ~250k that trips a retry produces drift output telling the operator to raise a window that is already correct. It needs a large prompt, so it is not the common case — but it is precisely the class of sample that predicate exists to refuse.

`provider-usage.mjs:154` already does `!event.retries`; mirroring that one line into `acceptedInputTokens` is the fix. Please also thread `retries` through `observeSubagentOutcome` (`src/router.mjs:2924`), which currently builds its event without it — that is on us, not you, but your change is what makes it reachable.

## Nit

`mergeMappedUsage` returns the other operand unchanged when either attempt reported no usage, so a retry whose first attempt had no usage emits no `retries` marker at all. And the forwarder's `retries` silently overwrites a genuine `upstreamRetries: 2` in the router event.

## Verified working

For the record, since these took real effort: the tools gate correctly rejects injected hosted-search tools (they are merged later in `toResponsesRequest`, so they cannot satisfy `requestOffersClientTools`); the kill switch and threshold overrides work; the retry body is drained via `cancel()` with an `arrayBuffer()` fallback; `x-grok-conv-id` is unchanged because `requestUpstream` passes the original message list on both attempts; and the nudge lands in `input` rather than `instructions`, so the #265 prefix cache stays intact — asserted at `test/grok-oauth-forwarder.test.mjs:668`.

Suite on your head merged onto current main: 1758 pass, 0 fail. That is not reassurance here, though — it is green because the retry tests are all non-streaming.

duolahypercho:
Correcting my own "should fix" item above — I gave you the wrong remedy, so please don't implement it as written.

I suggested adding `if (event?.retries) return undefined;` to `acceptedInputTokens`. That would be wrong on main, and it would throw away valid data:

- The **empty-completion** retry is the one that genuinely sums both attempts, and it already sets `emptyCompletionRetried: true` alongside `retries` (`src/router.mjs:2848`). `acceptedInputTokens` already excludes that, so the summed case is handled today.
- A plain **transport** retry (`upstreamRetries` from `fetchWithRetry`) sets `retries` but does *not* sum anything — the failed attempt is discarded and only the successful response's usage is recorded. That is a perfectly good capacity sample, and excluding it would discard real evidence from any turn that merely had a network hiccup.

`provider-usage.mjs:154` excludes `retries` for a different reason — a retried turn has misleading *latency* — which does not transfer to token counts.

**The real issue is specific to this PR**, and the fix is a marker rather than a predicate change. Your retry sums both attempts and sets `retries: 1` **without** any equivalent of `emptyCompletionRetried`, so it is the one shape that slips past the existing guard. Set an analogous marker — `progressOnlyRetried: true` — on the usage event, and add it beside `emptyCompletionRetried` in `acceptedInputTokens`. That extends the pattern the codebase already uses, keeps transport-retry samples, and excludes exactly the doubled ones.

The `retries` threading through `observeSubagentOutcome` (`src/router.mjs:2924`) is still worth doing so the subagent path can see the marker, but the same reasoning applies there: key the exclusion on the summed-usage marker, not on `retries` alone.

The demonstration in my earlier comment still stands — a summed 525,000 against grok-4.6's declared 500,000 does produce drift output telling an operator to raise a correct window. Only the remedy was wrong.

Zzy-min:
Addressed the three blockers and the corrected usage-marker note.

- **No hold.** Attempt 1 streams live again, including the short-with-tools path that was dropping deltas. A progress-only stop retries once and appends only the retry's `tool_calls` deltas plus `finish_reason: \"tool_calls\"` on the same open stream; retry text is discarded. The first answer is kept when the retry also has no tools.
- **Streaming coverage** added for short no-retry, retry-preferred, and keep-first.
- **`tracked.txt` / `seed` commit dropped.**
- **`progress_only_retried: true`** is the summed-usage marker. `acceptedInputTokens` excludes that (not bare `retries`). `observeSubagentOutcome` now forwards the same marker. Forwarder `retries` is added to `upstreamRetries` instead of overwriting it.

Targeted suite: 86 pass, 0 fail.

duolahypercho:
Verified the three blockers are actually fixed, then found one more and pushed the fix to your branch as `f99d1f3`.

## Blockers: confirmed fixed

Re-measured rather than taken on trust, against your head merged onto current main:

- **Held deltas.** Gone with the hold. Streaming short-answer-with-tools now delivers `{"content":"Done."}`; retry-preferred emits content deltas → tool_call deltas → `finish_reason: "tool_calls"` in that order on one open stream, which is protocol-valid; keep-first preserves the first answer and discards the retry text. Your three new streaming tests cover all three.
- **Liveness.** The 3 s-thinking probe that previously read 3,070 ms now reads **25 ms**. `writeHead(200)` fires as soon as the upstream headers arrive. No TTFT cost, and `firstTokenMs` is honest again.
- **`tracked.txt`.** Dropped.
- **Usage marker.** `progress_only_retried` is the right shape — it excludes exactly the doubled samples and keeps transport-retry samples. `acceptedInputTokens`, `provider-usage`, `observeSubagentOutcome` and `recentUsageEvents` all agree.

Full suite on your head + current main: **1766 pass, 0 fail**.

## What I found and fixed

`isProgressOnlyStop` matches a *shape*, and no better signal exists in the turn — so a **finished** task also matches it. "Yes, that is correct." after 1,500 reasoning tokens is byte-for-byte a turn that stopped after "Next I will update the deck.". Since grok-4.6 reasons by default, the `output_tokens` floor is met on essentially every turn and the 120-character limit is what actually decides. Short final answers are common, so this fires on the common path, not in the tail.

A wasted round trip would be fine. The nudge made it worse than that:

```
"Continue the same task by calling tools now."
```

That gives the model no way to say the task is done. Probed against a mock backend — turn 1 answers `"Yes, that is correct."` (1,500 output / 1,490 reasoning), retry gets the nudge:

```
finish_reason = tool_calls
content       = "Yes, that is correct."
tool_calls    = [{"function":{"name":"exec_command","arguments":"{\"cmd\":\"ls\"}"}}]
```

`shouldPreferRetryTurn` grafts a call the model never proposed onto a finished answer, and Codex runs it. The same happens to a hosted-search-only turn — `web_search_call` is not in `TOOL_ITEM_TYPES`, so a searched-and-answered turn is retried too (probe: `attempts=2`).

**The fix is the nudge wording, not new control flow** — offering the no-tool branch first routes the finished case into the keep-first path you already built and tested:

```js
const PROGRESS_ONLY_NUDGE =
  "If your previous message already completed the task, restate the final answer " +
  "and call no tool. Otherwise continue the same task now by calling the tools you need.";
```

A false positive now costs one round trip instead of a wrong action. Added `a finished task answered in one line declines the retry instead of calling a tool`, which asserts `finish_reason: "stop"`, the first answer intact, no `tool_calls`, and both attempts still billed with the marker set.

Also corrected the CHANGELOG and TROUBLESHOOTING wording: both said "a genuine progress-only stop", which claims a diagnosis the code cannot make. They now state plainly which turns pay for the retry and which knobs narrow it.

After the fix: **1763 pass, 0 fail**, `npm run check` clean.

## Nit, not blocking

`mergeTokenUsage` propagates `retries` but drops `progressOnlyRetried`. Harmless today — both merge sites are inside the empty-completion retry, which sets its own marker, and `provider-usage` still excludes on `retries`. Worth a line if that path ever grows a third caller.

Merging once CI is green.

PR #298, opened 2026-08-17
Fix DeepSeek reasoning replay in OpenCode Go

jepgambardella:
## Summary

Preserve DeepSeek V4 reasoning content when Codex history is replayed through OpenCode Go.

The router now:
- marks OpenCode Go DeepSeek V4 Flash and Pro as thinking models;
- keeps existing reasoning_content values;
- sends an empty reasoning_content field when the history has no value;
- carries the field through split assistant tool-call messages.

This is model-profile based and does not change other providers.

## Validation

- node --test test/routing.test.mjs (60 passed)
- codex doctor
- OpenCode Go routing test added

### DIFF
--- a/config/opencode/go/deepseek-v4-flash.json
+++ b/config/opencode/go/deepseek-v4-flash.json
@@ -33,8 +33,8 @@
       "searchTool": {
         "mode": "standalone"
       },
-      "requestProfile": "auto-tool-choice",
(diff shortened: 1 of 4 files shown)


jepgambardella:
Closing as duplicate of #275, which already fixes the generic reasoning replay path. #293 adds the regression test for the same OpenCode Go error. No separate OpenCode-specific fix is needed.

PR #308, opened 2026-08-18
feat(tray): support custom menu bar icons and icon-only mode (#307)

### LINKED ISSUE #307: Support Custom Menu Bar Icons and Icon-Only Mode
Codex Router currently displays the model name in the menu bar, which takes up a significant amount of space. It would be helpful to add menu bar display settings that allow users to show only an icon and select or upload a custom icon.

The current display mode should remain available, along with the following options:

- Show the icon only
- Use a custom menu bar icon
- Show or hide the model name

This would reduce menu bar clutter and allow users to customize the appearance to better match their desktop setup.

Zzy-min:
Resolves #307

Configurable menu bar layout and icons in `ModelRouterTray`.

- **Display mode:** standard (icon + name + usage) or icon-only
- **Show model name:** toggle in standard mode
- **Icon style:** activity dot by default (including existing installs); optional provider mark, preset SF Symbol, or custom image
- Settings persist under `ModelRouterTray.*` keys
- Custom images are copied into Application Support (5MB cap) and cached; missing files are shown in Settings
- Tooltip is localized; `.provider` reuses `ProviderIcon`
- Status item width is reserved: 180pt standard, 24pt icon-only

Self-review on the first push (defaults, ProviderIcon, reserved widths, persist/cache, resolver tests, tooltip) is folded into the feature commit. Inline threads are resolved.

### Commits

1. `feat(tray): support custom menu bar icons and icon-only mode`
2. `fix(test): stabilize dsh-web start/stop identity` — unrelated CI flake, kept separate
3. `chore: empty commit to restart CI` — Ubuntu runner hung on apt-get

### Verification

- `node --test test/tray-install.test.mjs test/tray-rebuild.test.mjs test/tray-service-windows.test.mjs`
- `node --test test/dsh-web.test.mjs` (WSL)
- macOS CI `swift test` (MenuBarSettings suite)


apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
```diff
@@ -2864,52 +2956,172 @@ struct ProviderSetupState: Decodable, Identifiable, Equatable {
...
+    let resources = installedBundle ?? Bundle.module
+    let url = resources.url(
+      forResource: assetName,
+      withExtension: assetExtension,
+      subdirectory: "ProviderIcons"
+    ) ?? resources.url(forResource: assetName, withExtension: assetExtension)
+    return url.flatMap(NSImage.init(contentsOf:))
+  }
+
+  private var assetName: String? {
```
Zzy-min: **[bug]** Default icon style is `.provider` (`menuBarIconStyle = .provider` when the key is absent), but `MenuBarIconView.assetName` only maps openai/grok/kimi/deepseek/anthropic/commandcode/github-copilot/chutes. The same target already has a complete, documented `ProviderIcon` in `IslandOverlay.swift` (opencode*, kilo-free, zai-*, qwen-plan, ollama-cloud/local, clinepass, minimax-token-plan, meta) and `ModelRouterTrayApp.swift` already uses it for quota rows. `assetExtension` is also keyed off `providerID` instead of the resolved asset name, so even a later prefix map would pick `.png` for `opencode-go` / `kilo-free`. Consequence: after upgrade, GLM/Qwen/Ollama/OpenCode/Cline/MiniMax/Meta/Kilo/local users lose the old activity dot and get a 6pt circle; while `activityState != .idle`, `StatusItemLabel` then adds a second 5pt badge (`menuBarIconStyle != .indicator`), so the menu bar shows two dots.

**Suggestion:** Delete the duplicate loader and render `ProviderIcon(providerID:size:)` for `.provider`. If a fallback is still needed, use the same `cpu` symbol `ProviderIcon` already uses, and only show the side badge when the main view is not already a status circle.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
```diff
@@ -2864,52 +2956,172 @@ struct ProviderSetupState: Decodable, Identifiable, Equatable {
...
+            .truncationMode(.tail)
+        } else if let usage = store.selectedUsageText {
+          Text(usage)
+            .font(.system(size: 10, weight: .medium, design: .monospaced))
+            .foregroundStyle(routerMuted)
+            .lineLimit(1)
+            .truncationMode(.tail)
+        }
+      }
+      .frame(width: store.menuBarShowModelName ? Self.reservedWidth : nil, alignment: .leading)
```
Zzy-min: **[bug]** Standard mode still renders usage or concurrent-provider text after the model-name toggle, but reserved width is now applied only when `menuBarShowModelName` is true (`.frame(width: nil)` otherwise). The previous `reservedWidth = 180` existed specifically to keep the `NSStatusItem` anchor stable while that text changed. Hiding the name therefore makes the status item jump between `"45% left"`, localized remaining strings, and `"Grok + Claude"` as activity starts and stops.

**Suggestion:** Keep a stable reserved width whenever standard mode still shows trailing usage/activity text (or reserve a smaller but fixed width for the usage-only layout). Do not make the status-item width follow the live string.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
```diff
@@ -2864,52 +2956,172 @@ struct ProviderSetupState: Decodable, Identifiable, Equatable {
...
+        MenuBarIconView(store: store, size: 14)
+          .scaleEffect(pulsing ? 1.4 : 1)
+          .animation(.easeOut(duration: 0.45), value: pulsing)
+        if store.menuBarIconStyle != .indicator && store.activityState != .idle {
+          Circle()
+            .fill(store.activityState.tint)
+            .frame(width: 5, height: 5)
+        }
+      }
+      .frame(minWidth: 18)
```
Zzy-min: **[bug]** Icon-only mode uses `.frame(minWidth: 18)` with no max/fixed width. Idle is a 14pt icon; non-idle non-indicator styles add a 5pt badge and 4pt spacing (~23pt), and the icon also `scaleEffect`s to 1.4 during `attentionPulse`. The design asked for a compact reserved slot (~22–26pt / ~24pt), not a growing `minWidth`. The item will shift its neighbors when activity starts, and the pulse can overflow the 18pt slot.

**Suggestion:** Give icon-only a fixed compact width (around 24pt) that already includes the badge, and pulse inside that frame instead of changing the status-item’s measured size.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
```diff
@@ -5237,6 +5569,18 @@ private struct TrayView: View {
...
 
+  private func chooseCustomIconImage() {
+    let panel = NSOpenPanel()
+    panel.allowedContentTypes = [.png, .jpeg, .svg, .icns, .tiff]
+    panel.allowsMultipleSelection = false
+    panel.canChooseDirectories = false
+    panel.canChooseFiles = true
+    panel.prompt = routerLocalized("Select")
+    if panel.runModal() == .OK, let url = panel.url {
+      store.setMenuBarCustomIconPath(url.path)
```
Zzy-min: **[suggestion]** `chooseCustomIconImage()` persists `url.path` only. The bundle is ad-hoc signed and not sandboxed, so a still-present user-readable file will usually reload after restart, but a moved, renamed, iCloud-evicted, or deleted file fails silently: `customImage` returns nil and the menu bar falls back to `cpu` while Settings still shows `URL.lastPathComponent`. `customImage` also calls `NSImage(contentsOfFile:)` from a SwiftUI computed property, so every `RouterStore` publish (activity poll can fire every 1s) re-reads the file.

**Suggestion:** Copy the chosen image into Application Support (or keep a security-scoped bookmark plus a cached `NSImage` on `RouterStore`). If the file is gone, clear or flag the setting and show that in the custom-image row instead of the stale filename.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
```diff
@@ -308,6 +319,28 @@ final class RouterStore: ObservableObject {
...
+      menuBarShowModelName = defaults.bool(forKey: menuBarShowModelNameKey)
+    } else {
+      menuBarShowModelName = true
+    }
+    if let raw = defaults.string(forKey: menuBarIconStyleKey),
+      let style = TrayMenuBarIconStyle(rawValue: raw)
+    {
+      menuBarIconStyle = style
+    } else {
+      menuBarIconStyle = .provider
```
Zzy-min: **[suggestion]** Absent `ModelRouterTray.menuBarIconStyle` becomes `.provider`. Every existing install therefore upgrades from the pulsing activity dot to a provider mark without opting in. Issue #307 asked to keep the current layout available; the design’s “preserve default mode” language describes the old Icon/Dot + name + usage row. Combined with the incomplete map in Issue 1, this is a silent visual change for everyone and a functional regression for unmapped providers.

**Suggestion:** Default missing keys to `.indicator` so current installs keep the shipped look. Use `.provider` only after an explicit Settings choice (or a one-time migration if you truly want the new default for fresh installs only).

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
```diff
@@ -308,6 +319,28 @@ final class RouterStore: ObservableObject {
     } else {
       presenceMode = .always
     }
+
+    if let raw = defaults.string(forKey: menuBarDisplayModeKey),
```
Zzy-min: **[suggestion]** New defaults load/defaulting lives only in `RouterStore.init` / the five setters. There is no extracted pure resolver and no Swift test, unlike `resolveIslandMode` + `IslandModeTests`. The PR’s “43 passed” run is `test/tray-install.test.mjs`, `tray-rebuild.test.mjs`, and `tray-service-windows.test.mjs`; those suites do not instantiate `RouterStore` or render `StatusItemLabel`. `LocalizationTests` only checks table-key parity, which these new strings already satisfy.

**Suggestion:** Extract a `resolveMenuBarSettings(...)` helper (unknown raw values, missing keys, default icon style) and add Swift tests in `apps/macos/ModelRouterTray/Tests`, the same way island mode is tested.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
```diff
@@ -2864,52 +2956,172 @@ struct ProviderSetupState: Decodable, Identifiable, Equatable {
...
+        }
+      }
+    }
+  }
+
+  private var tooltipText: String {
+    let provider = store.hasConcurrentActivity ? store.activitySummaryLabel : store.selectedUsageProvider.displayName
+    let state = store.activityState.label
+    if let usage = store.selectedUsageText {
+      return "Codex Router · \(provider) (\(state)) · \(usage)"
```
Zzy-min: **[suggestion]** The new `.help(tooltipText)` string is hardcoded English (`"Codex Router · \(provider) (\(state)) · \(usage)"`). Settings labels were translated in all five tables, so icon-only mode — the case that most needs the tooltip — is the one that stays English.

**Suggestion:** Build the tooltip with `routerLocalized` / `routerFormat` and add the source string to every language table.

Zzy-min (commented):
These seven findings were addressed in a9b69f0 (menu-bar defaults, ProviderIcon reuse, reserved widths, custom-icon persist/cache, resolver tests, localized tooltip). A follow-up d15892e unblocks macOS CI from an unrelated dsh-web process-identity flake (`exec sleep` raced `ps comm`).

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
Zzy-min: Addressed in a9b69f0: `.provider` now renders `ProviderIcon(providerID:size:showsHelp: false)`, so the full Island map is used and the inner `.help` no longer covers the localized tooltip.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
Zzy-min: Addressed in a9b69f0: standard mode always uses `MenuBarLayoutMetrics.statusItemWidth` (180pt), including when the model-name toggle is off.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
Zzy-min: Addressed in a9b69f0: icon-only mode uses a fixed 24pt frame. Pulse and the activity badge stay inside that slot.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
Zzy-min: Addressed in a9b69f0: the picker copies into Application Support via `persistCustomMenuBarIcon` (stage then replace, 5MB cap). `RouterStore` caches the NSImage and a missing flag; Settings shows a missing-file label instead of a stale filename.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
Zzy-min: Addressed in a9b69f0: missing `ModelRouterTray.menuBarIconStyle` resolves to `.indicator`. `.provider` is only an explicit Settings choice.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
Zzy-min: Addressed in a9b69f0: `RouterStore.resolveMenuBarSettings` is a pure helper, with `MenuBarSettingsTests` plus Node wiring guards in `test/tray-rebuild.test.mjs`.

apps/macos/ModelRouterTray/Sources/ModelRouterTrayApp.swift
Zzy-min: Addressed in a9b69f0: tooltip is `RouterStore.menuBarTooltip` via `routerFormat`, with keys in EN/ZH/AR/HI/JA/KO. `ProviderIcon` help is disabled on the menu-bar mark so the localized string wins.

PR #309, opened 2026-08-18
feat: add multi-credential pools with configurable rotation strategies (inspired by Hermes)

skyzea1:
## Problem

A single API key per provider dies on `429` / `402` / `“usage limit reached”` mid-session (rate limit, monthly quota, balance exhausted). Today Codex Router only fails over to a *different provider* (`model-failover`). Pools keep the *same provider* alive by trying the next healthy key first — critical for providers with strict per-key rate limits or plan quotas (DeepSeek, OpenRouter, Grok API, Command Code, Z.AI, etc.).

Inspired by **Hermes Agent — Credential Pools** (https://hermes-agent.nousresearch.com/docs/user-guide/features/credential-pools). Hermes design is the inspiration — no Hermes code is copied, the implementation is router-native and follows the spirit + four strategies.

## Solution

**Per-provider credential pools** stored safely in the existing protected storage:

* **Pool state:** `STATE_DIR/credential-pools.json` (`0600` + `protectPrivateFile`/`icacls` on Windows) — versioned, metadata-only (no secrets inline).
* **Each key:** `STATE_DIR/<credential.file>.pool.<id>.secret` (`0600`), fingerprinted `sha256:16`. Env-var / Keychain secrets are read-only and never persisted. Duplicate fingerprints rejected.
* **Seeding:** first `credential-pool add` auto-seeds the existing single `*.secret` as `primary` so enabling pools never discards a working key. Empty pools are pruned.
* **Scope:** `openai-compatible` providers with `credential.file`. `keyless` (local Ollama), `anonymous` (`opencode-free`/`kilo-free`) and `per-model` (`custom`) are explicitly rejected. `variantOf` families (e.g. `commandcode` + `commandcode-messages`, `opencode-go*`) share one pool via canonical id. Native ChatGPT/Codex (`openai` native path) is never pooled.

### Rotation strategies (per-provider, configurable)

| Strategy | Behaviour |
|---|---|
| `fill_first` *(default)* | Use the first healthy key until exhausted, then move to the next. |
| `round_robin` | Cycle evenly through healthy keys (persisted index). |
| `least_used` | Always pick the key with the lowest `requestCount`. |
| `random` | Random among healthy keys. |

Stored as `pools[provider].strategy` in `credential-pools.json`. Healthy = `cooldownUntil` absent or expired. `requestCount` + `lastUsedAt` are incremented on `selectCredential()` for `least_used`/`round_robin`.

### Automatic rotation

Wired into `src/api-forwarder.mjs` (selection is cooldown-aware; after each upstream response the used key is marked) and `src/provider-selection.mjs` (pooled providers are reported as *configured* when the pool has any entry — preserving single-key `picker` compatibility):

| Upstream | Action | Cooldown |
|---|---|---|
| `429` burst | Retry **same** key once (transient); second `429` → rotate | 1h (honours `Retry-After` capped 6h) |
| `402` or `out_of_usage` (`insufficient_quota`, `exceeded your current quota`, `balance too low`, `purchase extra usage`, `quota exhausted`, `arrears`, and Chinese `余额不足`/`欠费`/`额度不足/已用完` — same `QUOTA_PATTERNS` as `error-translation.mjs`) | **Immediately rotate** (no retry) | 1h |
| `401`/`403` | Rotate | 5m |
| `entitlement` (`plan doesn't include API` / `Go doesn't include Provider API`) | **No rotate** (different key won't help) | — |
| `200` | Clear `hasRetried429` + cooldown, keep `requestCount` | — |
| **All pool keys exhausted** | Surfaces `503`/`429` — router `model-failover` (cross-provider) then activates | — |

Next request picks the next healthy key per strategy; the forwarder logs `hasRetried429`/cooldown events.

## How to enable / configure

```sh
# add a second key (hidden prompt if --api-key omitted)
./bin/credential-pool add deepseek --label backup --api-key sk-...

./bin/credential-pool add openrouter   # prompts hidden
./bin/credential-pool add grok-api --label secondary --api-key xai-...

# inspect
./bin/credential-pool list
./bin/credential-pool list deepseek
./bin/credential-pool list deepseek --json

# set / view strategy
./bin/credential-pool strategy deepseek round_robin
./bin/credential-pool strategy deepseek  # show current
# also: fill_first | least_used | random

# reset cooldowns after a top-up / new billing cycle
./bin/credential-pool reset deepseek
./bin/credential-pool reset --all

# remove
./bin/credential-pool remove deepseek 2
./bin/credential-pool remove deepseek a1b2c3d4
```

Single-key path unchanged: `./bin/model-router codex provider-key <provider> set` still works. `credential-pool list` without pool reports `single-credential mode` plus current strategy hint.

Full guide: [`docs/CREDENTIAL-POOLS.md`](docs/CREDENTIAL-POOLS.md)

## CLI / Management (spec compliance)

* `list [provider]` / `status` – show pools, health, `requestCount`, `cooldownUntil`, `hasRetried429`
* `add <provider> [--label L] [--api-key K]` – label defaults to `key-N`, fingerprint deduped
* `remove <provider> <index|id>` – 1-based index or id/label, deletes only `STATE_DIR` files
* `strategy <provider> [strategy]` – get/set, validated
* `reset <provider|--all>` – clears `cooldownUntil`/`hasRetried429`/`lastStatus`

All commands support `--json` and accept both `list <provider>` / `<provider> list` ordering.

## Safety & compatibility

* No breaking change: absent pool file = original `resolveProviderCredential()` path. Existing `Kimi OAuth` / `Grok OAuth` / `DeepSeek` / `Grok` / `Command Code` / `opencode` / local flows untouched.
* Credential isolation preserved: `api-forwarder` injects only the selected pool key as `Authorization`/`x-api-key`; pool metadata never leaks secrets.
* Protected files stay `0600` (`protectPrivateFile` + `chmodSync`/`icacls`), pool member files only under `STATE_DIR`.
* `provider-selection` pooled check is pool-first, then classic file check — hides pooled providers correctly from `enabled-providers.json` and `ensure-configured`.

## Testing

* `src/credential-pool.mjs` + `src/credential-pool-cli.mjs` pass `node --check`.
* New suite `test/credential-pool.test.mjs` (isolated `CODEX_HOME`/`STATE_DIR`) — strategies, Quota/429/401 rotation, `hasRetried429`, `reset`, `remove`, non-poolable rejections. Passes: `node --test test/credential-pool.test.mjs` (2/2).

## Limitations / Future

* Pool rotation today marks for *next* request (not bounded in-request retry loop across keys — next turn already rescues sessions; tight same-turn loop is planned).
* OAuth pooling (`kimi-oauth`/`grok-oauth`) is scaffolded (pool guard) but not yet routed — single OAuth session still via `kimi login` / `grok login`. Native ChatGPT account remains single.
* No `Retry-After` header surface in `doctor` yet; could expose per-key `retryAt`.
* Future: `--pool` flag for `provider-key set`, and tray UI pool section.

## References

* Hermes Credential Pools docs: https://hermes-agent.nousresearch.com/docs/user-guide/features/credential-pools
* Prior art in-router: `src/model-failover.mjs` (cross-provider), `src/error-translation.mjs` (`QUOTA_PATTERNS`/`upstreamFailureKind`), `src/provider-credentials.mjs` (file/Keychain), `src/rate-limit-headers.mjs`.

### DIFF
--- a/bin/credential-pool
+++ b/bin/credential-pool
@@ -1,0 +1,5 @@
+#!/bin/sh
+set -eu
+
+repo_dir=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)
+exec node "$repo_dir/src/credential-pool-cli.mjs" "$@"
--- a/docs/CREDENTIAL-POOLS.md
+++ b/docs/CREDENTIAL-POOLS.md
@@ -1,0 +1,116 @@
+# Credential Pools — multi-account / multi-key pools
+
+Codex Router now supports **multiple credentials (API keys) for the same provider** and automatically rotates among them when one is rate-limited or quota-exhausted — similar to [Hermes Agent — Credential Pools](https://hermes-agent.nousresearch.com/docs/user-guide/features/credential-pools).
+
+Inspired by Hermes, implemented router-native: no Hermes code is copied, the design follows the same four rotation strategies.
+
+## Why
+
+A single API key dies on `429` / `402` / “usage limit reached” mid-session. Pools keep the session alive by trying the next healthy key for the *same* provider before falling back to a *different* provider (`model-failover`). The native ChatGPT/Codex account is never pooled.
+
+## Quick start
+
+```sh
+# Add a second key for the same provider (deepseek, openrouter, grok-api, etc.)
+./bin/credential-pool add deepseek --label backup --api-key sk-...
+# or hidden prompt:
+./bin/credential-pool add openrouter
+
+# Inspect
+./bin/credential-pool list
+./bin/credential-pool list deepseek
+
+# Choose strategy (default: fill_first)
+./bin/credential-pool strategy deepseek round_robin
+./bin/credential-pool strategy deepseek least_used
+./bin/credential-pool strategy deepseek random
+./bin/credential-pool strategy deepseek fill_first
+
+# Manual reset after a quota refills / limit raised
+./bin/credential-pool reset deepseek
+./bin/credential-pool reset --all
+
+# Remove one key (by 1-based index or id)
+./bin/credential-pool remove deepseek 2
+./bin/credential-pool remove deepseek a1b2c3d4
+```
+
+`bin/provider-key <provider> set` (single-key path) continues to work. The first `credential-pool add` auto-seeds the existing `STATE_DIR/<provider>*.secret` as `primary`, so enabling pools never discards a working key.
+
+## Rotation strategies
+
+Set per-provider via CLI or by editing `STATE_DIR/credential-pools.json` (`strategy` field).
+
+| Strategy | Behaviour |
+|---|---|
+| `fill_first` *(default)* | Use the first healthy key until exhausted, then move to the next. |
+| `round_robin` | Cycle evenly through healthy keys. |
+| `least_used` | Always pick the key with the lowest `requestCount`. |
+| `random` | Random among healthy keys. |
+
+`--json` on every command emits machine-readable output.
+
+## Automatic rotation
+
+`src/api-forwarder.mjs` selects a healthy key per strategy (cooldown-aware) and, after each upstream response, marks the used key:
+
+| Upstream | Action | Cooldown |
+|---|---|---|
+| `429` (generic/burst) | Retry same key **once** (transient); second `429` → rotate | 1h (honours `Retry-After` capped 6h) |
+| `402` or `out_of_usage` (`insufficient_quota`, `usage limit reached`, balance/arrears — including Chinese `余额不足/欠费/额度`) | **Immediately rotate** (no retry — cap won't clear) | 1h |
+| `401/403` | Rotate | 5m |
+| `entitlement` (“plan doesn't include API”, Go vs Provider) | **No rotate** (no key on same provider fixes it) | — |
+| `200` | Clear `hasRetried429` + cooldown, increment `requestCount` | — |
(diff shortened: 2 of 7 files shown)


duolahypercho:
Thanks for tackling credential rotation—the use case is real, but this implementation is not safe to merge as submitted.

The review found several correctness and security blockers:

- The CLI and documentation support `--api-key KEY`, which places provider secrets in process arguments. Router policy requires hidden interactive input so keys do not enter argv, logs, shell history, or tracked snippets.
- Rotation is only applied to the next request. The forwarder relays the failing 401/402/403/429 response after marking cooldown, so the PR does not provide the same-turn recovery described in its automatic-rotation behavior.
- Pool metadata can name an arbitrary credential `file`; reads are not constrained back beneath the router state directory. A damaged or tampered metadata file could therefore cause an unrelated local file to be used as an outbound credential.
- Pool state, counters, cooldowns, and member secrets are mutated without the repository's cross-process transaction/lock discipline, so concurrent CLI and forwarder writers can lose state.
- Error classification uses an unbounded `upstream.clone().text()`, allowing an upstream to force the forwarder to buffer an arbitrarily large error body.
- The two broad tests do not exercise the real forwarder boundary, concurrent mutations, path confinement, or same-request behavior.

A credential pool needs a bounded in-request retry design, hidden-only secret entry, state-directory confinement, atomic cross-process mutation, bounded body inspection, and end-to-end forwarder tests before it can protect core routing behavior. Closing this version rather than merging a partial credential-management path.

PR #312, opened 2026-08-18
feat: add generic OpenAI-compatible custom endpoint support

Recoba86:
## Summary

Adds support for registering generic OpenAI-compatible custom endpoints and gateways while decoupling transport identity from model compatibility requirements.

## Motivation

Codex Router already has provider-specific compatibility logic for built-in models, but models routed through a generic OpenAI-compatible gateway were previously identified primarily by the transport provider.

That becomes problematic when a single gateway exposes heterogeneous downstream model families. A Gemini model behind a custom gateway still needs Gemini request normalization, while Claude, DeepSeek, Grok, and other models may need different compatibility behavior even though they share the same transport.

Without separating transport identity from model compatibility, routed models can bypass the normalization applied to their native counterparts. This can surface provider-specific request failures, including strict schema validation errors when Codex-internal tool metadata such as `encrypted` reaches an upstream API unchanged.

This PR separates the custom endpoint transport from the model compatibility profile, allowing routed models to reuse the appropriate native compatibility behavior while keeping endpoint configuration, authentication, and routing generic.

### Key Changes
- **Compatibility Profiles**: Resolves model-level compatibility (`compatibilityFamily` / `compatibilityProfile`) so heterogeneous models behind the same custom gateway can reuse the appropriate provider-specific request behavior.
- **Tool-Schema Sanitizer**: Recursively strips client-internal `encrypted` parameter schema metadata before outbound dispatch to avoid strict upstream schema validation errors.
- **Model Metadata Defaults**: Applies an 85% `autoCompact` threshold for user-curated models that declare `contextWindow` but omit `autoCompact`, while preserving explicit overrides.
- **Documentation & Tests**: Adds generic custom endpoint documentation and regression coverage for compatibility resolution and recursive schema sanitization.

### Backward Compatibility

Existing built-in provider behavior remains backward-compatible and regression-tested.

## Test Plan

- `node --test test/custom-openai-endpoints.test.mjs test/tool-schema-sanitizer.test.mjs` (8/8 pass)
- `npm test` (1,771 passed, 0 failed, 12 skipped)
- `git diff --check`

### DIFF
--- a/docs/CUSTOM-OPENAI-ENDPOINTS.md
+++ b/docs/CUSTOM-OPENAI-ENDPOINTS.md
@@ -1,0 +1,141 @@
+# Custom OpenAI-Compatible Endpoints & Model Integration
+
+This document outlines the architecture, protocol translation, tool schema normalization, model overrides, and context window management for generic OpenAI-compatible endpoints in `codex-router`.
+
+---
+
+## 1. Architecture Overview
+
+`codex-router` routes OpenAI-compatible endpoints directly to downstream model APIs with full support for streaming, function calling, tool execution, and context compaction.
+
+```
+Codex Desktop / CLI Client
+           │
+           ▼ (HTTP POST /v1/responses or /v1/chat/completions)
+┌────────────────────────────────────────┐
+│             codex-router               │
+│  - Caller Authentication               │
+│  - Model Catalog Resolution            │
+│  - Namespace Flattening                │
+└──────────────────┬─────────────────────┘
+                   │
+                   ▼ (Loopback / Gateway Dispatch)
+┌────────────────────────────────────────┐
+│             api-forwarder              │
+│  - Downstream Provider Protocol Match  │
+│  - Compatibility Profile Resolution    │
+│  - Tool Schema Sanitization            │
+│  - Thought Signature Injection         │
+│  - Non-User Image Sanitization         │
+│  - Outbound Key / Header Management    │
+└──────────────────┬─────────────────────┘
+                   │
+                   ▼ (HTTP POST upstream)
+   Target Provider (e.g. Custom Upstream Gateway)
+```
+
+### Credential Protection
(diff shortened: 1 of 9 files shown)


duolahypercho:
Thanks for the proposal. The goal—explicit compatibility behavior for heterogeneous OpenAI-compatible endpoints—is useful, but the submitted implementation is not safe to merge.

The review found these blockers:

- The documented arbitrary custom-provider configuration is not actually registered or loaded by the runtime. The existing `custom` provider is a per-model endpoint container; this PR adds no complete provider-registration path, so the advertised feature and runtime behavior diverge.
- Compatibility profiles are inferred from provider/model name prefixes. Those heuristics can assign destructive behavior to unrelated endpoints (for example treating broad DeepSeek, Kimi, or Qwen names as specific request profiles) instead of requiring verified, explicit model capabilities.
- The recursive sanitizer removes every property named `encrypted`, including a legitimate user tool-schema field, rather than only known router metadata at a defined boundary.
- An 85% `autoCompact` policy is imposed on every user model with a context window without evidence that each endpoint supports that threshold.
- The new profile aliases duplicate the repository's explicit `requestProfile` architecture and are not validated against its closed enum.

This needs to be redesigned around the existing explicit per-model request-profile/capability model, with a real endpoint registration path and schema sanitation scoped to known metadata. Closing this version to avoid shipping behavior inferred from names or silently altering user schemas.

Recoba86:
Thanks for the detailed and constructive review!

The feedback makes total sense—especially relying on explicit `requestProfile` definitions instead of prefix heuristics and keeping sanitization strictly scoped to router metadata.

I'm planning to work on a redesigned v2 based on your guidance. That said, since you have the best architectural overview of the codebase, if you have the bandwidth or prefer to implement this directly upstream yourself, please feel free to do so as well! Either way, appreciate the direction.

PR #318, opened 2026-08-19
fix: preserve configured proxy in stable fetch transport

yishu-ziyu:
## Problem

The router service sets `NODE_USE_ENV_PROXY` together with `HTTP_PROXY` / `HTTPS_PROXY`, but `installStableFetchTransport()` immediately replaces Node’s environment-aware global dispatcher with a plain `Agent`.

That makes native Responses traffic bypass the explicitly configured outbound proxy. Long-lived streams such as remote compaction can then fail mid-response even though the router service was configured to use the proxy.

## Fix

- install `EnvHttpProxyAgent` when an HTTP(S) proxy environment variable is present
- retain the existing `allowH2: false` transport policy
- retain the direct `Agent` path when no proxy is configured
- add deterministic coverage for both branches

## Verification

- `node --test test/fetch-transport.test.mjs test/router-resilience.test.mjs` — 8/8 passed
- `npm run check` — passed
- `git diff --check` — passed

## Scope

This does not add Responses WebSocket forwarding. The router still rejects WebSocket upgrades and relies on Codex HTTP fallback; WebSocket support should be addressed separately because it changes the protocol boundary rather than the dispatcher selection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+- **Stable fetch transport preserves an explicitly configured environment
+  proxy.** `installStableFetchTransport` used to replace Node's
+  `NODE_USE_ENV_PROXY` dispatcher with a plain HTTP/1.1 agent, causing native
+  Responses traffic to bypass `HTTP_PROXY` / `HTTPS_PROXY`. It now installs an
+  `EnvHttpProxyAgent` when either proxy variable is set while retaining the
+  existing HTTP/2-disable policy.
+
 - **A Codex Stop hook continues grok-oauth mid-task stops once.** After a
   tool result, a short status sentence with no follow-up tool call used to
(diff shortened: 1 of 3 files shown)


duolahypercho:
The transport bug is valid, but this implementation is superseded by the hardened proxy work now integrated in #326 (from the corrected #323 path).

The original change selected `EnvHttpProxyAgent` merely because inherited `HTTP_PROXY`/`HTTPS_PROXY` existed, bypassing Node's explicit proxy opt-in contract. It also did not cover persisted service opt-in, Windows launcher credential protection, or the required loopback `NO_PROXY` guidance.

The replacement preserves `allowH2: false` and lowercase proxy precedence while requiring explicit opt-in, propagating that opt-in through service children, protecting credential-bearing Windows launchers, and testing service rendering and transport selection. Closing this duplicate in favor of that implementation.

PR #331, opened 2026-08-20
Preserve Codex default model during router updates

jepgambardella:
## Summary

Keep the user's Codex `model` and `model_reasoning_effort` when the router updates `config.toml`.

This stops installs and catalog refreshes from changing the default model. Explicit `login-free-enable <model>` behavior is unchanged.

## Validation

- `node --test test/config-manager.test.mjs`
- `npm run check`

### DIFF
--- a/src/config-manager.mjs
+++ b/src/config-manager.mjs
@@ -513,6 +513,20 @@ function replaceRootValue(contents, key, value) {
     .trimEnd();
 }
 
+// Enabling the router owns only its integration settings.  The selected Codex
+// model and reasoning effort belong to the user and must survive catalog
+// refreshes and reinstalls, even if a future managed-block rewrite changes
+// the surrounding root document.
+function preserveCodexDefaults(before, after) {
+  const { rootLines } = splitRoot(before);
+  let preserved = after;
+  for (const key of ["model", "model_reasoning_effort"]) {
+    const value = rootValue(rootLines, key);
+    if (value !== undefined) preserved = replaceRootValue(preserved, key, value);
+  }
+  return preserved;
+}
+
 function providerTableRanges(contents, providerId) {
   const { lines, headers } = scanTomlDocument(contents);
   const starts = headers.filter(({ path: header }) =>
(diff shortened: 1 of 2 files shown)


duolahypercho:
Closing this one, and I want to show the reasoning rather than just assert it, because the conclusion is unintuitive.

## The change is a no-op on every path it protects

I ran **this PR's own test against unmodified `main`** — it passes:

```
✔ config manager preserves Codex defaults and profiles (617ms)
```

`enabledContents()` never touches root `model` or `model_reasoning_effort`, and `clean()` only strips the router's own marker lines. Nothing in `src/config-manager.mjs` writes `model_reasoning_effort` at all. `main` already asserts both invariants: `test/config-manager.test.mjs:144` pins `model_reasoning_effort = "xhigh"` surviving enable, and `:161` pins `model = "gpt-5.6-sol"` surviving disable.

So `AGENTS.md:1307` ("preserve `model` and `model_provider` outside login-free mode") is a rule `main` already satisfies.

## What it does change is a regression

`preserveCodexDefaults` runs at the very bottom, **after** the `enable` branch has already called `applyRouterDefault(next)`. It re-reads `model` from `current` and overwrites whatever the router default installed. Reproduced in a scratch worktree:

```
# unmodified main
after router-default-set:  model = "deepseek/deepseek-v4-flash"
after enable:              model = "deepseek/deepseek-v4-flash"   <- re-asserted

# main + this PR
after router-default-set:  model = "deepseek/deepseek-v4-flash"
after enable:              model = "gpt-5.6-terra"                <- silently dropped
```

`codex-default-model.json` still records the default as managed, `control status` still reports `router_default_model`, and the Control Center still shows it active — while `config.toml` no longer contains it. That is a silent state desync, and it disables the sanctioned mechanism the control plane uses to migrate users off a retired model slug.

`main`'s own test at `test/config-manager.test.mjs:418` does not catch this, because there `current.model` already equals the router default, so the overwrite is a no-op. CI stays green while the feature breaks.

Two smaller things: `signed-enable` is included in the guard and also ends with `applyRouterDefault`, so it has the same clobber; and `replaceRootValue` returns a `.trimEnd()`-ed string that every other caller re-terminates with `\n` — this path does not, so every `enable` strips the config file's trailing newline for anyone with a root `model`.

## If there is a real scenario here

I may be missing the case that motivated this. `AGENTS.md:1735` — *"The default model is the user's. Restore only over a default this router wrote"* — is the principle, and `restoreRouterDefault` implements it with ownership tracking; the blunt "always restore whatever was in the file" is a coarser rule that overrides it.

If you have a concrete reproduction where `main` loses a user's model, please reopen with a test that **fails on `main` first**. That would be genuinely useful and I would want to fix it. Absent that, this trades a working feature for a no-op.

Thanks for looking at it — the config-preservation area is worth the scrutiny.

PR #333, opened 2026-08-21
fix: hide leaked MiniMax thinking output

moryk87:
## Summary

Prevent MiniMax M3 internal <think>...</think> reasoning from leaking into assistant-visible Responses text while keeping adaptive reasoning enabled.

The change is scoped to minimax-token-plan/minimax-m3.

## Root cause

MiniMax can return reasoning markup in the visible text channel. Restarting Codex does not restart the separately managed io.github.codex-router launchd service, so the installed process could continue running the old router code.

## Fix

- Keep MiniMax M3 adaptive thinking enabled and request split reasoning.
- Add a MiniMax-scoped Responses stream transform that removes <think>...</think> blocks, including tags split across SSE chunks.
- Preserve ordinary assistant text, tool calls, and follow-up routing.
- Add focused regression coverage and routing assertions.

## Verification

- node --test test/minimax-responses-compat.test.mjs — 2 passed
- node --test test/routing.test.mjs — 74 passed
- git diff --check
- Live io.github.codex-router restart: PID 47497, /health ok=true, degraded=[]
- Live MiniMax smoke request: HTTP 200, marker received
- Direct Responses visibility check: hasThink=false
- New MiniMax log entries after restart: status=200

### DIFF
--- a/src/api-forwarder.mjs
+++ b/src/api-forwarder.mjs
@@ -767,9 +767,13 @@ function normalizeBody(buffer, contentType, route) {
     }
   } else if (model.requestProfile === "minimax-m3") {
     // MiniMax uses its own thinking control on the OpenAI-compatible
-    // Chat Completions endpoint instead of reasoning_effort.
+    // Chat Completions endpoint instead of reasoning_effort. Without
+    // reasoning_split, MiniMax embeds the chain of thought in `content` as
+    // literal <think>...</think> markup, which makes it visible as ordinary
+    // assistant text in clients that do not recognize the vendor format.
     delete payload.reasoning_effort;
     payload.thinking = { type: "adaptive" };
+    payload.reasoning_split = true;
   } else if (model.requestProfile === "auto-tool-choice") {
     // Some models call tools happily under "auto" but reject being forced to,
     // the way DeepSeek and Qwen do in thinking mode. Their vendor profiles
--- a/src/minimax-responses-compat.mjs
+++ b/src/minimax-responses-compat.mjs
@@ -1,0 +1,124 @@
+import { Transform } from "node:stream";
+import { StringDecoder } from "node:string_decoder";
+
+function isMiniMaxModel(model) {
+  return [model?.provider, model?.upstreamModel, model?.gatewayModel, model?.slug]
+    .some((value) => typeof value === "string" && value.toLowerCase().includes("minimax"));
+}
+
+function stripThinking(text) {
+  return typeof text === "string"
+    ? text.replace(/<think\b[^>]*>[\s\S]*?<\/think\s*>/gi, "")
+    : text;
+}
+
+function parseSseBlock(block) {
+  const lines = block.split(/\r?\n/);
+  const index = lines.findIndex((line) => line.startsWith("data:"));
+  if (index === -1) return undefined;
+  const data = lines[index].slice(5).trim();
+  if (!data || data === "[DONE]") return undefined;
+  try {
+    return { lines, index, event: JSON.parse(data) };
+  } catch {
+    return undefined;
+  }
+}
+
+function rewrite(block, parsed, event) {
+  const lines = [...parsed.lines];
+  lines[parsed.index] = `data: ${JSON.stringify(event)}`;
+  return lines.join(block.includes("\r\n") ? "\r\n" : "\n");
+}
+
+function rewriteTextFields(value, sanitize = stripThinking) {
+  if (!value || typeof value !== "object") return value;
+  if (Array.isArray(value)) return value.map((entry) => rewriteTextFields(entry, sanitize));
+  const copy = { ...value };
+  for (const key of ["text", "delta"]) {
+    if (typeof copy[key] === "string") copy[key] = sanitize(copy[key]);
+  }
+  if (copy.item) copy.item = rewriteTextFields(copy.item, sanitize);
+  if (copy.part) copy.part = rewriteTextFields(copy.part, sanitize);
+  if (copy.output) copy.output = rewriteTextFields(copy.output, sanitize);
+  if (copy.content) copy.content = rewriteTextFields(copy.content, sanitize);
+  return copy;
+}
+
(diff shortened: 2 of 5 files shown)


duolahypercho:
Reviewed this while clearing the PR queue. It's still a draft so no action is expected — but the findings are concrete enough to be worth passing on now rather than after you've built more on top.

The problem is real and worth fixing. The implementation has four independent ways to silently delete legitimate output, and silent deletion is the hardest failure to attribute back to its cause.

**1. The filter runs on every field of every event, including tool-call arguments.** `rewriteTextFields` (line 34) walks blindly into `item`, `part`, `output`, `content` with no `event.type` dispatch. That includes `response.function_call_arguments.delta`, which streams tool-call arguments as incremental JSON. Concrete break: MiniMax calls `apply_patch` on a file whose content mentions `<think>` — a prompt template, an SSE parser, documentation about this very feature. The filter eats from `<think` onward, the reassembled arguments are invalid JSON, and the tool call fails.

`src/zai-responses-compat.mjs:230` is the established pattern here: it dispatches on `event.type` explicitly. This should do the same, restricted to `response.output_text.delta` / `.done`.

**2. `this.#buffer.indexOf("\n\n")` (line 68) never matches CRLF-delimited SSE.** `"\r\n\r\n".indexOf("\n\n") === -1`, so against a CRLF upstream no block is ever emitted until `_flush`: the whole response buffers in memory and arrives in one burst at end-of-stream. Streaming is destroyed and memory is unbounded. `zai-responses-compat.mjs:89-98` handles both separators, and `rewrite()` at line 30 of your own file already checks for `\r\n` — the splitter just didn't get the same treatment.

**3. An unclosed `<think>` swallows the rest of the response.** `#consume` sets `#insideThink = true` and discards everything until `</think>`. If the model opens one in legitimate visible content and never closes it, the entire remainder is dropped — and `_flush` doesn't emit `#pending` either, so a stream ending on a partial tag loses that text too. Note your non-streaming `stripThinking` uses a non-greedy regex that *requires* a close tag and is therefore safe; the streaming machine has no equivalent guard.

**4. `response.completed` is not sanitized.** `rewriteTextFields` recurses into `item`/`part`/`output`/`content` but not `response`. The terminal `{type:"response.completed", response:{output:[…]}}` carries the full unfiltered text, and Codex builds session history from it — so the thinking still leaks into history, and history diverges from what was displayed.

There's also an interaction worth knowing about: in `router.mjs` the transform list puts `EmptyCompletionGuard` **downstream** of this filter. A turn whose visible text is entirely consumed by issue 3 reaches the guard as an empty completion and triggers a retry — a second billed turn on content that wasn't actually empty.

**Separately — `reasoning_split`.** `src/api-forwarder.mjs:792` sends it on every request with `requestProfile: "minimax-m3"`. The scoping is tight (only `config/minimax/token-plan/minimax-m3.json` sets that profile), but I couldn't find documentation or a fixture establishing that it's a real MiniMax parameter. If their OpenAI-compatible endpoint rejects unknown body fields, this 400s every Token Plan turn. The added test asserts the router *sends* it against a mock, which proves nothing about acceptance. Worth a live probe before this lands — and if it works, it fixes the problem at the source and the transform can shrink to a narrow belt-and-braces on `output_text` only.

**On the tests:** the "removes leaked think markup" case splits the payload across two SSE events but keeps `<think>` and `</think>` each intact within one chunk — so it doesn't cover the split-marker path, which is the `#pending` path and the source of issues 3 and 4. Nothing covers `_flush` with pending state, CRLF, a missing close tag, function-call arguments, or `response.completed`.

The state machine is a reasonable sketch and the intent is right. I'd rewrite it against `zai-responses-compat.mjs` — type-dispatched, dual-separator, per-item state — rather than patch these individually. Happy to review again whenever you take it off draft.

duolahypercho:
Fixed and shipped in #373 — closing this one, with the CHANGELOG crediting the report to you.

**Your diagnosis was right on the part that mattered:** identifying `reasoning_split` as MiniMax's own control for this. I ran a controlled probe against `api.minimax.io` with the same prompt, changing only that field:

| | HTTP | `<think>` in `content` | separate reasoning field |
|---|---|---|---|
| `thinking: {type:"adaptive"}` (main) | 200 | **yes** | none |
| `+ reasoning_split: true` | 200 | **no** | `reasoning_content`, 351 chars |

Without it, `content` starts `<think>The user wants me to calculate 17 * 23…`. With it, `content` carries only the answer. And since this router already relays `reasoning_content` as reasoning, nothing downstream needed changing — so it shipped as a one-line fix.

**What #373 left out is the 163-line stream transform**, and I want to explain that rather than just drop it, because the reasoning generalises.

Once the source fix is confirmed, a downstream filter is pure risk. Four specifics, all reproducible from the code as written:

1. `rewriteTextFields` walks `item`/`part`/`output`/`content` with no `event.type` dispatch, so it also rewrites `response.function_call_arguments.delta`. A model writing a file that merely *mentions* `<think>` — a prompt template, an SSE parser, docs about this feature — gets its tool arguments truncated into invalid JSON.
2. `indexOf("\n\n")` never matches CRLF, so against a CRLF upstream nothing emits until flush: the whole response buffers and lands in one burst.
3. An unclosed `<think>` discards the rest of the response, and `_flush` drops pending text too.
4. `response.completed` isn't sanitized, so the markup still reaches session history — which then disagrees with what was displayed.

Plus a downstream interaction: `EmptyCompletionGuard` runs *after* the filter in `router.mjs`, so a turn emptied by (3) triggers a retry — a second billed turn on content that wasn't empty.

The underlying bug was cosmetic; a filter that eats tool calls is not. That asymmetry is what made "fix it at the source" the right call rather than "repair the four blockers."

If you want to keep a filter as belt-and-braces later, `src/zai-responses-compat.mjs` is the pattern to build on — type-dispatched, dual-separator, per-item state.

Thanks for the report and for doing the live verification in the PR body; the `HTTP 200` line is what let me skip straight to isolating the variable.

PR #335, opened 2026-08-21
fix(opencode-go): correct GLM-5.3 context and enable standalone search

mirucon:
## Summary

- advertise the OpenCode Go GLM-5.3 route with its current 1M context window
- compact at 900K, matching the repository's other million-token routes
- enable Codex standalone tool discovery for this model
- bump the capability hash and add registry regression coverage

## Context sizing

The entry was originally added at 200K as a conservative fallback because the contemporaneous Z.ai documentation exposed 1M only through a `[1m]` suffix. OpenCode's current model metadata now maps `opencode-go/glm-5.3` to `zhipuai/glm-5.3`, whose published limit is 1,000,000 tokens. This change uses the exact provider-published 1M value rather than GLM-5.2's older 1,048,576 convention.

## Standalone tool discovery

On a tool-heavy Codex installation, a fresh trivial GLM-5.3 request reached the old declared 200K ceiling immediately because Codex eagerly included the direct tool/schema surface.

Local A/B through the installed router:

```text
before: estimatedInputTokens = 200000 (clamped)
after:  estimatedInputTokens = 48241
```

A deferred GitHub-tool workflow then stayed bounded:

```text
47362 -> 53550 -> 56096
```

The model successfully discovered and invoked the GitHub connector and returned the requested issue title in a live testing.

OpenCode reported `inputTokens = 0` on these turns, so the figures above are router estimates. The pre-change 200K value was clamped to the declared window, not an exact measurement.

## Verification

- live fresh-session A/B through OpenCode Go
- live deferred GitHub connector discovery and invocation
- registry regression coverage for:
  - `contextWindow = 1_000_000`
  - `autoCompact = 900_000`
  - `searchTool.mode = "standalone"`
- CI run #854 passed:
  - `npm run check`, `npm test`, and audit on Linux, macOS, and Windows
  - desktop checks and binary builds on Linux and Windows
  - Electron control-center checks, packaging, and platform smoke tests

OX-alpha and curated-model metadata are deliberately out of scope for this focused fix.

### DIFF
--- a/config/opencode/go/glm-5.3.json
+++ b/config/opencode/go/glm-5.3.json
@@ -21,12 +21,15 @@
           "description": "Maximum reasoning depth"
         }
       ],
-      "contextWindow": 200000,
-      "autoCompact": 180000,
+      "contextWindow": 1000000,
+      "autoCompact": 900000,
       "inputModalities": [
         "text"
       ],
-      "compHash": "opencode-go-glm-5-3-v1"
+      "searchTool": {
+        "mode": "standalone"
+      },
+      "compHash": "opencode-go-glm-5-3-v2"
     }
   ]
 }
--- a/test/registry.test.mjs
+++ b/test/registry.test.mjs
@@ -419,6 +419,7 @@ test("provider registry exposes configured API and OAuth model families", () => 
   const standaloneSearchSlugs = new Set([
     "deepseek/deepseek-v4-flash",
     "opencode-go/deepseek-v4-flash",
+    "opencode-go/glm-5.3",
     "xiaomi-mimo/mimo-v2.5",
     "zai-coding/glm-5.3",
   ]);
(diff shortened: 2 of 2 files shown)


duolahypercho:
Closing as superseded — **the context-window half landed in #359**, with your values unchanged.

You were right, and the evidence was stronger than the PR claimed. `908a096` established via a live probe (990,020 prompt tokens accepted) that 200,000 belongs to the **GLM-5.1** lineage, not 5.3, and corrected both Z.ai entries to 1,000,000/900,000. This opencode Go entry was simply missed, so sessions on it were compacting at 180K against a model serving a million.

The extra evidence: `glm-5.1` and `glm-5.2` on the *same* opencode Go gateway already declare 1,048,576. So the gateway demonstrably does not clamp the GLM family the way it clamps `grok-4.5` to 500K — which makes your 1,000,000 conservative for this relay rather than optimistic. That reasoning is now recorded in the entry's `description`.

## What I did not take: `"searchTool": { "mode": "standalone" }`

`docs/HOW-IT-WORKS.md` is explicit: *"Only enable it after verifying that the provider's model accepts Codex's web-search result items and preserves tool/function-call history."* Every existing opt-in in the tree carries that evidence — `zai-coding/glm-5.3` records its probe in the `description` field — and `908a096` specifically declined to claim `zai-api/glm-5.3` was probe-verified because no credential was available for that endpoint.

The opencode Go relay is a **different transport** from the Z.ai endpoint where the capability was proven, so the Z.ai probe does not carry over. The failure mode is also quiet rather than loud: if the relay drops Codex's web-search result items, the model answers as though it never searched. That is a wrong answer, not an error, which is the worst kind to ship on an unverified flag.

If you can run a web-search turn through opencode Go's GLM-5.3 and confirm the tool/function-call history survives, please open a follow-up with the `searchTool` block and note the probe — it will go straight in. `README.md:212` and `docs/HOW-IT-WORKS.md:193` list the enabled routes and should be updated at the same time; I corrected that list in #359, since it named two of the five actually in the registry.

Thanks — this was a real bug and the fix is shipped.

PR #341, opened 2026-08-21
feat(opencode): support OpenCode Zen Muse 1.2 Free and Ox Alpha Free

fromiron:
## Why

OpenCode Zen continues to add popular models to its free catalog. Supporting these routes in Codex allows users to evaluate current models through the anonymous provider without configuring credentials.

The two current Zen Free models use different wire contracts: Muse Spark 1.2 Contributor Free uses Responses, whereas Ox Alpha Free uses Chat Completions. Routing both through the existing Chat Completions route directs Muse to the wrong endpoint. Correct endpoint routing alone is also insufficient for file editing: Codex exposes `apply_patch` as a native custom tool, whereas these Zen adapters expect standard function tools.

## Changes

| Model | Documented Zen endpoint | Router provider |
| --- | --- | --- |
| `muse-spark-1.2-contributor-free` | `/zen/v1/responses` | `opencode-free-responses` |
| `x-preview-f-free` | `/zen/v1/chat/completions` | `opencode-free` |

- Add an internal Responses variant for Muse while retaining Chat Completions for Ox, following the [OpenCode Zen endpoint table](https://opencode.ai/docs/zen).
- Preserve shared OpenCode Free selection, picker visibility, deduplication, and user-edited metadata when explicit curation migrates Muse to the Responses route.
- Use published context windows when Zen's live model list omits them: 1,048,576 tokens for Muse and 1,000,000 for Ox, under the existing 85% auto-compaction policy.
- Bridge native custom `apply_patch` definitions, calls, results, history, compaction inputs, JSON responses, and streamed deltas exclusively for these two routes.
- Fix missing assistant-message SSE envelopes occurring when LiteLLM translates Ox Chat Completions to Responses.
- Propagate upstream HTTP errors and cancellations rather than masking failures behind fallbacks.

|Model|SS|
|------|----|
|Muse Spark 1.2 Contributor Free|image|
|Ox Alpha Free|image|

## Scope and side-effect controls

- `opencode-free-responses` is an internal `variantOf` route with an anonymous allowlist restricted to Muse Contributor Free.
- Only an explicit `curate-models opencode-free` run migrates existing Chat-routed Muse entries. Installation, discovery, updates, and registry loads do not rewrite user curation.
- OpenCode Go, paid Zen, all other OpenCode Free models, credentials, provider selection, usage accounting, and cooldown scopes retain their existing contracts.
- The Go Muse setting `supportsApplyPatchTool: false` remains intentionally unchanged; this PR includes no Go write/readback verification and does not infer Zen Free behavior across the separate billed Go service.
- Custom-tool conversion is collision-safe and does not reinterpret malformed model arguments as valid native custom-tool payloads.

## Verification ledger ??scoped PASS

Verified at head `1de9bfa4948886f2e4094a45a09630025e6dca41` on Windows with Node 26 and Codex CLI `0.149.0-alpha.4`.

### Automated verification

- `npm run check`: passed.
- Curation, discovery, picker, provider-selection, registry, and user-model tests: 94 passed, 1 POSIX-only test skipped on Windows.
- Namespace relay and schema tests: 72/72 passed.
- Full `test/routing.test.mjs`: 77/77 passed.
- `git diff --check` and `git show --check`: clean.

### Clean-install verification

- Uninstalled the prior managed service and skill pack, verified removal of router ports and managed config markers, and backed up all previous active state.
- Installed from this exact head with only `opencode-free` selected; confirmed that no previous keys, models, usage ledgers, or `OPENCODE_GO_BASE_URL` settings were transferred.
- Verified that generated LiteLLM config routes Muse through `openai/responses/...` without `use_chat_completions_api`, and routes Ox through `openai/...` with `use_chat_completions_api: true`.
- `model-router codex doctor`: passed all required checks, including catalog parity, service health, routing config, caller/internal keys, and five managed skills. Warnings were limited to credential and curation notices for unselected providers.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -64,7 +64,9 @@ user.
    interactive terminal to choose models. If they did not specify and
    credentials already exist, use
    `configured` rather than showing providers that cannot authenticate.
-   The anonymous providers `opencode-free` and `kilo-free` are also selectable,
+   The anonymous providers `opencode-free` and `kilo-free` are also selectable
+   (`opencode-free-responses` is an internal, single-model protocol variant of
+   the former and is never selected or curated separately),
    but they need no credential only for their documented free model subsets.
    Both are catalog-only: they ship no preselected models and need
    `bin/curate-models PROVIDER` after selection. `custom` is selectable on the
--- a/README.md
+++ b/README.md
@@ -399,7 +399,13 @@ in code to its official endpoint.
 
 Both are catalog-only and deliberately ship no checked-in model metadata: the
 provider's live `/models` response is filtered to the free subset and then added
-locally with `./bin/curate-models`.
+locally with `./bin/curate-models`. OpenCode Free curation routes
+`muse-spark-1.2-contributor-free` through its internal Responses sibling while
+keeping Ox Alpha Free (`x-preview-f-free`) and the other free IDs on Chat
+Completions; the provider remains one selection in setup and the picker. An
+existing Chat-routed copy of that one Muse model is migrated only when the
+operator explicitly runs `curate-models`; install, update, and catalog reads do
+not rewrite the user model or picker state.
 
 ```sh
 ./bin/model-router codex providers enable opencode-free
--- a/config/opencode-free/opencode-free.json
+++ b/config/opencode-free/opencode-free.json
@@ -10,6 +10,21 @@
       "authMode": "anonymous",
       "anonymousModelPolicy": "opencode-console",
       "anonymousNote": "No API key is needed for OpenCode Zen free models. Use at your own risk: access is a published exception, not an entitlement, and the free catalog and limits can change without notice."
+    },
+    {
+      "id": "opencode-free-responses",
+      "displayName": "OpenCode Free Responses",
+      "kind": "openai-compatible",
+      "ownedBy": "opencode",
+      "variantOf": "opencode-free",
+      "protocol": "openai-responses",
+      "baseUrl": "https://opencode.ai/zen/v1",
+      "authMode": "anonymous",
+      "anonymousModelPolicy": "explicit-models",
+      "anonymousModels": [
+        "muse-spark-1.2-contributor-free"
+      ],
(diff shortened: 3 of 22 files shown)


duolahypercho:
Closing as superseded — **your work is in `main`**, not discarded.

This PR and #357 had byte-identical diffs against `main` (22 files, +2745/−55). #357 was the rebase of this branch, and it went in via #363 along with two review fixes:

- **The `apply_patch` bridge was dropping Codex's lark grammar.** `bridgeCustomTools` replaced the custom tool with a one-string function forwarding only `description` — but Codex's native `apply_patch` ships the V4A grammar and *no* description, so the model received no specification of the patch format at all. The grammar is now folded into the bridged description. Worth noting the test fixture hid this: its grammar was `"start: /.+/"` and it supplied a description, so nothing was lost in the test. It now uses the real 17-line grammar with no description.
- **`migrateModelVisibility`'s destination check** was `seeded.has(to)` only, so a hand-edited state file could have an explicit "show" silently overwritten.

One thing your PR got right that survived review intact, and that I want to call out: the gating discipline. `needsZenFreeToolCompatibility` matching on exact `(provider, upstreamModel)` pairs, every hot-path branch sitting behind it, the response-side bridge unreachable without a per-request `WeakMap` entry only that gate writes — plus a genuine five-provider negative control asserting the custom tool, the recursive `$ref`, `search_content_types`, `agent_message` and `detail:"original"` all survive untouched on other routes. That is a higher standard than most changes to shared request-path code arrive with, and it is why the review found no cross-provider regression risk to fix.

The context windows were also re-checked rather than taken on trust, and your figures held up: models.dev keys `limit` per *free id* and demonstrably throttles them (`deepseek-v4-flash-free` is 200,000 against the paid 1,000,000), so `muse-spark-1.2-contributor-free` matching its paid twin at 1,048,576 is affirmative evidence the route is not throttled. Both values were kept, with the sourcing now recorded on the entries.

Thanks for the contribution.

PR #343, opened 2026-08-21
api-forwarder: keep streamed usage flowing from OpenAI-compatible providers

DavidAGInnovation:
## Problem

Running a routed model through the Responses bridge against OpenCode's free
Zen models (e.g. `opencode-free/x-preview-f-free`), Codex showed
"Context window: 100% full (125k / 125k tokens)" after a single "hello!"
turn.

Root cause chain, observed on 0.4.0-beta.4 (`9b2b88a`) with Codex CLI/App
0.149.0:

1. OpenCode's chat-completions stream sometimes emits the usage chunk
   **after** the terminating `data: [DONE]` line — reproducibly when the
   request carries an empty `tools: []` array alongside
   `tool_choice`/`parallel_tool_calls`, but also intermittently without.
   Codex always sends empty `tools`, so routed turns hit this constantly.
2. LiteLLM's bridge stops at `[DONE]` (correctly), so the completed event
   reaches the client with `input_tokens: 0`.
3. `ResponseUsageTransform` then substitutes its byte-based estimate, which
   for real app payloads is large enough to hit the clamp at
   `contextWindow` — and the client renders a full window.

## Fix

Two normalizations on the outbound `/chat/completions` body in
`normalizeBody`:

- set `stream_options.include_usage: true` on streamed calls;
- drop empty `tools` arrays, which trigger the post-`[DONE]` usage
  placement on OpenCode.

With this, streamed turns meter real upstream tokens again (verified
end-to-end through gateway -> LiteLLM -> forwarder ->
opencode.ai/zen/v1; doctor green).

## Related

- LiteLLM itself drops the post-`[DONE]` chunk by spec; when a provider
  places usage there nothing client-side can recover it. We carry a small
  venv patch that copies `chunk.usage` into the assembled model response in
  `LiteLLMCompletionStreamingIterator`
  (`patches/litellm-responses-stream-usage.patch`, included with a README)
  so operators have a recovery path; happy to rework how it's shipped if
  you'd prefer a different mechanism.
- Operators can also opt out of estimate substitution entirely with
  `CODEX_ROUTER_ZERO_INPUT_ESTIMATE=0`.

## Testing

- `node --check src/api-forwarder.mjs`
- End-to-end streamed request via the authenticated gateway:
  `response.completed.usage` now reports real upstream counts
  (`input_tokens: 9`) instead of `0`/clamped estimates; recorded
  `usage-events.jsonl` shows `inputTokens: 9, estimatedInputTokens: null`.
- `bin/model-router codex doctor`: 0 failures before and after.

### DIFF
--- a/patches/README.md
+++ b/patches/README.md
@@ -1,0 +1,22 @@
+# Local LiteLLM patches
+
+## litellm-responses-stream-usage.patch
+
+`LiteLLMCompletionStreamingIterator` (the Responses -> Chat Completions
+bridge) never copies the upstream chat-stream usage chunk into the final
+`response.completed` event, so clients receive `input_tokens: 0` whenever
+the provider reports usage only on that chunk.
+
+The patch captures `chunk.usage` in both the sync and async iteration
+loops and applies it to the assembled model response before the
+completed event is emitted.
+
+Apply after dependency installation:
+
+```sh
+patch -d .venv/lib/python3.12/site-packages \
+  -p3 < patches/litellm-responses-stream-usage.patch
+```
+
+Note: `bin/install --force-deps` and `model-router codex doctor --fix`
+rebuild the venv, which reverts this patch; re-apply it afterwards.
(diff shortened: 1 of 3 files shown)


duolahypercho:
Superseded by #346, which preserves the original streamed-usage intent while keeping the fix inside the router: it injects `stream_options.include_usage` only for streamed Chat Completions, preserves existing stream options/tools, leaves non-stream and other API surfaces byte-equivalent, adds zero-usage regression coverage, and passed the full local and cross-platform CI suites. Closing this version because its manual LiteLLM site-packages patch would mutate installed dependencies outside the repository.

PR #351, opened 2026-08-22
fix: support DeepSeek Harness credential refs

jepgambardella:
## What changed

Support the current DeepSeek Harness credential file format with a top-level `refs` map. Keep compatibility with older flat credential files.

## Why

DeepSeek Harness writes `version` and `refs` around its credentials. The picker update changed router state, then catalog refresh failed because Codex Router rejected this valid structure.

## Validation

- `node --test test/dsh-config-manager.test.mjs test/target-integration.test.mjs`
- Live `dsh-config-manager install`: 14 models published
- Live `dsh-config-manager status`: credential detected
- Live picker update completed without error

### DIFF
--- a/src/dsh-config-manager.mjs
+++ b/src/dsh-config-manager.mjs
@@ -87,6 +87,12 @@ function normalizeTrailing(lines) {
   return copy;
 }
 
+function credentialPath(document, reference) {
+  return document.root.children.has("refs")
+    ? ["refs", reference]
+    : [reference];
+}
+
 /**
  * Splices the router's route into the settings document text.
  *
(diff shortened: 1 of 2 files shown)


duolahypercho:
Leaving this open rather than closing it, because **the gap you identified is real** — current harness builds do wrap credentials in a `version`/`refs` envelope and `applyCredential` only understands the flat root map. But it can't merge as written, and I'd rather hand you the specifics than close it.

I replayed the PR's `applyCredential`/`removeCredential` against the unmodified `src/yaml-structure.mjs` and parsed the output. Three reproduced problems:

**1. Hardcoded 2-space indent emits YAML no parser accepts.** `const indent = nestedRefs ? " ".repeat(refs.indent + 2) : ""` uses the column of the `refs:` key, not the indentation the existing children actually use. Given a document indented with 4 spaces:

```yaml
version: 1
refs:
    OTHER: existing
```

the result is:

```yaml
version: 1
refs:
    OTHER: existing
  CODEX_ROUTER_CALLER_KEY: "sek"
```

→ `ParserError: expected <block end>, but found '<block mapping start>'`. The user's `.credentials.yaml` becomes unreadable and the harness loses **every** credential in it, not just ours.

The module already solves this correctly 65 lines above, in `applyRouteToSettings`, with a comment spelling out why: *"Follow whatever indentation the document already uses for a sibling route rather than assuming two spaces."* It derives the indent from `[...providers.children.values()][0].indent`. Do the same, falling back to `refs.indent + 2` only when `refs` is empty.

**2. A first install against a current harness writes to the wrong place.** `credentialPath` keys off `document.root.children.has("refs")`. Verified:

- `"version: 1\n"` → `version: 1\nCODEX_ROUTER_CALLER_KEY: "sek"\n` (root level)
- `""` → `CODEX_ROUTER_CALLER_KEY: "sek"\n` (root level)

That is exactly the state before any ref exists, so the *first* install writes the legacy flat shape this PR exists to move away from. The harness resolves `apiKeyEnv` under `refs`, finds nothing, and the route 401s with no diagnostic — while `status()` reports `credentialPresent: true`, because it uses the same presence heuristic and finds the root-level key. Detect the envelope by `version`, or adopt the modern shape by default.

**3. `removeCredential` no longer restores the document exactly.** Applying then removing our ref when it is the only entry yields `version: 1\nrefs:\n`, which parses as `{'version': 1, 'refs': None}` — not the original, and semantically distinct from an absent key. `AGENTS.md:266` requires that removal restore the document exactly. `removeRouteFromSettings` already handles this, walking up and pruning emptied ancestors, with a comment noting *"An empty mapping there is not the same as an absent one."*

Also worth tightening: the guard change to `if (node.children.size && node.key !== "refs")` weakens the nested-document refusal to one level and never inspects `refs`' own children, so a foreign document like `refs:\n  server:\n    host: example.com` is now edited where `main` refused it — the docstring this PR rewrites still claims it refuses those.

## The security review came back clean

No path traversal (the ref never builds a filesystem path; `DSH_CREDENTIAL_REF` is a module constant), no secret in any thrown message or log, `status()` returns a boolean rather than the value, and file permissions are unchanged at 0600 under 0700. The one soft spot is problem 2, which fails silently at request time as a 401 rather than loudly.

## The test

The single added test asserts the right three properties — exact rendered output, idempotency, and the removal round-trip — and would fail on `main`. It covers one shape: 2-space indent, `refs` present, a sibling surviving removal. Every failure above is a shape it does not exercise: 4-space indent, `refs` absent with `version` present, empty file, empty `refs`, foreign nested doc, removing the last ref.

`AGENTS.md:269` is the governing rule here — *"Refuse rather than guess… A refusal costs a command; a wrong guess rewrites a file whose only copy is on the user's disk."* The corrections are all small, and the correct patterns are 60 lines away in `applyRouteToSettings`/`removeRouteFromSettings`. Happy to review a revision.

duolahypercho:
Superseded by #371, which closes the gap you found. Closing this one — but the report was correct and the fix is shipping because of it, so the CHANGELOG entry credits `#351 by @jepgambardella`.

#371 keeps your diagnosis intact (current harness builds use the `version`/`refs` envelope; `applyCredential` only understood the flat root map) and rewrites the write path around the four issues in the review above:

- indent derived from an existing sibling rather than assumed, so a 4-space envelope no longer produces a document YAML cannot parse
- envelope detected by `version`/emptiness rather than by `refs` presence, so a **first** install against a current harness writes into `refs` instead of silently landing at root level
- `status()` resolves through the same path as the writer, so present-but-unreadable is no longer representable
- emptied `refs` pruned on removal, and the `refs`-children leaf check restored

One thing worth passing on, since it is the kind of detail that is easy to miss and expensive to get wrong: the correct patterns were already in the same file, about 60 lines from the code being changed. `applyRouteToSettings` derives its indent from `[...providers.children.values()][0].indent` with a comment spelling out why assuming two spaces is wrong, and `removeRouteFromSettings` walks up pruning emptied ancestors with a comment noting that an empty mapping is not the same as an absent one. Both of those existed because of earlier bugs in this exact area.

Thanks for finding it — the envelope gap was real and would have kept 401ing quietly.

PR #355, opened 2026-08-22
fix(installer): install Node dependencies before guided credential setup

mayrazing:
### Why

On a fresh install, `node_modules` does not exist yet.

However, the installer tries to configure the DeepSeek API key using code that depends on `proper-lockfile`. This causes an `ERR_MODULE_NOT_FOUND` error.

In short, the installer runs the code before installing its dependencies.

### What

Fix the installation order:

- Install the Node dependencies after the user confirms they want to enter an API key.
- Save the API key after the dependencies are installed.
- Do not install dependencies if the user declines.
- Add regression tests to prevent this issue from happening again.

### Tests

All 51 related tests passed, with 2 skipped on the current platform.

The following checks also passed:

```text
npm run check
sh -n install.sh
git diff --check
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -274,6 +274,9 @@ command -v node >/dev/null 2>&1 ||
 command -v npm >/dev/null 2>&1 ||
   die "npm is required and is normally included with Node.js"
 
+if [ -n "$configure_provider_keys" ]; then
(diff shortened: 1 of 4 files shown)


mayrazing:
Closing in favour of #362 and #364

PR #356, opened 2026-08-22
fix(setup): add model selection to guided setup

mayrazing:
## Why

Guided setup currently lets users select providers, but not individual models.

The project requires models to be explicitly selected before they appear in the Codex model picker. As a result, setup can finish successfully and save the API key, while the new models remain hidden.

## What

- Add a model selection step after provider selection.
- Show only models from the selected providers.
- Publish only the models selected by the user.
- Preserve model visibility settings for other providers.
- Show the selected models on the installation review screen.

For example, selecting DeepSeek lists the available DeepSeek models for the user to choose from.

## Test

- Verified that selecting DeepSeek only lists DeepSeek models.
- Verified that selecting only Flash shows Flash and hides Pro.
- Verified that changing DeepSeek models does not affect other providers.
- Relevant tests: 123 passed, 1 skipped.
- `npm run check` passed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -274,6 +274,9 @@ command -v node >/dev/null 2>&1 ||
 command -v npm >/dev/null 2>&1 ||
   die "npm is required and is normally included with Node.js"
 
+if [ -n "$configure_provider_keys" ]; then
+  node "$repo_dir/src/node-dependency-install.mjs"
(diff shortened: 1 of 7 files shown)


mayrazing:
Closing in favor of #364

PR #374, opened 2026-08-23
refactor: harden Windows private-file ACL handling (scope b)

Zzy-min:
## Summary

Scope **(b)** from the split of the earlier bundled PR #349: the `src/file-security.mjs` rewrite.

## What changes

- Removes `currentWindowsSid()` and `windowsFullControlGrant()` (and the exported `windowsFullControlGrant` they exposed) that shelled out to `icacls.exe` and carried the numeric-SID 1332 pitfall.
- `protectPrivateFile()` now performs the ACL rewrite inline in a single PowerShell call: it disables ACL inheritance, drops every existing rule, and grants only the current Windows identity `FullControl`. It no longer depends on `icacls` and its SID prefix handling.
- `privateFileIsProtected()` is stricter and truly verifies the invariant it names: the ACL must be protected, have no inherited rule, grant the current SID full control, and contain no foreign `Allow` grant.

## Why

The old code produced an ACL, but the check did not prove "only the current identity has full control" — foreign `Allow` rules and inherited rules were not rejected, and the `icacls` SID prefix logic was a correctness and portability hazard. The rewrite makes pose and verify arrive at the same definition.

## Verification

- `test/file-security.test.mjs` rewritten to assert the removal of foreign grants, the exact single `FullControl` rule set, inherited-rule rejection, and the strict re-protection path. Runs on Windows (ACL exercise) and passes on POSIX (mode 0600 path).
- Regression run on `test/file-security.test.mjs`, `test/caller-auth.test.mjs`, `test/config-manager.test.mjs`: **40 pass / 1 skip / 0 fail** — the stricter checker does not regress existing private-file handling.

## Scope boundary

This is strictly the `file-security.mjs` change (scope b). It does **not** include the Antigravity provider (scope a, PR #372), the Windows tray/installer work (scope c: separate PR), or the catalog picker ordering (scope d: separate PR). No credentials, no third-party secrets, no unrelated hunks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+- **Private state files are hardened with a canonical owner-only ACL on
+  Windows.** The previous Windows path asked `GetAccessControl` about the
+  file's existing DACL and then edited it with `SetAccessRuleProtection` and
+  `RemoveAccessRuleSpecific`, so a file whose DACL was already non-canonical
+  could make it throw (or silently keep foreign ACEs), the exact drift an
+  install or `doctor --fix` is meant to repair. Windows hardening now builds a
+  fresh, empty `FileSecurity`, replaces the DACL outright with a single
+  current-identity FullControl Allow rule and inheritance cleared, and skips
+  persisting owner/group so it needs only `WRITE_DAC` and cannot fail where an
+  `icacls /inheritance:r /grant:r` path would have succeeded. A hardening
+  failure now exits non-zero with the PowerShell diagnosis on stderr instead of
+  masquerading as success.
+
 - **MiniMax M3 no longer shows its chain of thought as assistant text.** The
   Token Plan route asked for adaptive thinking but not `reasoning_split`, so
   MiniMax embedded the reasoning in `content` as literal `<think>...</think>`
--- a/src/file-security.mjs
+++ b/src/file-security.mjs
@@ -10,38 +10,31 @@ import {
 } from "node:fs";
 import path from "node:path";
 
-let windowsSid;
-
-function currentWindowsSid() {
-  if (windowsSid) return windowsSid;
-  const script =
-    "[Console]::Out.Write([Security.Principal.WindowsIdentity]::GetCurrent().User.Value)";
-  windowsSid = execFileSync(
-    "powershell.exe",
-    ["-NoLogo", "-NoProfile", "-NonInteractive", "-Command", script],
-    { encoding: "utf8", stdio: ["ignore", "pipe", "ignore"] },
-  ).trim();
-  if (!windowsSid) throw new Error("Could not resolve the current Windows user SID.");
-  return windowsSid;
-}
-
-export function windowsFullControlGrant(sid) {
-  if (!/^S-\d+(?:-\d+)+$/i.test(sid)) {
-    throw new Error("Could not format an invalid Windows user SID.");
-  }
-  // icacls requires an asterisk before a numeric SID so it is not resolved as
-  // an account name. Without it, hosted runners fail with system error 1332.
-  return `*${sid}:(F)`;
-}
-
 export function protectPrivateFile(target) {
   chmodSync(target, 0o600);
   if (process.platform !== "win32") return target;
-  const sid = currentWindowsSid();
+  const script = [
+    "$target = $env:CODEX_ROUTER_PRIVATE_FILE",
+    "$sid = [Security.Principal.WindowsIdentity]::GetCurrent().User",
+    "$acl = [System.IO.File]::GetAccessControl($target)",
+    "$acl.SetAccessRuleProtection($true, $false)",
+    "$rules = @($acl.GetAccessRules($true, $true, [Security.Principal.SecurityIdentifier]))",
+    "foreach ($rule in $rules) { [void]$acl.RemoveAccessRuleSpecific($rule) }",
+    "$fullControl = [System.Security.AccessControl.FileSystemRights]::FullControl",
+    "$none = [System.Security.AccessControl.InheritanceFlags]::None",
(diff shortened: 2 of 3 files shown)


duolahypercho:
Thanks for splitting this out — reviewable on its own, this is a much better shape.

**The security core is genuinely stronger than `main`, and stronger than the version in #349.** Building the DACL from a fresh `[FileSecurity]::new()` rather than mutating `GetAccessControl` means the pre-existing DACL is replaced wholesale, so no foreign ACE can survive — `main`'s `/inheritance:r` left explicit ACEs in place, and #349's `RemoveAccessRuleSpecific` could silently fail over a non-canonical DACL. Reading ACEs as raw SIDs via `GetAccessRules($true,$true,[SecurityIdentifier])` also removes a real fragility: `main`'s NTAccount translation throws `IdentityNotMappedException` for an orphaned SID or an unreachable DC, making a correctly-protected file report unprotected. And the stricter `privateFileIsProtected` closes a false-positive hole `main` had.

Path handling is injection-clean: the path never enters the command string, it crosses via `CODEX_ROUTER_PRIVATE_FILES` as JSON and is parsed by `ConvertFrom-Json`, so spaces, quotes, `&`, `%` and non-ASCII are safe by construction.

I also confirmed the deleted `windowsFullControlGrant()` has **zero importers** anywhere, and that removing the numeric-SID regression test is correct *here* — it guarded the `*SID:(F)` prefix that `icacls` needed to avoid "system error 1332", and with `icacls` gone that bug class cannot recur. That is a real difference from #349, where the same deletion accompanied a weaker replacement.

Three things to fix.

**1. This is now slower than `main` on every credential write, and the comment says the opposite.**

`main` memoizes the SID — `if (windowsSid) return windowsSid` — so it pays one `powershell.exe` cold start *per process* and then runs `icacls.exe` per call. This version spawns `powershell.exe` on **every** call with no memoization, and `powershell.exe -NoProfile` is roughly an order of magnitude more expensive to start than `icacls.exe`.

That matters on a real path: `src/local-download.mjs:305-322` calls `writeLocalDownload` on every distinct `(percent, detail)` tuple, so one model download makes hundreds of synchronous `writePrivateJson` calls — each now blocking the event loop on a PowerShell cold start, mid-stream. `src/service-windows.mjs:143-151` pays two spawns per launcher write.

The header comment frames this as *lowering* spawn count. It does relative to #349's two-per-write design, not relative to `main`. Either restore an `icacls` fast path, re-memoize what can be memoized, or state the tradeoff honestly in the comment. (Also worth noting `protectPrivateFilesWin32` is never called with more than one path anywhere in the tree, so the batching the comment advertises has no user yet.)

**2. A hardening failure can plausibly exit 0, and its diagnostics are discarded.**

`stdio: "ignore"`, and the script has no `$ErrorActionPreference = 'Stop'`, no `try/catch`, and no explicit `exit 1`. `execFileSync` only throws on a non-zero exit, and whether PowerShell returns non-zero for an uncaught `SetAccessControl` exception depends on its method-invocation error semantics rather than on anything this script asserts. `icacls.exe` returned documented non-zero codes, so this is strictly less reliable than `main`.

The failure mode is the invisible one: the credential file is written, `writePrivateFile` returns success, and the DACL was never applied. Even when it *is* detected, stderr is thrown away so the error carries no message for a `doctor` report.

Suggest `$ErrorActionPreference='Stop'`, `try { … } catch { [Console]::Error.Write($_); exit 1 }`, and capturing stderr into the thrown error.

**3. `SetOwner`/`SetGroup` narrow the set of DACLs you can repair.**

Persisting the OWNER and GROUP sections requires `WRITE_OWNER`. Windows grants a file's owner implicit `READ_CONTROL | WRITE_DAC` but **not** `WRITE_OWNER`, whereas `icacls /inheritance:r /grant:r` needed only `WRITE_DAC`. So in exactly the scenario this rewrite exists to fix — a protected but non-canonical DACL that doesn't grant the current user FullControl — `SetAccessControl` can throw `UnauthorizedAccessException` where `main` would have succeeded. Since all sections persist in one call it is all-or-nothing, so the DACL doesn't get fixed either.

`SetGroup($sid)` in particular buys nothing: the POSIX group field isn't consulted by Windows access checks. Suggest dropping both, or persisting the DACL first and treating owner/group as best-effort.

## Tests

The two new tests are **real behavioral tests**, not source-regex change-detectors — they mutate an actual DACL, run the real `protectPrivateFile`, and `deepEqual` the resulting rules against exactly one expected ACE. The second one asserts the *before* state is unprotected-with-foreign-allow, so the repro can't silently degrade into a no-op. Good discipline.

One gap worth closing, and it is the important one: **nothing tests that `writePrivateFile` leaves the target protected on Windows.** This PR *removed* `protectPrivateFile(target)` from that path and rests entirely on rename carrying the DACL. The reasoning is right — `temporary` is always a same-directory sibling, so the move is same-volume — but if temps ever move to a system temp dir, the cross-volume copy would inherit the destination's DACL and hardening would vanish with nothing catching it. Three lines under the win32 skip would guard it.

## Verification

Full suite on the PR: 2193 tests, **0 failures**, skips 13 → 14. That shape is exactly right — one always-run test removed, one Windows-only test added — and the new skip is expected, since `ci.yml` runs `npm test` on `windows-latest` where it does execute.

Note I could not run Windows, so everything about `powershell.exe` behavior above is reasoned from source and Win32 semantics rather than executed. One thing to watch in CI specifically: the `foreach (…) {` line is split across `join("; ")` elements, producing `…) {;   $acl = …`. A bare `;` should be a legal empty statement, but this is the only place in the repo that splits a PowerShell block this way — and if it does not parse, every Windows credential write throws. `windows-latest` will tell us.

## Verdict: NEEDS-FIXES

Fix (2) and (3), add the `writePrivateFile` target test, and either restore an `icacls` fast path or make the comment honest about the cost — then this is a clear improvement worth merging. The `CHANGELOG.md` entry is also missing.

PR #375, opened 2026-08-23
feat: harden Windows tray lifecycle and add deploy helpers (scope c)

Zzy-min:
## Summary

Scope **(c)** from the split of the earlier bundled PR #349: the Windows tray + installer PowerShell work.

## What changes

**`src/tray-service-windows.mjs`** — the Task Scheduler tray lifecycle gets hard, bounded state handling:
- Every `schtasks`/COM call now runs under a per-command timeout (`TASK_COMMAND_TIMEOUT_MS`, 4 s), so a hung Task Scheduler cannot stall the wrapper forever.
- `taskState()` reads Task Scheduler's localized-independent `State` enum instead of parsing `schtasks` text, and falls back between `powershell.exe` and `pwsh.exe`.
- New `waitForTaskState()` turns `stop`/`start` into verified transitions (`TASK_STOP_TIMEOUT_MS`/`TASK_START_TIMEOUT_MS` polling at 250 ms) instead of issuing an unverified `/Run`.
- New `registeredTaskAction()` lets `status` report which executable the registered task actually points at, and whether that companion exists on disk.
- `installTask()` injects an explicit `FullControl` ACE for the task principal through Task Scheduler's COM API, so an elevated install does not leave the unelevated account unable to update its own tray task.
- `uninstall` now re-checks `taskExists()` and fails loudly if Task Scheduler rejected the deletion.

**`codex-router.ps1`** — adds `tray repair`:
- `Get-ValidatedTrayTask` refuses to touch a task whose principal is not the current user, is not interactive, or whose action is not this checkout's companion.
- `Test-TrayTaskFullControl` / `Repair-TrayTaskPermissions` run a narrowly scoped elevated child (`-Verb RunAs`) that validates the task identity from fixed env fields again (closing the UAC race) before rewriting its DACL, then re-installs the companion.

**`install.ps1`** — preserves an already opted-in tray across updates (best effort, never installing on fresh routers), and refreshes managed Codex skills as a best-effort post-health step. Antigravity-specific lines are deliberately absent (they belong to scope a).

**`deploy-codex-router.ps1`** (new) — working-tree deployment helper: robocopy without blanket purge, manifest-based pruning of only prior-deployed files (protects operator notes / unrelated files), then runs the canonical `install.ps1 -CheckoutInstall -Target codex` + `doctor`, and strictly verifies an opted-in tray.

**`restart-codex-router.ps1`** (new) — graceful `service.mjs restart` helper.

## Verification

- `test/tray-service-windows.test.mjs` (new, 17 tests) covers rendering, platform refusal, build gating, dispatcher routing, and the new bounded lifecycle source assertions.
- `test/windows-operations.test.mjs` (new, 6): Windows PowerShell parses the three `.ps1` scripts; asserts the supported service transaction, manifest-based deploy, tray-opt-in preservation, and skills refresh.
- `test/service-render.test.mjs`: unchanged behavior; the one failure (`launchd preserve proxy`) is a pre-existing baseline failure present on `main` too, unrelated to this change.

## Scope boundary

Strictly the Windows tray + installer PowerShell (scope c). Excluded: Antigravity provider (scope a, PR #372), `file-security.mjs` rewrite (scope b, PR #374), catalog picker ordering (scope d, separate PR).

### DIFF
--- a/codex-router.ps1
+++ b/codex-router.ps1
@@ -28,6 +28,168 @@ function Invoke-RouterNode([string]$Script, [string[]]$ScriptArguments = @()) {
   }
 }
 
+function Resolve-AccountSid([string]$Identity) {
+  try {
+    return ([Security.Principal.SecurityIdentifier]::new($Identity)).Value
+  } catch {
+    return ([Security.Principal.NTAccount]::new($Identity)).Translate(
+      [Security.Principal.SecurityIdentifier]
+    ).Value
+  }
+}
+
+function Get-ValidatedTrayTask {
+  $TaskName = "Codex Router Tray"
+  $Task = Get-ScheduledTask -TaskName $TaskName -ErrorAction Stop
+  $CurrentSid = [Security.Principal.WindowsIdentity]::GetCurrent().User.Value
+  $PrincipalSid = Resolve-AccountSid ([string]$Task.Principal.UserId)
+  if ($PrincipalSid -ne $CurrentSid) {
+    throw "Refusing to repair '$TaskName': its principal is not the current user."
+  }
+  if ($Task.Principal.LogonType.ToString() -ne "Interactive") {
+    throw "Refusing to repair '$TaskName': it is not an interactive user task."
+  }
+
+  $Actions = @($Task.Actions)
+  if ($Actions.Count -ne 1) {
+    throw "Refusing to repair '$TaskName': it does not have one recognized action."
+  }
+  $TaskAction = $Actions[0]
+  $Execute = [IO.Path]::GetFullPath(
+    [Environment]::ExpandEnvironmentVariables([string]$TaskAction.Execute)
+  )
+  $Argument = [string]$TaskAction.Arguments
+  $TauriExecute = [IO.Path]::GetFullPath(
+    (Join-Path $Root "apps\desktop\src-tauri\target\release\codex-router-desktop.exe")
+  )
+  $ElectronExecute = [IO.Path]::GetFullPath(
+    (Join-Path $Root "apps\electron\node_modules\electron\dist\electron.exe")
+  )
+  $ElectronDirectory = [IO.Path]::GetFullPath((Join-Path $Root "apps\electron"))
+  $TauriAction = [string]::Equals($Execute, $TauriExecute, [StringComparison]::OrdinalIgnoreCase) -and
+    [string]::IsNullOrWhiteSpace($Argument)
+  $ElectronAction = [string]::Equals($Execute, $ElectronExecute, [StringComparison]::OrdinalIgnoreCase) -and
+    [string]::Equals($Argument.Trim().Trim('"'), $ElectronDirectory, [StringComparison]::OrdinalIgnoreCase)
+  if (-not ($TauriAction -or $ElectronAction)) {
+    throw "Refusing to repair '$TaskName': its action is not this checkout's tray companion."
+  }
+
+  return [pscustomobject]@{
+    Name = $TaskName
+    Sid = $CurrentSid
+    Execute = [string]$TaskAction.Execute
+    Argument = $Argument
+  }
+}
+
+function Test-TrayTaskFullControl([string]$TaskName, [string]$SidValue) {
+  try {
+    $Service = New-Object -ComObject "Schedule.Service"
+    $Service.Connect()
+    $Registered = $Service.GetFolder("\").GetTask($TaskName)
+    $Descriptor = [Security.AccessControl.RawSecurityDescriptor]::new(
+      $Registered.GetSecurityDescriptor(7)
+    )
+    foreach ($Ace in $Descriptor.DiscretionaryAcl) {
+      if ($Ace -is [Security.AccessControl.CommonAce] -and
+          $Ace.AceQualifier -eq [Security.AccessControl.AceQualifier]::AccessAllowed -and
+          $Ace.SecurityIdentifier.Value -eq $SidValue -and
+          ($Ace.AccessMask -band 0x1f01ff) -eq 0x1f01ff) {
+        return $true
+      }
+    }
+  } catch {
+    # A descriptor we cannot inspect is exactly the case the elevated repair
+    # is allowed to address after validating the task's principal and action.
+  }
+  return $false
+}
+
+function Repair-TrayTaskPermissions {
+  $Validated = Get-ValidatedTrayTask
+  if (Test-TrayTaskFullControl $Validated.Name $Validated.Sid) {
+    Write-Output "Tray task permissions are already repairable by the current user."
+    return
+  }
+
+  # The elevated side reads only fixed environment fields, validates the task
+  # again to close the UAC race, and changes its DACL through Task Scheduler's
+  # supported COM API. The tray process itself remains Interactive/Limited.
+  $ElevatedScript = @'
+$ErrorActionPreference = "Stop"
+function Resolve-RepairSid([string]$Identity) {
+  try { return ([Security.Principal.SecurityIdentifier]::new($Identity)).Value }
+  catch { return ([Security.Principal.NTAccount]::new($Identity)).Translate([Security.Principal.SecurityIdentifier]).Value }
+}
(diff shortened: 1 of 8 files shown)


duolahypercho:
Thanks for splitting this out. Reviewed `897af21` in an isolated worktree; full suite is green on macOS (2233 tests, 0 failures merged into current main; skips 13 → 14, the one new skip being the Windows-gated parse test, which is expected).

**I have no Windows machine and no `pwsh`**, so everything about PowerShell behavior below is reasoned from source and Win32 semantics rather than executed. I have flagged the items that must be exercised on real Windows before this lands.

## The UAC path is well built — with one cheap must-fix

I came into this prepared to reject the elevation path and could not construct an attack against its design. The pre-elevation validation is substantive, not decorative: principal SID must equal the current identity, `LogonType` must be `Interactive`, exactly one action, and the action's executable must be byte-equal (after `ExpandEnvironmentVariables` + `GetFullPath`) to this checkout's Tauri or Electron companion. Crucially it is **re-validated on the elevated side** against the pre-elevation values, closing the TOCTOU window. The payload is a single-quoted here-string so no caller data is interpolated into script text, and the blast radius is one `AccessAllowed` ACE on a *Limited*, user-principal task — which does not yield elevation even in the worst case.

**The defect is the launcher itself, not the design.** `codex-router.ps1:175`:

```powershell
Start-Process -FilePath "powershell.exe" -Verb RunAs ...
```

`-Verb` forces `UseShellExecute`, so resolution goes through `ShellExecuteEx`, whose search order includes the **current working directory** and every `PATH` entry. The script does not `Push-Location` first, so the CWD is wherever the user's shell happened to be — plausibly `Downloads`, or an extracted archive. An unprivileged attacker who can drop a `powershell.exe` into the CWD or any user-writable `PATH` entry gets their binary launched **behind the UAC prompt, fully elevated**. That is the classic local-privilege-escalation shape, and I confirmed this is the **only** `-Verb RunAs` in the codebase — so it is the one place it matters.

One line: `Join-Path $env:SystemRoot 'System32\WindowsPowerShell\v1.0\powershell.exe'`, ideally with `-WorkingDirectory $env:SystemRoot`.

## Blockers

**1. `TASK_COMMAND_TIMEOUT_MS = 4_000` (2 s per host) contradicts this repo's own precedent for the identical call.** `src/service-windows.mjs:250` uses `TASK_STATE_TIMEOUT_MS = 15_000` for the same `Get-ScheduledTask … .State` read, and gives `Register-ScheduledTask` **no timeout at all**. A cold `powershell.exe -NoProfile` running a `ScheduledTasks` CDXML cmdlet is routinely 1.5–3 s on a healthy desktop and worse domain-joined or under install load.

The PR also inverts the failure philosophy at the same time: `service-windows.mjs:262-271` deliberately *returns* rather than throwing when no PowerShell can answer, with a comment explaining why; `tray-service-windows.mjs:244` throws. Combined, a single slow `Get-ScheduledTask` now aborts `install`, `start`, `restart` and `uninstall` where main proceeded.

**2. The 4 s cap on `installTask` manufactures the exact state `tray repair` exists to fix.** That one child does CLR start + module load + four CDXML cmdlets + `Register-ScheduledTask` RPC + COM connect + `GetSecurityDescriptor`/`SetSecurityDescriptor`. A timeout kills it *between* registration and the DACL fix-up, leaving a task without the full-control ACE. And `:108-116` reports every failure as an ownership problem, steering the user into a **UAC prompt they did not need** — where repair will likely say "already repairable" and loop them back.

**3. `deploy-codex-router.ps1:171` breaks on PowerShell 7.4+ on the success path.** `$ErrorActionPreference = "Stop"` at `:9`, then a bare `& robocopy`. From 7.4, `$PSNativeCommandUseErrorActionPreference` is on by default — and robocopy returns **1** on normal success ("one or more files copied"), so line 171 raises a terminating error and lines 172-173, the entire reason the script knows robocopy's 0-7 convention, are never reached. Needs `$PSNativeCommandUseErrorActionPreference = $false` around the call or a `try`/`catch`. (5.1 is unaffected.)

**4. Likely fatal, and I cannot verify it: environment variables probably do not cross the UAC boundary.** `:158-177` passes all four repair parameters via `$env:CODEX_ROUTER_TRAY_REPAIR_*` then elevates with `Start-Process -Verb RunAs`. That goes `ShellExecuteEx` → AppInfo service → `CreateProcessAsUser`, which builds the child's environment from the elevated token rather than inheriting the caller's process environment. If that holds, the elevated script binds `-TaskName $null`, fails validation, and `tray repair` has never worked. It **fails safe** — no wrong task gets rewritten — but the headline feature would be dead on arrival, and the only test covering it is a source regex that cannot detect this. Please exercise it on real Windows. If confirmed, pass the values inside the `-EncodedCommand` payload.

## Regression risks

- **`status` becomes a throwing command.** Demonstrated: on main `tray-service.mjs status` prints JSON and exits 0; on this PR it throws `"Task Scheduler did not answer whether the tray task is registered."` and exits 1. Callers that parse that JSON must now tolerate a non-zero exit.
- **`uninstall` can no longer complete when the scheduler is unreadable** — `endTask()` throws before `schtasks /Delete` is attempted.
- **`tray repair` triggers a full companion rebuild** (`:260` sets `$Action = "install"`) — a multi-minute Rust/Electron build for someone who asked only to fix permissions. Worth saying so in the output.
- **`tray repair` refuses when run from the wrong root** — it compares against `$PSScriptRoot` while the task was registered from `SOURCE_ROOT`, so a dev-checkout user whose task points at `%LOCALAPPDATA%` is told "not this checkout's tray companion". That is precisely the person reaching for repair.
- **`src/tray-service-windows.mjs` has no `skipServiceManagerCall` guard**, unlike `service-windows.mjs` which guards every mutation. Pre-existing, but this PR widens what a stray `npm test` on a Windows dev box would do — `installTask` now also rewrites the task DACL and `startTask` actually launches the companion. Given the earlier incident where the suite hijacked a live launchd registration, worth closing.

**`install.ps1` is fine** — both new blocks are `try`/`Write-Warning`, neither can enter the rollback path, and the tray refresh is gated on `$TrayWasInstalled` so a fresh install never gains a tray nobody asked for. That is *more* conservative than `bin/install`. Non-Windows users are unaffected; verified empirically.

**Command construction is clean** — every attacker-influenced value crosses via environment variables and is referenced as `$env:…` inside the script text, never interpolated. No injection or quoting bug found.

## Tests

Of the 15 new tests, **14 execute nothing** — they `readFileSync` a `.ps1`/`.mjs` and `assert.match` regexes against its text. Some assert on things with no behavioral content at all: `:88` pins a local variable name; `:113` pins an expression whose `"missing"` branch is unreachable; `:86` pins a parameter (`options.timeoutMs`) no caller passes; `:130` asserts on the text of a **source comment**.

I understand the constraint — the module exports almost nothing and throws at import on non-Windows — but that should be named rather than papered over with regex assertions that create the appearance of coverage.

**One concrete regression in existing signal:** `test/tray-service-windows.test.mjs:76-80` asserts `doesNotMatch(result.stdout, /"supported":false/)`. On main that passed because stdout held real JSON. On this PR `status` throws and stdout is **empty**, so the assertion now passes vacuously. The test that exists specifically to prove Windows dispatch works no longer proves anything — and the same masking means `windows-latest` CI cannot detect blockers 1 and 2, because a "did not answer" throw is indistinguishable from success to it.

## Verdict: NEEDS-FIXES

Much of this is better than the diff size suggests — the tri-state task query, locale-independent state read, UTF-8 pinning, stop-before-rebuild ordering, and manifest-based deploy pruning are real improvements, and the elevation *design* is sound. What blocks it is the bare `powershell.exe` on the elevation path, a timeout budget that contradicts the repo's own precedent for the same call and converts routine installs into hard failures with a misleading message, a robocopy pattern that breaks on 7.4+ success, and a UAC env-var handoff I believe has never worked. Also missing: `tray repair` is undocumented, and `deploy-codex-router.ps1` is 254 lines of **destructive** operator surface (it robocopies over an installed router and prunes files) referenced by nothing but its own tests.

PR #377, opened 2026-08-23
feat: Antigravity OAuth provider (reviewed, integration seams closed)

duolahypercho:
Supersedes #372. Carries **@Zzy-min's four commits unchanged** and adds one commit closing the four integration blockers from the review on that PR. Please close #372 in favour of this — the provider is their work, and before that @FIocker's in #349.

I would have pushed this to the fork branch so #372 could merge directly, but writing to a third party's repository is blocked here.

## The OAuth work is unchanged, because it was already right

Nothing in the credential design was touched. Re-stating what was verified so it is on the record: PKCE with an 86-character verifier and S256, `state` compared *before* the authorization code is consumed with a replay guard, the redirect URI pinned to loopback and re-derived through the same helper at both authorize and exchange so it cannot drift, 0600 token files via `writePrivateJson` with `proper-lockfile` around every read-modify-write, tokens added to `knownLocalSecrets()`, project cache keyed on `sha256(refresh_token)`, and `invalid_grant` tombstoning while `invalid_client` does not.

## The four seams

**1. A dropped `await` broke credential removal for every provider.** `removeApiCredential` went sync→async; `control.mjs` was updated, `provider-key.mjs` was not. Deletion still happened, but `removal.removedFiles` was `undefined`, so `provider-key deepseek remove` printed *"No managed DeepSeek API key file exists"* about a file it had just deleted, skipped the picker refresh, and **silently suppressed** *"a key is still available from the macOS Keychain; remove it there to fully disconnect"*. That last part is why this was blocking: someone who believes they disconnected a provider was no longer told the credential still resolves. Now awaited at `provider-key.mjs:277`, with every call site of all three sync→async functions swept.

**2. The forwarder launched unconditionally on port 4212, so a conflict killed the whole router.** Now gated at `start.mjs:135` on `antigravityOAuthStatus().configured`, with spawn, health-wait and exit-race all using the `...(x ? [y] : [])` spread Devin already uses.

Worth recording why *that* predicate: the review suggested mirroring Devin's `devinCliRouted`, and that would have gated nothing — Antigravity's models are checked in, so `MODELS.some(provider === "antigravity-oauth")` is always true. The stored session is the real equivalent, since `providers enable` refuses before sign-in and `configuredProviderIds()` decides on the same status. `AGENTS.md` gains a rule so this is not "corrected" back later.

Four test files spawn `start.mjs` (not the eleven the review claimed — the other matches were comments and `readFileSync` of the source); each now gets a pooled Antigravity port. The reported `gateway-restart` failure was reproduced first, then fixed.

**3. `ANTIGRAVITY_CLIENT_SECRET` was documented nowhere and failed only after the user granted Google consent.** The preflight now runs at `antigravity-oauth-onboarding.mjs:192` — before the URL is built (195), printed (198), the listener bound (290) or the browser opened (292) — and names the variable. Documented in `README.md` and `docs/INSTALL.md`.

The test for this is nicer than an assertion on a message: it holds the callback port open, so "no EADDRINUSE" *proves* the check ran before the bind.

**4. The background service never received the secret**, so even a correct build 502'd within the hour — launchd/systemd do not inherit the shell env, and the allowlist gained the three new ports but not the secret. Now forwarded through one shared helper in all three service files, and added to `knownLocalSecrets()` since it lives in a generated file.

**Also:** `ANTIGRAVITY_REDIRECT_URI` validation moved off module top level. It had been evaluated at import in a module now inside `doctor.mjs`'s and `setup.mjs`'s static graph, so a malformed value took down unrelated commands.

## Verification

- All 6 new tests fail on the unmodified #372 tip.
- `node scripts-check.mjs` — passes
- `node --test test/*.test.mjs` — **2253 tests, 0 failures**, 13 pre-existing skips

## Deliberately not done

The review's four follow-ups are untouched — the missing `AbortSignal` in `ensureFreshAntigravitySession`, the now-dead `source: "fallback"` scaffolding, the unreachable `setup-shared.mjs` copy of `onboardAntigravityOauth`, and the misleading "Re-run sign-in" message when `loadCodeAssist` merely failed to reach Google. Each is a behavior change rather than an integration fix, and they belong in their own change.

## Still a maintainer decision

Whether the project wants to ship a client extracted from Google's Antigravity surface — not a public API — is independent of whether this code is correct. This PR does not settle that.


duolahypercho:
Superseded by #383, which carries every piece of this PR that #372 did not have.

Closing rather than merging because the premise this PR was opened on — "supersedes #372, carries their four commits unchanged" — stopped being true after it was written. @Zzy-min pushed four more commits to #372 that independently fixed the same four seams, and the two branches diverged. Each ended up with work the other lacked, so merging either one alone would have shipped a regression.

Ported from here into #383:

- **`ANTIGRAVITY_CLIENT_SECRET` in `knownLocalSecrets()`** — the one that mattered most. The service definitions now carry the secret, so without this a support bundle could leak it.
- **the `project_source: "fallback"` state** — validator acceptance, a fallback never persisting a `project_id`, `alreadyResolved()` refusing inside the cache TTL instead of routing through a project that does not exist, and `forceFallbackRefresh` letting the forwarder's retry path past that TTL
- **`antigravityClientSecretEnvironment()`** — one helper replacing the same conditional spread inlined in all three service modules
- **the error message naming the missing variable** and explaining that launchd/systemd/Task Scheduler do not inherit a shell
- **the post-sign-in restart notice**

Not carried over: the extraction of `configureProvider` into `setup-shared.mjs`. It is factoring with no behavioural difference, and #372 already has the guided Antigravity path inline at `setup.mjs:336`.

Two notes for the record:

- The Windows failure that blocked this was never this PR's fault. It was a latent `process.exit()` race on `main`, fixed separately in #380; after that landed, `test (windows-latest)` here went green with no change to this branch.
- `test (macos-latest)` was failing here at close (`installer-scripts.test.mjs:204`, *"POSIX install and enable preserve the PATH-selected Node wrapper"*) and was never root-caused. It does not reproduce on #383, which is green across the board — but it is unexplained rather than explained away, so worth remembering if it resurfaces.

PR #379, opened 2026-08-23
feat: ship Ox Alpha on six routes and add Venice and Nous Research providers

duolahypercho:
Ox Alpha is one stealth 1M-context reasoning model that six of this repository's providers resell, each under its own id. Every upstream id and context window below was read from that provider's **own live catalog**, not inferred from the family name.

| Picker label | Model ID | Upstream id | Needs a key |
| --- | --- | --- | --- |
| Ox Alpha (OpenCode Free) | `opencode-free/ox-alpha` | `x-preview-f-free` | no |
| Ox Alpha (opencode Go) | `opencode-go/ox-alpha` | `ox-alpha-free` | opencode |
| Ox Alpha (OpenRouter) | `openrouter/ox-alpha` | `stealth/ox-alpha` | OpenRouter |
| Ox Alpha (Command Code) | `commandcode/ox-alpha` | `stealth/ox-alpha` | Command Code |
| Ox Alpha (Nous Research) | `nousresearch/ox-alpha` | `stealth/ox-alpha` | Nous Portal |
| Ox Alpha (Venice) | `venice/ox-alpha` | `stealth-ox-alpha` | Venice |

All six advertise 1,048,576 tokens of context with 131,072 of output and text+image input, a **low / high / max** effort ladder defaulting to `max`, and zero pricing during the preview, so they carry `isFree` and show a **Free** badge in the control center.

## The effort ladder belongs to the model, not the reseller

Three rungs, and the upstream enumerates them itself:

```
HTTP 400 — [1210] This model always engages in thinking and cannot be
disabled; please use low, high, or max
```

`none`, `off`, `minimal`, `medium`, `xhigh`, `ultra`, `default` and `auto` all draw that response. `low`, `high` and `max` return 200 with monotonically rising reasoning-token counts (2 / 25 / 45), so the three rungs are real behavior, not just enum validation.

**Venice is the one catalog that disagrees, and it is the one to distrust.** It advertises `none/low/medium/high` for this id — but that is its most generic shape, shared with eight unrelated models, and it contains `none`, a rung this model refuses by name. Venice publishes model-specific ladders when it has them (`low/high/max` for GLM-5.3, `none/high/max` for GLM-5.2), so the generic shape here reads as an unverified onboarding default. OpenRouter's live API, Nous Portal's live API, and models.dev for `openrouter`, `opencode-go`, `opencode`, `kilo` and `nano-gpt` all say `low/high/max`.

AGENTS.md records this as the standing exception to "the provider's own catalog decides": when a reseller advertises a rung the model rejects by name, the model wins — and the disagreement gets written down rather than silently resolved.

## Why the request profile is load-bearing

Codex only gained the `max` variant in 0.143.0, so on anything older the clamp in `src/catalog.mjs` rewrites this model's default down to `xhigh` — a rung the upstream refuses. Without a fix, every Ox Alpha turn on a pre-0.143 Codex would 400.

The new `ox-alpha` request profile clamps whatever Codex sent onto the rungs the model's own registry entry declares (`xhigh`/`ultra` → `max`, `medium`/`minimal` → `low`). It reuses the ladder walk the GLM profile already had (`glmEffort` → `declaredEffort`; no behavior change for GLM).

## New providers

**Venice** (`venice`, `https://api.venice.ai/api/v1`) and **Nous Research (Hermes)** (`nousresearch`, `https://inference-api.nousresearch.com/v1`) are new API-key providers: selectable through `install.sh --providers`, `providers enable`, the tray setup card, and `doctor`, each with its own credential isolation and provider mark. Venice carries a `planNote` because a free Venice account has no API entitlement at all — API access needs Pro (the Explorer tier), a funded USD balance, or staked VVV granting VCU.

## Subscription state in the tray and control center

- **Venice** — `api_keys/rate_limits` reports every pool that can fund a request (USD, VCU, and the daily DIEM allowance) plus the API tier, so a full VCU balance no longer displays as a zero.
- **OpenRouter** — per-key spend cap from `/api/v1/key`, plus the account-wide credit pool when the stored key is a management key; the per-key numbers survive when `/credits` answers 403 for an inference key.
- **Nous Portal** — publishes no credits route at all (404 on both `/v1/credits` and `/v1/key`), so it degrades to its subscription page and observed router traffic rather than inventing a number.

Provider marks for Venice, Nous Research, and OpenRouter ship for both the tray and the control center, sourced from each official site and adapted to the white-on-transparent form the Island's black bezel needs. Provenance is recorded in both `SOURCES` files.

## Verification

- `npm run check` and `npm test` — 2062 passing, 0 failing.
- `apps/control-center`: `npm run check` (tsc) and `npm run build` both clean; the new assets bundle.
- `apps/macos/ModelRouterTray`: `swift build -c release` and `swift test` both clean.
- **Live, at zero cost**, through the credential-free OpenCode route: every rung probed individually (see the ladder section above); raw `xhigh` is a 400 upstream while the same request through the router's own forwarder returns 200 with reasoning tokens; forced `tool_choice: "required"` returns a well-formed tool call.
- `doctor`, `providers list`, `provider-key … status`, and `control provider-usage` all report the three providers correctly.
- New coverage in `test/ox-alpha.test.mjs` and `test/routing.test.mjs` pins the upstream ids, the ladder, the pre-0.143 clamp/undo round trip, and each new account-usage surface.

## One related fix

The official-name table in `src/user-models.mjs` was overwriting a checked-in fragment's `displayName`. It exists to fill in a name *curation* could not know, so it now applies only to curated entries — otherwise it flattened "Ox Alpha (OpenCode Free)" back to the curated label and lost the only thing telling the six routes apart.

## Note

Free here is a preview, not a property. No lab has claimed this model and the routes can narrow or withdraw it without notice. If they start billing it, the honest change is to drop `isFree` and rewrite the descriptions rather than leave a Free badge on a metered model.


mirucon:
I independently tested the OpenCode Go `ox-alpha-free` upstream through this router (locally curated as `opencode-go/ox-alpha-free`). The repository's live compatibility test passed basic response, streaming, forced function calling, and compaction; separate Codex probes also passed native `apply_patch`, low/high/max reasoning with reasoning kept separate from the final message, and native image input without the vision bridge.

One context-source discrepancy seems worth resolving. The current OpenCode Go-specific record in `anomalyco/models.dev`, added in commit `f0d0881`, publishes:

```toml
[limit]
context = 1_000_000
output = 131_072
```

https://github.com/anomalyco/models.dev/commit/f0d08819f487576b0ef15ba178bacfad753901f8

This PR's `config/opencode/go/ox-alpha.json` uses `1_048_576` and says all six routes advertise that value. The public OpenCode Go `/models` payload does not expose context lengths, so I could not find another Go-specific published source for `1_048_576`.

I did not probe the upper boundary above 1M, so this is not evidence that `1_048_576` is rejected; it is a source/provenance question. Could you point to the Go-specific source or served-cap probe behind `1_048_576`? Absent that, `1_000_000` appears to be the currently published conservative value for this route.

PR #381, opened 2026-08-23
fix: keep test runs out of this machine's systemd

duolahypercho:
The service write guard has two halves. The unit-file half was already on the Linux module — a test that has not redirected `XDG_CONFIG_HOME` cannot write the unit. The service-manager half was not, and it is the half that does the damage.

`XDG_CONFIG_HOME` moves where the unit is written, not where the running systemd reads from. So an install that isolated itself correctly still reached `systemctl --user enable --now codex-router.service` against the developer's own unit, stopped their router, and — once the test removed its temporary directory — left systemd pointing at a unit that no longer existed. Nothing in the run said so.

launchd had exactly this failure and already skips it (`eff69e7`, `bc9a39c`); Windows followed in `f57af1a`. systemd was left behind when the skip was scoped to launchd, so `src/service-linux.mjs` had zero references to `skipServiceManagerCall` and the guard suite had zero Linux coverage.

## Change

- Route the mutating `systemctl` verbs (`daemon-reload`, `disable`, `enable`, `restart`, `start`, `stop`) through `skipServiceManagerCall`, matching `service-macos.mjs`.
- Reads stay live, for the same reason they do under launchd: `status` has to keep answering whether the unit is really active, or the doctor reasons from a state the skip invented.

## Tests

- A Linux-gated test mirroring the existing launchd one — deliberately omits the opt-out flag, asserts the machine's own unit enablement is unchanged and the fixture still gets its unit. Runs on CI's `ubuntu-latest` job.
- Two recorder tests that run anywhere a POSIX shell does, so the verb list is pinned on macOS too, where the module is driven against stubs and only the explicit flag suppresses the call.

Verified the skip test fails against the unpatched module and passes with it. `npm run check` and the four service suites pass (27 passed, 0 failed, 2 platform-skipped), with the machine's launchd registration byte-identical before and after.

## Context

This surfaced from a live incident: the router was dead with `Connection failed: error sending request`, and `launchctl print` showed the label registered against a deleted test fixture in a 65-iteration crash loop. That was the macOS instance of this bug, already fixed on `main`. This closes the remaining platform.

Note that any branch predating `eff69e7` still re-breaks a developer machine when its tests run — see PR #379.


duolahypercho:
Refreshed onto current `main` (`566c9e4`, which now carries #380, #374, #376, #375). One failure, and it is in the control test rather than in the guard — the shipped change looks right.

## What fails

`test (ubuntu-latest)` → `test/service-write-guard.test.mjs:329`, *"without the flag the same install still drives systemd"*:

```
+ actual - expected
+ []
- [ 'daemon-reload', 'stop codex-router.service', 'enable --now codex-router.service' ]
```

## Why

```js
export function skipServiceManagerCall({ hostManaged = true, env = process.env } = {}) {
  if (serviceManagerDisabled(env)) return true;
  return Boolean(env.NODE_TEST_CONTEXT) && hostManaged;
}
```

On the ubuntu job `HOST_MANAGED` is `process.platform === "linux"` → `true`, and `linuxInstall()` sets `NODE_TEST_CONTEXT: "child-v8"`. So every mutating verb is skipped **regardless of `MODEL_ROUTER_SKIP_SERVICE_MANAGER`** — which is exactly the protection this PR adds. The recorder therefore logs nothing and the control's `deepEqual` cannot hold.

It passes on macOS for the reason the commit message gives: `HOST_MANAGED` is false there, so only the explicit flag suppresses the call. The message reasons about macOS and win32, but not about the module's *own* host, where the skip is unconditional inside a test context by construction. The skip clause only excludes win32:

```js
{ skip: process.platform === "win32" && "the recorder stub needs a POSIX shell" },
```

## Suggested fix

The control cannot be satisfied on linux, so exclude it there as well:

```js
{
  skip:
    (process.platform === "win32" && "the recorder stub needs a POSIX shell")
    || (process.platform === "linux"
      && "on systemd's own host the guard suppresses the call regardless of the flag"),
},
```

Coverage is not lost: the macOS recorder still pins the exact verb list, and the linux-gated test above it still proves the machine's own unit is left alone — which is the claim that actually matters here.

## Ordering note

`src/service-linux.mjs` is also touched by #372 and #377. Whichever of those lands first will make this stale on a shared file, so it is worth landing this one before the Antigravity work rather than after.

*Not pushed — leaving the branch alone since it is being worked in another session.*

PR #382, opened 2026-08-23
fix(windows): bound service process identity probes

virajchoudhary:
## Summary

Bound the synchronous PowerShell process-identity probes used by the Windows managed service so a stuck `Get-Process`, CIM, or WMI query cannot block router startup indefinitely.

## Why

While diagnosing a Windows install, the Scheduled Task launcher chain was alive (`wscript -> cmd -> node start.mjs`) but none of the router ports had bound. An absolute-path foreground `start.mjs` invocation later started normally, and the managed service also recovered on a later retry, so this PR does **not** claim that the process-identity probe was conclusively the root cause of that one transient incident.

It does address a concrete robustness issue found during that investigation: `processStartIdentity()` and `processCommandLine()` use synchronous PowerShell subprocesses on Windows with no timeout. If one of those OS queries stalls under Task Scheduler, `start.mjs` can remain blocked before the router starts listening.

## Change

- add a 5 second timeout to the Windows `Get-Process` identity probe
- add the same timeout to both command-line probes (`Get-CimInstance`, then `Get-WmiObject` fallback)
- leave non-Windows probe behavior unchanged
- preserve fail-closed ownership semantics: a timeout still produces no verified identity/command line, so the router refuses to adopt or act on an unverified process

## Tests

Added `test/process-identity.test.mjs` covering:

- bounded Windows identity probing
- bounded CIM/WMI fallback behavior
- unchanged non-Windows spawn options

Targeted validation performed before opening this PR:

```text
node --check src/process-identity.mjs
node --check test/process-identity.test.mjs
node --test test/process-identity.test.mjs
```

Result: 3/3 tests passed.

The repository CI can provide the full macOS/Linux/Windows suite required by the development guide.

## Scope

This is intentionally a narrow service-start robustness change. It does not alter process ownership checks, service lifecycle behavior, routing, provider handling, or credentials.


virajchoudhary:
Additional live reproduction after the earlier exit-with-result-1 case:

The bundled venv interpreter is healthy when invoked directly five times in a row:

```text
Python 3.13.4
runs: 0.084-0.144 s, exit 0
```

Then starting the existing Scheduled Task directly:

```text
Start-ScheduledTask -TaskName "Codex Router"
```

produced, after 5 seconds:

```text
TaskName: Codex Router
State: Running
LastTaskResult: 267009 (0x41301, task currently running)
ports 4200-4203: none listening
http://127.0.0.1:4202/health: connection failed
```

So on the same machine we have now observed both Windows failure shapes:

1. task exits with result 1 and the outer readiness loop keeps waiting;
2. task remains Running with no router ports bound.

This second shape again matches the earlier live-but-not-listening observation and makes a pre-main/synchronous-probe stall worth investigating, but it still does not prove that `process-identity.mjs` is the exact root cause. Capturing the live process tree next, especially any `powershell.exe` descendant and its command line, before stopping the task.

virajchoudhary:
Status check from the machine that produced the #384 reproductions, since this PR's scope was re-examined during that investigation:

- The failure occurrence that motivated this PR showed the launcher chain alive (`wscript -> cmd -> node start.mjs`) with no ports bound; a later reproduction in #384 found **no** PowerShell process-identity probe running at all during a stuck state. That confirms what the PR description already says carefully: the unbounded probes were a real robustness gap, but they are not proven to be the root cause of every observed startup stall.
- The branch still applies cleanly against current `main` (`src/process-identity.mjs` is otherwise unchanged upstream); GitGuardian passes. No code changes pushed here — no reviewer feedback has arrived and nothing in the new evidence contradicts the bounded, fail-closed design (timeout -> no verified identity -> refuse to act).
- The remaining startup failures are now covered by #384 / #387.

PR #385, opened 2026-08-23
fix: filter empty DeepSeek assistant messages after tool calls

mayrazing:
## Why

While using DeepSeek, we found that tool-call responses can include an additional empty assistant message.

Codex treats this empty message as a separate content block, causing unusually large gaps between tool activity entries.

In the actual response, the empty message was preceded by a `reasoning_text` event. The previous filter incorrectly treated this event as visible assistant text and released the empty message.

## What

- Correct the detection of empty assistant messages.
- Do not treat `reasoning_text` as visible assistant text.
- Treat only non-empty `output_text` or `text` as visible assistant text.
- Prevent empty messages associated with tool calls, along with their related events, from being forwarded to Codex.
- Add a regression test matching the event sequence observed in the actual DeepSeek response.

## Test

- Reproduced and confirmed the event sequence with one real DeepSeek request.
- Updated the test fixture to include the `reasoning_text` event observed in the actual response.
- Before the fix, the regression test failed because the empty assistant message reached Codex.
- After the fix, the same regression test passed.
- `node --check src/tool-message-compat.mjs` passed.
- `git diff --check` passed.

## Test Scope

Only DeepSeek was tested with a real request in this change.

The filtering logic applies to external models connected through the Router. If another external model returns the same empty-message pattern, this logic will also run, but no real requests were made to other external models as part of this change.

Native Codex GPT models do not pass through this external-model filtering logic and are therefore unaffected.

### DIFF
--- a/src/router.mjs
+++ b/src/router.mjs
@@ -45,6 +45,7 @@ import {
   writeStreamErrorEvent,
 } from "./http-utils.mjs";
 import { EmptyCompletionGuard } from "./empty-completion-guard.mjs";
+import { toolMessageCompatTransform } from "./tool-message-compat.mjs";
 import {
   ZaiResponsesCompatTransform,
   zaiResponsesCompatTransform,
@@ -2912,6 +2913,10 @@ async function handleResponses(request, response, requestUrl) {
             : undefined,
       });
       const transforms = [usageObserver];
+      const toolMessageCompat = route
+        ? toolMessageCompatTransform(contentType)
+        : undefined;
+      if (toolMessageCompat) transforms.push(toolMessageCompat);
       let envelopeCompat = route
         ? zaiResponsesCompatTransform(route.provider, contentType)
         : undefined;
--- a/src/tool-message-compat.mjs
+++ b/src/tool-message-compat.mjs
@@ -1,0 +1,339 @@
+import { Transform } from "node:stream";
+import { StringDecoder } from "node:string_decoder";
+
+function parsedEventBlock(block) {
+  const newline = block.includes("\r\n") ? "\r\n" : "\n";
+  const lines = block.split(/\r?\n/);
+  const dataLineIndex = lines.findIndex((line) => line.startsWith("data:"));
+  if (dataLineIndex === -1) return undefined;
+  const dataText = lines[dataLineIndex].slice(5).trimStart();
+  if (!dataText || dataText === "[DONE]") return undefined;
+  try {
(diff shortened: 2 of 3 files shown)


duolahypercho:
Not safe to merge yet. The blank-message transform buffers an unbounded SSE stream before any bounded/liveness-aware guard, delays reasoning/tool output, suppresses output index 0 without renumbering later indices, and is applied globally although only DeepSeek is covered. Please scope it to the affected provider, bound the transform before buffering, preserve streaming, renumber emitted indices, and add mixed reasoning/tool/large-stream tests.

jepgambardella:
Agreed. I will not treat this PR as merge-ready until the transform is provider-scoped, bounded and streaming-safe, indices are renumbered, and mixed reasoning/tool/large-stream coverage is added.

duolahypercho:
A hardened replacement is now open as #435. It addresses this PR's review blockers with direct-DeepSeek/SSE scoping, exact-empty proof, full output-index and terminal-order validation, strict byte/time/frame bounds, and fail-open behavior for ambiguous streams. This PR should remain blocked; review and CI should proceed on #435.

mayrazing:
Thanks for the review. I was a bit surprised to see https://github.com/duolahypercho/codex-router/pull/435 supersede this PR, since I had understood the feedback above as changes for me to address and was ready to work on them.

For future contributions, I'd appreciate having a chance to address the review feedback before the implementation is replaced.

12duolalalala:
Hold. This is CONFLICTING with main.

It overlaps #435, which is meant to supersede this with a provider-scoped, bounded DeepSeek-only transform. This one adds a global tool-message compat transform on router Responses.

Close this, or rebase only after #435 lands or is rejected. Do not merge both.

12duolalalala:
You're right to call that out. The original report and the reasoning_text catch were yours. #435 is a scoped, bounded rewrite of the same bug, not a dismissal of that work. Sorry it landed as a replacement before you had a chance to address the review.

This PR stays open. I'm rebasing #435 onto current main because that branch is already DeepSeek-only and fail-open. If you'd rather keep #385 as the landing vehicle and take the review comments yourself, say so and I'll hold #435.

We will not merge both. Either way you get credit for finding this.

mayrazing:
Thanks for the follow-up. Since #444 has now landed the replacement implementation, I’ll leave #385 as the original report and initial investigation for this bug. Glad the DeepSeek issue is fixed.

PR #387, opened 2026-08-23
fix(windows): fail readiness early on a dead task; retry transient interpreter probes

### LINKED ISSUE #384: Windows service startup can exit early on transient spawn timeout, then repair waits full 300s
## Summary

On Windows, `doctor --fix` can spend the full 300-second readiness budget after the Scheduled Task has already exited with `LastTaskResult = 1`. The service log also shows a transient `spawnSync(...python.exe) ETIMEDOUT` being classified as a broken LiteLLM venv even though an immediate later `doctor` successfully runs the same interpreter.

This was found while debugging #382, but it appears to be a separate failure mode from that PR's process-identity timeout hardening.

## Observed behavior

`doctor --fix`:

```text
Router did not become healthy within 300 seconds: fetch failed
Background-service installation failed.
```

Immediately afterward:

```text
service.mjs status: {"installed":true,"loaded":false,"state":"ready"}
Scheduled Task: Ready
LastTaskResult: 1
ports 4200-4203: none listening
router process tree: no matching wscript/cmd/node process
```

So this occurrence was not a live-but-stuck service. The task had already exited and the outer readiness loop continued polling until the full 300 s budget expired.

Relevant startup log entries from the same machine include:

```text
[model-router] startup failed: Timed out waiting for OAuth forwarder to become healthy (the service did not answer within 1000 ms).

Error: The LiteLLM virtual environment is broken at ...\.venv\Scripts\python.exe (spawnSync ...\python.exe ETIMEDOUT).
```

An immediate later `doctor` reports:

```text
OK LiteLLM venv runtime: ...\.venv\Scripts\python.exe runs
```

## Code paths

- `src/service.mjs` waits `READINESS_TIMEOUT_MS = 300_000` after a successful platform `start` command and does not observe whether the Windows Scheduled Task exits during that wait.
- `src/venv-runtime.mjs` performs `spawnSync(python, ['--version'])` with a 15 s timeout and returns any `probe.error` as a runtime problem. That makes `ETIMEDOUT` indistinguishable from persistent venv corruption to `start.mjs`.
- `src/health-probe.mjs` already treats an internally timed-out HTTP health probe as inconclusive and retries with a wider window, which suggests similar transient-timeout handling may be appropriate for process launch.

## Expected behavior

Two possible improvements, likely separable:

1. Windows service start/repair should fail early with the Scheduled Task's exit/result when the managed task has already terminated, instead of polling router health for five minutes.
2. A transient interpreter spawn timeout should not be reported as a definitively "broken" venv. It could be retried with a wider bounded window or surfaced as a distinct transient startup failure while preserving a hard bound.

## Notes

This issue does not claim the OAuth-forwarder timeout and Python spawn timeout have one root cause. Both point to severe transient Windows process/scheduling delay on this machine, but the actionable defects above are about how those conditions are classified and reported.

No credentials or caller URLs are included here.

virajchoudhary:
## Summary

Fixes the two actionable defects from #384. Fixes #384.

1. **Readiness no longer polls a dead task for 300 seconds.** Task Scheduler can report `State=Running` while `GetInstances(0)` enumerates zero running instances, and `LastTaskResult` preserves why the task exited. `src/service.mjs` now waits through a readiness guard that consults that enumeration: a brief absence is tolerated as launch lag, and an absence persisting past a grace window fails early with the task's own result (e.g. `LastTaskResult=0x1`) instead of waiting out the full budget.
2. **A transient interpreter spawn timeout is no longer reported as permanent venv corruption.** A `spawnSync ETIMEDOUT` on the venv's `python.exe` means Windows never finished scheduling the child — on the reporting machine the same interpreter ran five times in 0.084-0.144s immediately afterward. The probe now retries once with a wider hard bound before reporting, and the persistent-timeout wording names scheduling pressure rather than a broken venv.

## Design notes

- Fail-closed semantics preserved everywhere: a timeout or query failure produces no verified state. An unavailable Task Scheduler query stays inconclusive so a restricted shell cannot turn a slow-but-valid startup into a false failure; only a *confirmed persistent* zero-instance state fails early.
- No timeout budgets were raised except one bounded retry inside the existing probe contract (15s then up to 45s), per the issue's "do not blindly increase timeouts" note.
- Non-Windows behavior unchanged: the guard delegates to the existing health wait.
- Tests never touch the real Scheduled Task: task-state queries and health waits are injected fakes in every case.

## Tests

- `test/windows-task-state.test.mjs`: COM-based instance count + LastTaskResult parsing, query-failure/malformed-output staying inconclusive, non-Windows never invoking Task Scheduler.
- `test/service-readiness.test.mjs`: early failure with the task's result after persistent absence, launch lag tolerated, unavailable queries failing open, non-Windows health-only, and the service delegating its budget to the guard.
- `test/venv-runtime.test.mjs`: transient timeout retried with the wider bound; persistent timeout worded as scheduling pressure; non-timeout failures not retried.

`npm run check` passes. Targeted suites pass locally on Windows; the repository CI provides the full macOS/Linux/Windows matrix.


virajchoudhary:
Live Windows validation on the machine that produced #384, against the real Scheduled Task (no test fixtures):

**Reproduced the stale-instance mode this guard exists for.** After a stop/start cycle, Task Scheduler reported one running COM instance with `LastTaskResult=0x41306` while zero router processes existed and nothing listened on 4200-4203. The first version of this branch, which relied on `instanceCount === 0`, waited its full budget in exactly that state — which forced the second fix in this PR:

- `waitForRouterHealth` resolves `{ ok: false }` rather than rejecting; the guard originally treated fulfillment as success and could resolve with a failure payload instead of throwing.
- The COM instance enumeration outlives its process, so instance count alone is not a death signal.

The probe now also scans `Win32_Process` for a live wscript/cmd/node command line referencing the generated launcher (excluding the probing PowerShell's own PID — its `-Command` string carries the match literals, which self-matched until excluded). Against that same broken state the guard now fails in ~22s with `Windows Scheduled Task has no running launcher process (LastTaskResult=0x41306)`.

**Healthy-path validation after `doctor --fix`:**

- `Start-ScheduledTask` → readiness resolved healthy in 35s
- ports 4200/4201/4202/4203/4208 all listening
- `/health`: `ok=true`, `degraded=[]`
- `doctor` exit 0, all core checks OK
- live compatibility probes through the installed router: `opencode-free/x-preview-f-free` PASS (basic/streaming/tool-calling/compaction) and `opencode-free/ox-alpha` PASS (same four)

PR #399, opened 2026-08-24
fix: isolate test service and runtime fixtures

jepgambardella:
Tests must never replace the user's running router or its Node runtime.

This change keeps redirected LaunchAgents out of the real launchd domain and uses a temporary executable wrapper for the broken-venv fixture. The fixture can now be rewritten safely without following a symlink into the Node installation.

Validation:
- `node --test test/service-write-guard.test.mjs test/service-render.test.mjs`
- `node --test test/install-plan.test.mjs test/venv-runtime.test.mjs`
- `node --test test/routing.test.mjs`
- `npm run check`

### DIFF
--- a/src/dsh-config-manager.mjs
+++ b/src/dsh-config-manager.mjs
@@ -87,6 +87,12 @@ function normalizeTrailing(lines) {
   return copy;
 }
 
+function credentialPath(document, reference) {
+  return document.root.children.has("refs")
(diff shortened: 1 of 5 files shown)


jepgambardella:
Superseded by the service-manager isolation already merged on main. The remaining venv fixture fix will be evaluated separately against current main.

PR #402, opened 2026-08-24
feat: add secure provider credential primitives

jepgambardella:
## Summary

Adds provider-neutral credential references with strict metadata validation, target binding, confined protected storage, redacted support output, and safe migration rollback. Existing provider files and Keychain entries remain the secret authority.

## Reason

Prevent provider credentials from being copied into router metadata, logs, support bundles, or arbitrary paths while keeping migration recoverable.

## Validation

- `npm run check`
- `node --test test/provider-credential-store.test.mjs test/provider-credentials.test.mjs test/support-bundle.test.mjs` (13 passed)
- `git diff --check`
- Broader provider/catalog suites require missing local dependencies (`undici`, `proper-lockfile`) in this checkout.

## Remaining risks

This is protected metadata and reference-based credential handling, not encrypted secret storage. Existing provider-specific files, environment variables, and Keychain remain responsible for secret protection. The PR does not claim encryption.

PR #403, opened 2026-08-24
feat: add safe provider API-key pools

jepgambardella:
## Summary

Add a provider-scoped API-key pool that stores validated secret references and metadata, never raw keys. A configured pool is authoritative, uses locked atomic state updates, rejects duplicate references and resolved secrets, and retries only transient failures before response bytes are committed. Provider protocol variants share their canonical pool; pool references are constrained to registry-declared sources and honor `--no-discovery`.

## Reason

Keep one broken key from taking down a provider while preventing accidental legacy-key fallback, cross-provider credential use, duplicate-secret routing, and retries after the client has received response bytes.

## Validation

- `npm run check`
- `node --test test/provider-api-key-pool.test.mjs test/provider-api-key-routing.test.mjs test/provider-credentials.test.mjs test/routing.test.mjs` (101 passed)
- `git diff --check`

## Remaining risks

This PR adds the provider-scoped pool primitive and its forwarder boundary. It does not add automatic account pooling or ChatGPT subscription routing.

### DIFF
--- a/src/api-forwarder.mjs
+++ b/src/api-forwarder.mjs
@@ -30,7 +30,7 @@ import { stripImages, supportsImageInput } from "./vision-bridge.mjs";
 import {
   credentialLabel,
   credentialStatus,
-  resolveProviderCredential,
+  resolveProviderCredentialReference,
 } from "./provider-credentials.mjs";
 import {
   ensureFreshGitHubCopilotSession,
@@ -45,6 +45,12 @@ import { relayCommandCodeGenerate } from "./commandcode-relay.mjs";
 import { VERSION } from "./version.mjs";
 import { installStableFetchTransport } from "./fetch-transport.mjs";
 import { zaiCacheUsageTransform } from "./zai-cache-usage.mjs";
+import { threadIdFromHeaders } from "./codex-session-names.mjs";
+import {
+  recordProviderApiKeyRequestOutcome,
+  resolveProviderApiKeyForRequest,
+} from "./provider-api-key-routing.mjs";
+import { runProviderApiKeyAttempts } from "./provider-api-key-pool.mjs";
 
 installStableFetchTransport();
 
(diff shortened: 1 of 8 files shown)


duolahypercho:
Closing because the pool implementation is not safe. It does not provide pre-commit failover, can mark attempts committed after bytes were relayed, fails open to the legacy key, permanently disables keys on ordinary 400/404 responses, permits the same secret through multiple refs, and writes shared state without locking. It also depends on the unsafe credential foundation in #402.

duolahypercho:

12duolalalala:
Hold. Changes live auth in api-forwarder: resolveProviderApiKeyForRequest plus runProviderApiKeyAttempts. A configured, invalid, empty, or locked pool fails closed instead of using the legacy single key.

That can take a provider offline if pool state is wrong. Also overlaps #402 credential helpers and #405 on api-forwarder. Do not merge them together without a combined rebase.

duolahypercho:
Thanks for revising the fail-closed behavior. This still needs reconciliation with #402 before it can enter the live auth path. The branch defines a second credential-reference format instead of consuming the shared credential IDs/resolver, and no CLI/control path invokes the pool upsert, pause, or policy operations, so users can configure it only by hand-editing state. Please restack on the accepted credential foundation, expose a safe lifecycle, and add an end-to-end configure, route, and pre-commit failover test. I am keeping it open rather than merging a second credential contract.

MeltingMadness:
I restacked this work on current `main` and addressed the three blockers from the latest maintainer comment in a review branch:

https://github.com/MeltingMadness/codex-router/tree/codex/pr403-key-pools-restack

Integration commit: https://github.com/MeltingMadness/codex-router/commit/201d40b

Changes:
- removed the second pool-specific secret-reference contract; pool entries now contain only shared `cred_...` IDs and resolve through the accepted provider credential store/resolver
- added safe lifecycle commands through `codex-router key-pool`: existing credential upsert, registry-constrained `add-env`, pause/resume, policy, and sanitized status
- added an end-to-end OpenCode forwarder test that configures two credential IDs, receives a pre-commit 429 on the first key, rebinds to the second key, and returns only the successful response
- retained fail-closed behavior for configured invalid/empty/unresolvable pools and kept pool state free of secret bytes

Targeted verification on the branch:
- syntax checks for the changed runtime modules
- 29 relevant tests passed, 0 failed
- both CLI dispatch paths checked
- `git diff --check` clean

I cannot directly update the contributor-owned PR branch from this account, so this branch is ready to cherry-pick or compare while keeping #403 as the review thread.

MeltingMadness:
The restacked implementation is now available as an official pull request against current `main`: #477. Please continue implementation review there; this keeps the revised branch directly visible in the repository PR list and CI.

duolahypercho:
Maintainer disposition: please do not merge the exact #403 head. Its useful key-pool work is preserved in current-main successor #477, which removes the duplicate secret-reference contract and adds the missing lifecycle and pre-commit rotation path.\n\nThe independent #477 audit is complete and recommends ADAPT rather than merge: winning-account identity is not propagated for Command Code, pooled error bodies are unbounded, retry classification can duplicate 500 work while missing nested transport errors, stale selected keys can still be sent after pause, removal/Windows/doctor/health integration is incomplete, and pool-state bounds fail open. Detailed reproduction and repair guidance are recorded on #477.\n\nI am keeping #403 open as the original contribution/review history. We will repair the successor rather than discard this contributor work or merge the superseded stack.

duolahypercho:
Closing as superseded by #477. The revised provider API-key pool implementation is being maintained and reviewed there on current main. This original PR and its review history are preserved for context; this head will not be merged.

PR #404, opened 2026-08-24
feat: add generic OpenAI-compatible providers

jepgambardella:
## Depends on

- #402 provides the reviewed provider credential metadata and protected-state primitives used by this change.

## Summary

- Add user-defined OpenAI-compatible provider descriptors with versioned, private metadata storage.
- Validate endpoint schemes, private-network opt-in, request paths, headers, DNS results, redirects, and response sizes at the request boundary.
- Keep credential references opaque and fail closed when the referenced credential is unavailable or is not an API key.
- Expose generic provider CRUD and health probes through the existing provider CLI.

## Reason

Users need a safe way to describe OpenAI-compatible local or remote endpoints without putting secrets in provider descriptors, logs, or model catalogs. The request boundary must re-check the destination for every call so a DNS change or redirect cannot turn a configured endpoint into an unintended network pivot.

## Validation

- `npm run check`
- `node --test test/generic-providers.test.mjs test/provider-credential-store.test.mjs test/provider-credentials.test.mjs test/providers.test.mjs` (20 passed)
- `git diff --check`
- Merge-tree checked against #402 (`25521f7`); the branch contains no duplicate credential-store implementation.

## Remaining risks

This PR deliberately stops at the safe descriptor and request boundary. It does not merge generic descriptors into the model catalog or claim end-to-end Codex routing. A follow-up adapter/catalog change must use `requestGenericProvider` and its bounded response handling before advertising models. Credential metadata remains subject to the provider-specific binding supplied by #402; unavailable references fail closed.

### DIFF
--- a/src/control.mjs
+++ b/src/control.mjs
@@ -660,6 +660,11 @@ async function printProviderOnboarding() {
   process.stdout.write(`${JSON.stringify(providerOnboardingSnapshot(), null, 2)}\n`);
 }
 
+async function handleGenericProviders(...commandArgs) {
+  const { runGenericProviderCli } = await import("./generic-providers.mjs");
+  await runGenericProviderCli(commandArgs, { output: process.stdout });
+}
+
 async function installProviderCli(providerId) {
   const { installOauthCli, providerOnboardingSnapshot } = await import("./provider-onboarding.mjs");
   installOauthCli(providerId);
@@ -2786,6 +2791,8 @@ if (args.includes("--probe")) {
   await printProviderUsage();
 } else if (args[0] === "providers") {
   await printProviderOnboarding();
+} else if (args[0] === "generic-providers") {
+  await handleGenericProviders(...args.slice(1));
 } else if (args[0] === "install-cli") {
   if (!args[1]) throw new Error("Usage: control install-cli <oauth-provider>");
   await installProviderCli(args[1]);
--- a/src/generic-provider-identity.mjs
+++ b/src/generic-provider-identity.mjs
@@ -1,0 +1,14 @@
+import { PROVIDERS } from "./model-registry.mjs";
+
+export const GENERIC_PROVIDER_ID_PATTERN = /^[a-z0-9][a-z0-9-]*$/;
+
+export function normalizeGenericProviderId(value) {
+  const providerId = typeof value === "string" ? value.trim() : "";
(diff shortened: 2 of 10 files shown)


duolahypercho:
Closing as unsafe. Endpoint validation is onboarding-only and remains vulnerable to DNS rebinding/redirect/private-network pivots on the actual request path; credentials are not reliably resolved, static headers can carry secrets around the reference model, and persisted changes are unlocked/lossy. Generic providers are not yet safely routable. Rebuild around per-request destination validation, redirect controls, bounded reads, confined credentials, and atomic state.

duolahypercho:

12duolalalala:
Hold. +2170 stacked on #402. Adds generic provider CRUD through the existing provider CLI/control path plus request-boundary DNS/redirect checks.

This is not a tiny isolated patch and shares credential commits with #402 / #406. Land #402 first, then this, then any catalog adapter.

duolahypercho:
The request-boundary DNS, redirect, and response-size work is much improved. After #402 lands, please rebase so this PR contains only the generic-provider layer and add a usable credential creation path. Currently the credential foundation rejects IDs outside the checked-in registry while this PR requires credential.providerId to equal the generic provider ID; every authenticated generic provider is therefore permanently unavailable through public APIs. Add a regression test that creates the credential and provider only through public APIs, then proves Authorization reaches the permitted mock endpoint without appearing in descriptors or output.

duolahypercho:
I pushed maintainer repair 1196c70 and restacked this PR onto current main, so the merged #402 foundation is no longer duplicated. Generic providers now have an explicit public credential creation/resolution path while built-in provider validation remains strict; generic IDs cannot collide with built-ins; provider-file paths, redirects, and destination boundaries fail closed; and secret values are redacted from descriptors, CLI/list output, metadata, and support bundles. The end-to-end test proves Authorization reaches only the permitted mock endpoint and the redirect trap receives nothing.

Verification: clean npm ci with zero vulnerabilities; focused credential/generic/support tests pass 24/24; npm run check and git diff --check pass; full suite passes 2,588 with 17 platform skips. Waiting for refreshed CI before merge.

duolahypercho (approved):
Maintainer repair supplies a confined public credential lifecycle for generic providers without weakening built-in validation. End-to-end non-disclosure and destination confinement plus the full suite are green; approved pending refreshed CI.

PR #408, opened 2026-08-24
feat: add virtual model combos with weighted failover

jepgambardella:
## Summary

- Harden virtual model-combo selection with canonical slugs, explicit health, capability checks, bounded sticky affinity, weighted cursor state, and attempt-boundary commits.
- Reject stale completions instead of overwriting concurrent state.
- Keep combo policy unadvertised: no catalog publication or request-path wiring is included.

## Reason

This makes the resolver safe to integrate later without optimistic health, unstable weights, stale cursor/affinity writes, or provider metadata mismatches.

## Validation

- `node --test test/model-combos.test.mjs` (11 passed)
- `npm run check`
- `git diff --check`

## Remaining risks

Runtime catalog publication and provider request failover are intentionally not implemented here. The future integration must persist the returned state at the actual request attempt boundary and only advertise combos after end-to-end provider/recovery tests.

### DIFF
--- a/docs/MODEL-COMBOS.md
+++ b/docs/MODEL-COMBOS.md
@@ -23,9 +23,10 @@ Targets use the canonical registry `slug` and its owning `provider`:
 
 Use `beginComboAttempt` to select a target without changing state. Call
 `completeComboAttempt` only after the request attempt has ended. Successful
-attempts commit bounded stickiness; retryable failures clear the affinity and
-advance the weighted cursor past the failed target. A stale concurrent attempt
-is rejected instead of overwriting the persisted cursor.
+attempts commit bounded stickiness; retryable failures clear the session
+affinity and, for round-robin, advance the weighted cursor past the failed
+target. Every completion advances a per-combo revision, so a stale concurrent
+attempt is rejected instead of overwriting persisted state.
 
 Health is fail-closed: every candidate needs an explicit healthy snapshot (or
 an `isHealthy` callback returning `true`). Missing health, missing model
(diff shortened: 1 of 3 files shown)


duolahypercho:
Closing as unsafe. Combination routing is not wired to real failures, resolves the wrong model identity field, rejects valid tool-capable models, reports optimistic health, lacks cursor persistence, and has unstable weight/stickiness behavior. Reintroduce only after the underlying routing primitives are safe, with real attempt-boundary and recovery tests.

duolahypercho:

12duolalalala:
Hold. The macOS test job failed. This is also a 589-line weighted combo resolver.

Keep it unadvertised until CI is green and there is a real request-path integration. Virtual combo models can change which upstream actually answers.

duolahypercho:
The policy is safer than the original revision, but it should remain unadvertised until integrated. Candidate capabilities currently come from the caller-supplied models array, so a future discovery caller could grant tools, vision, or search simply by setting metadata. Please bind capability and route identity to an authoritative registry record, wire persistence at the actual request-attempt boundary, and add concurrent failover tests. The current macOS CI failure also needs a rebase and rerun.

duolahypercho:
Closing as superseded. The current main branch already ships request-path-integrated failover in src/model-failover.mjs and src/router.mjs with stricter invariants: pre-first-byte only, bounded hops/deadline, provider-family separation, authoritative model identity, cooldown handling, pristine payload rebuilds, and usage/logging coverage. This PR would add a second unadvertised resolver without production persistence or attempt-boundary integration. Its macOS check is also red. If a future combo product is wanted, it should extend the existing failover path rather than introduce a parallel policy module.

PR #410, opened 2026-08-24
fix: keep sub-agent provider identity in fallback

jepgambardella:
## Summary

Keep subagent fallback planning fail-closed. Runtime candidates are matched by exact model and provider identity to immutable checked-in `multiAgentVersion = "v2"` routes; caller metadata cannot grant capabilities or certification. Retry plans are deterministic, deduplicated, weight/attempt bounded, exclude the exact failed route, and stop for committed bytes, application failures, or incomplete failure identity.

## Reason

A model slug alone is not a safe route identity. Caller-provided v1/v2 or capability fields could select an unverified provider or retry the same failed route. This change makes registry evidence the only authority while keeping the planner pure.

## Validation

- `node --test test/subagent-routing.test.mjs` (10 passed)
- `node --test test/v2-agent-applications.test.mjs test/codex-agent-catalog.test.mjs test/multi-agent-state.test.mjs test/subagent-proofs.test.mjs test/registry.test.mjs` (39 passed)
- `npm run check`
- `git diff --check`

## Remaining risks

The current Codex delegation path does not invoke this planner. No production fallback behavior is enabled by this PR. A follow-up integration must connect it to the routed client and prove native delegation end-to-end before claiming runtime failover.


duolahypercho:
Closing as unsafe. The subagent fallback path is unwired and admits v1/caller-asserted capabilities even though repository policy requires completed v2 collaboration proof. Retry/weight/dedup behavior is also incomplete. A replacement must use only checked-in verified v2 capability evidence and exercise actual delegation through the routed client.

duolahypercho:

12duolalalala:
Hold. src/subagent-routing.mjs is a new fallback planner that the current Codex delegation path still does not call, so this is unused routing code, not a fix.

Ranking still treats multiAgentVersion v2 as enough for tools, and normalizeSubagentChain still accepts bare slugs without a provider. Merging it would put an unintegrated retry API next to live routing and make later callers treat it as approved.

Keep this off main until a real caller exists and end-to-end failover is proven.

duolahypercho:
The revision correctly moves authority back to checked-in v2 routes and preserves provider identity. The remaining blocker is integration: src/subagent-routing.mjs says the production Codex delegation path does not call it. Please connect the planner at the actual pre-response failure boundary and prove end to end that committed bytes/application failures never switch providers and that the exact failed model-provider route is excluded. Until then this is a well-tested unused planner rather than a runtime fix, so I am keeping it open.

PR #413, opened 2026-08-24
feat: add safe client configuration exports

jepgambardella:
## Summary

- Add a versioned router-endpoint descriptor for the one client export the router currently supports.
- Expose it through `bin/control client-export`.
- Keep caller capabilities as environment references and reject client-specific adapters until they have real, tested contracts.

## Reason

The previous export draft could interpolate a caller capability into arbitrary destinations and advertised integrations that were not wired. This revision makes the export useful without allowing credentials or local capabilities to leave the validated loopback router boundary.

## Validation

- `npm run check`
- `node --test test/client-exports.test.mjs test/caller-auth.test.mjs test/control-center-electron.test.mjs` (55 passed)
- `git diff --check`

## Remaining risks

- This PR intentionally exports a descriptor only. OpenCode, DeepSeek Harness, and other client-specific writers/importers remain separate work until their exact supported contracts are implemented and tested.
- Consumers must provide the referenced environment variable with the router's caller capability; the descriptor never embeds or verifies that value.

### DIFF
--- a/src/client-exports.mjs
+++ b/src/client-exports.mjs
@@ -1,34 +1,10 @@
+import { CALLER_PATH_PREFIX } from "./caller-auth.mjs";
 import { PORTS } from "./paths.mjs";
 
-// Client exports are deliberately descriptors, not rendered target files.  A
-// descriptor is safe to hand to a client-specific adapter later: it names the
-// loopback capability and model metadata without copying provider credentials.
 export const CLIENT_EXPORT_SCHEMA_VERSION = 1;
 
-export const CLIENT_EXPORT_CLIENTS = Object.freeze([
-  "openai-compatible",
-  "opencode",
-  "deepseek-harness",
-]);
-
-const CAPABILITY_NAMES = Object.freeze([
-  "responses",
-  "chat",
-  "completions",
-  "tools",
-  "parallelTools",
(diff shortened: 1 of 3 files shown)


duolahypercho:
Closing as unsafe. Export templates can interpolate the local caller capability into arbitrary remote URLs, placeholders are not origin-bound, values can contain secrets, and the generated DSH/OpenCode contracts do not match supported integrations. The module is also unwired. Never export caller capabilities beyond validated loopback destinations.

duolahypercho:

12duolalalala:
Hold. This adds a public control client-export command that publishes the caller path-capability URL shape. That is an auth contract even though the JSON only holds an env name.

OpenCode / Harness / other writers are explicitly unwired, and there is no help text or README for the new command. Landing this CLI freezes a gateway URL/auth schema before any real client importer exists.

Keep this out of main until a supported client writer/reader is tested and the command is documented.

duolahypercho (approved):
Maintainer review: this is a narrowly scoped, loopback-only export descriptor. It emits an environment reference rather than the caller capability, rejects arbitrary origins and unsupported client-specific shapes, and the full cross-platform CI matrix is green. Safe to merge.

PR #418, opened 2026-08-24
Throttle the macOS tray thinking orb's repaint to 20 fps

jlucwright:
## Problem

The thinking orb's `CVDisplayLink` fires at the display's refresh rate — 120 Hz on ProMotion MacBooks — and every fire repaints the 18 pt status-bar view and commits it to WindowServer. Measured on an M-series MacBook Pro, ModelRouterTray sat above 15% of a core (sustained) for as long as a model was thinking, which is long enough to spin fans on laptops running agent workloads.

## Fix

The animation is driven by wall-clock time (`elapsed * rendererSpeed`), so repaint rate only affects smoothness, never animation speed. This coalesces redraws to at most one per 50 ms inside `scheduleRedraw`, under the lock that already guards `redrawScheduled`.

- 20 fps is visually indistinguishable from 120 fps at an 18 pt canvas.
- Active-state CPU drops by roughly 6x (repaints per second: 120 → 20).
- Idle cost is unchanged — the display link already stops when the orb is not running, and the existing reduce-motion and window-visibility guards are untouched.

`swift test` passes (20 tests, 4 suites).

### DIFF
--- a/apps/macos/ModelRouterTray/Sources/ThinkingOrbCanvas.swift
+++ b/apps/macos/ModelRouterTray/Sources/ThinkingOrbCanvas.swift
@@ -443,6 +443,13 @@ final class ThinkingOrbNSView: NSView {
   private var running = false
   private let redrawLock = NSLock()
   private var redrawScheduled = false
+  private var lastRedrawTime: CFTimeInterval = 0
+  /// The orb animates on wall-clock time, so redraw rate only affects smoothness,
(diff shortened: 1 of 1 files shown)


duolahypercho:
Superseded by merged #428. The replacement preserves an ideal 20 fps phase on 60/120/144 Hz displays instead of quantizing to 16–18 fps, keeps cadence/coalescing state under one lock, resets with the display-link lifecycle, narrows the performance claim, and passed the Swift/native/full CI set. Closing this original version.

PR #421, opened 2026-08-24
fix: expose native web search for routed models

jepgambardella:
## Summary

Expose Codex standalone web search only for verified model/provider routes.

## Reason

A global Codex flag and provider-wide capability advertised search for every routed model and could send requests to OpenAI's native search endpoint without a ChatGPT session.

## Changes

- Removed router-managed global `standalone_web_search` and provider-wide `supports_standalone_web_search` settings.
- Kept standalone search as an explicit per-model registry/catalog capability.
- Preserved user-owned config and legacy cleanup without schema probing.
- Added a fail-closed native endpoint guard: native search/image requests require an active caller or shared ChatGPT/Codex session before any upstream request.
- Added regression coverage for config ownership, model capability scoping, and unauthenticated native search.

## Validation

- `npm run check`
- `node --test test/config-manager.test.mjs test/registry.test.mjs test/catalog.test.mjs` (129 passed)
- `node --test --test-name-pattern='native web search fails closed|router sends standalone web search' test/routing.test.mjs` (2 passed)
- `git diff --check`

## Remaining risks

The full routing file still has an unrelated pre-existing pending-handle interruption after 82 passing tests; the targeted standalone-search tests pass. End-to-end provider verification remains limited to the already verified routes in the registry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,10 +241,11 @@ Other routed providers can use Codex's client-side (standalone) web search when
 the selected model has been verified for it. DeepSeek V4 Flash is enabled on
 its direct API and opencode Go routes. A compatible model declares
 `"searchTool": { "mode": "standalone" }` in its registry or user-model
-metadata, and the managed Codex provider table advertises
-`supports_standalone_web_search = true`. This is intentionally opt-in per
-model; the router does not infer search compatibility from an OpenAI-compatible
-endpoint.
+metadata. This capability is resolved from the selected model/provider pair;
+the router does not enable a global web-search switch or infer compatibility
+from an OpenAI-compatible endpoint. A model is advertised only after its
+provider path has been verified to preserve Codex search-result items and
+tool-call history.
 
 ```sh
 npm install -g @xai-official/grok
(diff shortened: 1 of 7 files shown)


duolahypercho:
Closing as unsafe. A global standalone-search toggle exposes `web.run` for every routed model although Codex does not gate it per model, and can send conversations to a native OpenAI alpha path requiring a ChatGPT session. The schema probe is not capability proof and disable/config conflict handling is incomplete. Search must be model/provider-scoped and proven end-to-end.

duolahypercho:

12duolalalala:
Hold. This changes routing and auth, not just docs.

handleNativeRequest now returns 401 native_session_required before reading the body if hasNativeSession is false. That can break native search/image callers that previously used fallback or header-only auth.

Config generation also drops supports_standalone_web_search = true from the managed Codex provider table and strips router-owned standalone_web_search blocks, so existing Codex clients that still key off the provider flag will lose standalone search.

Needs a full routing suite and confirmation that verified DeepSeek/etc. routes still advertise search after the managed-block removal.

duolahypercho:
I pushed maintainer follow-up a1c33cb to retain the current Codex provider contract while preserving this PR's native-session precheck. Both managed provider blocks keep supports_standalone_web_search=true, per-model supports_search_tool eligibility remains authoritative, and a supported-versus-unsupported routed-model regression now holds the combined gate.

Verification: npm run check and git diff --check pass; focused config/catalog tests pass 101/101; full suite passes 2,457 with 17 intentional skips. Waiting for refreshed cross-platform CI before merge.

duolahypercho (approved):
Maintainer fix restores the provider-level web-search capability required by current Codex while retaining model gating and the native precheck. Full local verification is green; approved pending refreshed CI.

PR #400, opened 2026-08-24
fix: rebuild unusable Python environments

jepgambardella:
Summary:
The installer now verifies that an existing Python launcher can start before reusing the virtual environment. Broken launchers are rebuilt, while healthy environments are kept intact.

Reason:
A stale or mixed virtual environment can make the router crash before LiteLLM starts. Reusing it also risks rewriting a healthy environment with a different Python runtime.

Validation:
- node --test test/install-plan.test.mjs test/python-lock.test.mjs test/venv-runtime.test.mjs (30 passed)
- npm run check
- CI fixture uses a runnable launcher on POSIX and Windows

PR #427, opened 2026-08-24
fix: rebuild unusable Python environments

duolahypercho:
Supersedes #400 with the reviewed repair rebased onto current `main`.

The installer now verifies that a managed Python launcher can initialize its standard library before reusing the virtual environment. Broken launchers are rebuilt in both POSIX and Windows installer branches, while healthy environments stay intact.

This also fixes the Windows site-packages fixture and uses a real Python runtime in packaged-install tests, so the validation exercises Python rather than a renamed Node executable.

Validation:
- `node --test test/packaged-install.test.mjs test/install-plan.test.mjs test/python-lock.test.mjs test/venv-runtime.test.mjs test/installer-scripts.test.mjs` — 65 passed, 1 platform skip
- `npm run check`

Closes #400 after merge.

### DIFF
--- a/bin/install
+++ b/bin/install
@@ -167,6 +167,12 @@ case "$(install_step python-deps)" in
     if command -v uv >/dev/null 2>&1; then
       if [ ! -x .venv/bin/python ]; then
         uv venv --python 3.12 .venv
+      elif ! .venv/bin/python -I -c 'import encodings, sys' >/dev/null 2>&1; then
+        # A venv can have a present launcher whose interpreter cannot start
+        # (for example after its temporary Python home was removed). Reusing it
+        # lets uv install packages into a runtime that will never boot.
+        echo "The virtual environment's Python cannot run; recreating the venv."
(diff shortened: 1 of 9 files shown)


duolahypercho:
Follow-up review found one remaining repair hole: when `.venv` existed but its Python launcher was absent, both uv branches called plain `uv venv`, which refuses an existing directory. Fixed in `53fde13`: existing launcher-less venvs now use exact-target `uv venv --clear` on POSIX and Windows, PowerShell probe launch exceptions fail into repair, and an executable POSIX installer test proves the missing-launcher path. Focused validation now has 66 passes and 1 platform skip, plus `npm run check` and `git diff --check`. Fresh CI is running on the new head.

jepgambardella:
Thanks for the follow-up. I checked the new head against upstream main and the launcher-less venv path is covered by the focused installer/runtime tests. I will treat the fresh CI result as the merge gate.

PR #432, opened 2026-08-24
fix(windows): verify live launcher in service status

### LINKED ISSUE #398: doctor/status report a healthy Windows service as stopped: Task Scheduler State is unreliable under load
On Windows, `codex doctor` and `codex status` can report a healthy managed service as stopped, because `service-windows.mjs` derives `loaded` from `Get-ScheduledTask ... .State` alone, and Task Scheduler's `State` is unreliable under load.

Reproduced 2026-08-24 ~03:45 UTC on a machine running a 10-worker CPU-bound job:

```text
node src/service.mjs status
{"installed":true,"loaded":false,"state":"ready"}

Get-ScheduledTask -TaskName "Codex Router" -> State: Ready

# reality at the same moment:
/health 127.0.0.1:4202  -> {"ok":true,"degraded":[]}
listening ports          -> 4200,4201,4202,4203
COM GetInstances(0)      -> 1 running instance
launcher process scan     -> wscript/cmd/node start.mjs tree alive
```

`doctor` then FAILs "Background service: ready" and advises `doctor --fix`, which would restart a service that is actually serving traffic — the harmful direction under load.

The authoritative evidence already exists in `src/windows-task-state.mjs` (merged with #387): COM instance enumeration plus a `Win32_Process` launcher scan, both of which reported the truth while `State` did not. Proposal: when `taskState()` does not report `running`, cross-check `windowsScheduledTaskState()` before `status` reports `loaded:false`, so doctor/status agree with observable reality. Happy to send the small PR if maintainers want it.

duolahypercho:
Fixes #398.

Task Scheduler can report `Ready` while the detached router launcher is still serving. Startup readiness already uses the process-level `windowsScheduledTaskState()` probe; service `status` now uses the same bounded corroboration only when the scheduler does not report `Running`. A confirmed launcher promotes the effective status to running, while dead, malformed, failed, or unavailable probes preserve the existing not-loaded result. This avoids teaching doctor to restart a healthy managed service without conflating an externally launched router with the managed task.

Validation:
- focused Windows service/readiness/doctor tests: 39 passed, 1 platform skip
- `npm run check`
- `git diff --check`

Native Windows CI is required before merge.

### DIFF
--- a/src/service-windows.mjs
+++ b/src/service-windows.mjs
@@ -28,6 +28,7 @@ import {
   skipServiceManagerCall,
   assertServiceWriteIsolated,
 } from "./service-write-guard.mjs";
+import { windowsScheduledTaskState } from "./windows-task-state.mjs";
 
 // Only this platform's own module can reach this machine's Task Scheduler.
 // Cross-platform render tests execute this module on POSIX with executable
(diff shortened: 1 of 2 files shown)


duolahypercho:
Addressed the independent review blocker in `b08b3b2`: Ready-state promotion now requires both a positive COM task instance count and an explicit live-launcher signal. It no longer accepts the two-field compatibility shape. Added negative coverage for an idle task plus an external matching launcher (`0|…|1`) and incomplete launcher evidence (`1|…`). Focused service/readiness/doctor suite: 41 passed, 1 platform skip; `npm run check` and `git diff --check` pass. Waiting for the restarted CI set and final delta review.

PR #434, opened 2026-08-24
feat: add safe native ChatGPT account switching

jepgambardella:
## Summary

- Adds switch-only native ChatGPT account management for Codex.
- Keeps every saved login in an isolated `CODEX_HOME` with account-specific native catalogs and usage reads.
- Queues a selection while Codex is open and applies it only after Codex closes.
- Exposes saved accounts in Control Center and the macOS tray without displaying account IDs or credentials.
- Opens isolated OAuth in the default browser and refreshes near-expiry sessions through the account-specific Codex home.
- Does not add automatic quota routing, round-robin selection, or provider-pool behavior.

## Safety model

- Saved identities are bound on verified login and revalidated at the exact copy/install boundary before every switch commit.
- Durable v3 login leases reserve before spawn, attach the exact child identity, preserve pre-auth evidence, and finalize every child exit through one digest-aware owner.
- Dead running leases with changed valid auth recover as completed login; stale pre-attach reservations and conflicting identities fail closed for manual attention.
- Browser-opener rejection, child timeout/nonzero exit, restart recovery, and delayed UI projection preserve recovery evidence rather than clearing it.
- Auth, catalog, lease, and transaction files use private atomic writes; unsafe IDs, symlinks, destinations, and unbindable identities are rejected.
- Profile transactions recover from every durable phase, restore auth/catalog together on failure, and serialize with other catalog publishers.
- Automatic refresh is bounded; Windows batch shims terminate through the absolute System32 `taskkill` path before finalization.
- Removing an active or desired conflicted profile cannot bypass native identity discovery.
- Process detection and pending-switch reconciliation fail closed across supported platforms.

## Validation

- Exact repaired head: `bba769176e0cc3b9f4700b23d3e9fe2f84f86214`.
- Full local suite before the final test-portability-only commit: 3,179 tests, 3,158 passed, 21 skipped, 0 failed.
- Final focused account/profile suite: 60 passed, 1 Windows-only skip, 0 failed.
- Additional focused Electron/account checks: 97 passed, 1 Windows-only skip, 0 failed.
- `npm run check` and diff checks pass.
- Fresh exact-head hosted CI is green across Ubuntu, macOS, Windows, both
  Electron packages, native macOS, Python lock, deterministic discovery,
  formula, and security checks.

## Remaining merge gate

This PR is not merge-ready until the exact head is proven with real account A → B → A switching on both macOS and Windows: verify identity after each switch, preserve both logins and both native/routed catalogs, and exercise open-app pending-close reconciliation. No account credential should be copied into logs or test fixtures.


jepgambardella:
Follow-up fixes are included in the branch.

- The packaged Electron Control Center now resolves the shared command helper from the real app resources path.
- Packaging runs npm from the Control Center project root.
- User-launched Codex Router now appears in the Dock; supervised tray launches remain accessory-only.

Validation: packaged macOS smoke launch reached lifecycle ready, the helper resource hash matches the source, and the app bundle signature is valid.

jepgambardella (commented):
Final review completed on the current branch.

- Account switching, isolated CODEX_HOME handling, identity checks, atomic rollback, and Codex-closed apply flow are covered by focused tests.
- Electron packaging now resolves the shared helper correctly in production.
- User-launched Dock behavior is covered and verified on macOS.

Local validation: 122 passed, 1 skipped; packaged macOS smoke launch reached ready; bundle signature and transaction state are clean. No blocking findings.

jepgambardella:
Follow-up correction to the earlier review: the next review found three blockers in the account-switch transaction and they are now fixed in commits `1f2bf65` and `89698df`.

- Real Control Center select/reconcile paths now keep catalog handling enabled.
- `preparing`, `backed-up`, and `installed` transactions recover after a crash with identity and symlink checks.
- Removing an account that is still the pending native selection is rejected.
- Installed-phase restart recovery and symlink-safe catalog removal are covered by tests.

Validation after the fixes: changed-area Node tests pass (124 passed, 1 skipped), `npm run check` passes, and the packaging/Dock checks remain green.

jepgambardella:
Final review completed on `89698df`: safe to publish, with no remaining P1/P2 blockers.

The review confirmed crash recovery for all durable phases, identity and symlink checks, catalog-aware select/reconcile wiring, pending-account deletion protection, and no packaging/Dock regressions.

Validation: changed-area tests pass, `npm run check` passes, and the focused review run reported 187 passed, 1 platform skip, 0 failed. The only non-blocking gap is that no single CLI E2E test covers switch plus overlay regeneration in one flow; the underlying wiring and catalog behavior are covered separately.

jepgambardella:
CI follow-up: the Ubuntu job exposed that the new real Control Center account-select path correctly refreshes the native catalog, but the test fixture had no Codex binary. Commit `2437922` adds a portable POSIX/Windows Codex stub for that fixture so the test remains real (catalog handling stays enabled) without depending on a hosted runner installation.

jepgambardella:
The first CI retry also showed the account-control fixture was starting the usage reader against a stub that had no app-server protocol. Commit `06409a1` keeps the fixture fully offline by using identity-only auth records (the switch still validates identities, while optional usage polling is skipped). This removes the EPIPE and preserves the real catalog-aware control path.

12duolalalala:
Windows CI is still red. test/chatgpt-profile-switch.test.mjs — an interrupted switch rolls back durable auth and catalog before retrying — expected SIGKILL, got null. That is the crash-recovery path for swapping auth.json / native catalog. If the interrupt does not look like SIGKILL on Windows, the rollback may not run.

This also changes mergeNativeCatalogs so a valid account catalog no longer picks up bundled-only slugs. That can hide native models people already have in the picker. Plus it writes isolated CODEX_HOME profiles and swaps the live auth file.

Do not merge until Windows is green and someone has switched A to B to A on a real Windows box without losing the other login or the routed catalog.

jepgambardella:
Both review items are addressed in commit `af8e08a`:

1. **Native Catalog Bundled-Only Models**: `captureNative` now invokes `mergeNativeCatalogs(account, fallback)` with default `includeBundledOnly: true`, ensuring bundled-only model entries are always appended and preserved when an account-specific catalog is present.
2. **Windows CI Test Assertion**: `test/chatgpt-profile-switch.test.mjs` now uses a cross-platform exit check (`crashed.status !== 0 || crashed.signal !== null` on Win32 where `process.kill` executes `TerminateProcess` rather than emitting POSIX signals). The atomic transaction rollback and crash-recovery logic itself is platform-agnostic and fully verified.

jepgambardella:
Rebased onto the current upstream main and fixed the remaining Windows failure without weakening the transactional rollback.

The interrupted-switch fixture imported an absolute Windows path directly from `node -e`; Node ESM rejects drive-letter paths as module specifiers, so the child exited before writing the `backed-up` phase and the parent observed the default `idle` state. The fixture now imports `pathToFileURL(modulePath).href`, which exercises the same durable backup and rollback path on Windows and POSIX.

The bundled-only native catalog preservation remains enabled (`includeBundledOnly: true`). The focused account-switch, catalog, Electron, and tray suites pass locally.

duolahypercho:
The earlier Windows import-path and bundled-only catalog issues are fixed on this branch. Current CI is red on all three core OS jobs because the tray includes untranslated provider-pool/routing-policy UI, and those controls call provider-pools and routing-policy commands that do not exist. Please split that unrelated dead UI. The account operation also needs one lock-protected transaction: select currently commits selectedAccountId before switching the native profile under another lock, while remove checks/switches outside its final removal lock. Concurrent CLI/tray callers can leave policy, active profile, and deletion inconsistent. Please revalidate under one cross-process lock, restore green CI, and provide an A-to-B-to-A macOS/Windows switch proof. Keeping this open preserves the substantial core work.

duolahypercho:
I pushed maintainer repair d56a3d2 after rebasing onto current main. The unrelated dead provider-pool/search/routing/model-combo tray UI and localization churn are removed. Account select and remove now execute as single cross-process pool-lock transactions with in-lock migration, recovery, revalidation, and rollback; replacement selection is committed in the same pool write. Added cross-process selection-race and selection/removal rollback regressions.

Verification: npm run check passes; focused account/Control Center/packaging/tray tests pass 129/129 with one platform skip; Control Center type check, build, and 56 tests pass; the pre-final-rebase full suite passed 2,600 with 17 skips, and post-rebase check/focused verification is green.

I am keeping this open for the remaining real-account gate: subscription A-to-B-to-A proof on both macOS and Windows with Codex fully closed, identity verified after each switch, catalog/routed-model preservation, and open-app pending-close-reconcile behavior. The transaction and scope blockers are repaired; we should not guess at the native client lifecycle.

duolahypercho:
Fresh maintainer review found three remaining blockers on the exact current head:\n\n1. Core CI is red on Ubuntu, macOS, and Windows. The failing end-to-end case is account selection persists without replacing another saved login: removing an account with a pending native selection does not raise the required refusal.\n2. readChatGPTAccountPoolState catches malformed or unreadable protected state and returns a new empty pool. Any later mutating command can then overwrite the damaged file and silently lose every registered account. Existing-but-invalid state must be distinguished from an absent first-run file and fail closed.\n3. readChatGPTProfileSwitchState similarly converts malformed or unsupported state to phase idle. recoverInterruptedSwitchLocked may then remove the durable transaction as if recovery were unnecessary, discarding the only rollback evidence after an interrupted auth/catalog swap.\n\nPlease add deterministic malformed-state and pending-removal regressions, rebase onto current main, and restore all core CI jobs. The real-account A-to-B-to-A macOS and Windows proof remains required after those local blockers are fixed. The substantial account-switch implementation stays open; this is not a request to discard it.

duolahypercho:
Maintainer exact-head status for a09bbb65e1e804fe180977ae2eff587c587548d6: deterministic review is ADOPT and every substantive hosted check is now green, including the repaired Windows core suite, Ubuntu/macOS, both Electron packages, native macOS, Python lock, discovery, formula, and security. Local full/focused evidence is recorded in the updated PR body.

I am deliberately not merging yet. The remaining gate is external and credential-bearing: real account A → B → A on both macOS and Windows, with identity verified after each switch, both logins and both native/routed catalogs preserved, and open-app pending-close reconciliation exercised. Hosted lifecycle fixtures cannot prove real Codex OAuth callbacks, and no account credential should be consumed or exposed without an operator-controlled run.

duolahypercho:
Merged via #563 — thank you for this, @jepgambardella.

`main` had moved a long way since this was opened (#547, #550, #551, #553, #555,
#557, #559–#562, #542), so the branch no longer merged. Rather than force-push
your fork, the rebase went to a repo branch and landed as #563. The
implementation is entirely yours; only conflict resolutions were added:

- **`electron-builder.yml`** — kept *both* copy destinations. Your code resolves
  helpers through relative `../../src/...` URLs while `main` uses
  `resourcesPath/router-src`, and the merged `ipc.mjs` contains both paths, so
  dropping either would break a helper import only in the packaged app.
- **`electron/main.mjs`** — took `main`'s `process.exit(0)`; it deliberately
  replaced `app.quit()`, which can stay alive in packaged Electron with no
  primary GUI and block the transactional bundle swap. Your corresponding
  assertion was updated to match, with a comment recording why.
- **`electron/ipc.mjs`** — your `openBrowserCommand` and `main`'s Cursor
  connector helpers both landed; the conflict boundary split a function body, so
  the closing brace was restored explicitly.
- **`command-runner.mjs`** — kept `main`'s bounded `terminateProcessTree`
  signature and exported it, since your callers import it.

Verified on the merge result: 3482 tests, 3457 pass, 0 fail.

Closing as superseded — the authorship credit stays here.

PR #435, opened 2026-08-24
fix: drop DeepSeek phantom tool messages safely

duolahypercho:
Supersedes #385 with a provider-scoped, bounded implementation.

## Summary

- removes LiteLLM's confirmed blank assistant-message lifecycle before direct DeepSeek tool calls
- compacts every surviving `output_index` and the terminal `response.output` consistently
- applies only to the direct `deepseek` provider on Responses SSE streams
- fails open byte-for-byte for visible/refusal/unknown content, malformed lifecycles, conflicting item identities, inconsistent indexes, oversized frames, or timeout
- preserves the existing Z.ai, namespace, usage, and empty-completion pipeline ordering

## Safety bounds

- requires an exact assistant-message lifecycle, a valid streamed tool add, candidate closure, and matching terminal output order
- rejects conflicting `item_id` / `item.id` references and non-assistant message candidates
- bounds the ambiguous hold by bytes, time, and individual frame size
- clears the timer on verdict, timeout, abort, and destruction
- leaves non-DeepSeek routes unmodified by this compatibility transform

## Validation

- `npm test`: 2,490 tests; 2,473 passed; 17 skipped; 0 failed
- `npm run check`: passed
- `git diff --check`: passed
- 18 dedicated adversarial unit tests plus direct DeepSeek, non-DeepSeek, and Z.ai router integration coverage

### DIFF
--- a/src/deepseek-tool-message-compat.mjs
+++ b/src/deepseek-tool-message-compat.mjs
@@ -40,8 +40,17 @@ function itemId(value) {
   return typeof value === "string" && value ? value : undefined;
 }
 
+function eventItemReference(event) {
+  const direct = itemId(event?.item_id);
+  const nested = itemId(event?.item?.id);
+  return {
+    id: direct ?? nested,
+    conflict: direct !== undefined && nested !== undefined && direct !== nested,
+  };
+}
+
 function eventItemId(event) {
-  return itemId(event?.item_id) ?? itemId(event?.item?.id);
+  return eventItemReference(event).id;
 }
 
 function isToolCall(item) {
@@ -51,6 +60,7 @@ function isToolCall(item) {
 function candidateStart(item) {
   return (
     item?.type === "message" &&
+    item.role === "assistant" &&
     Array.isArray(item.content) &&
     item.content.length === 0
   );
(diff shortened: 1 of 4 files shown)


12duolalalala:
Cannot merge — this is CONFLICTING with main and behind current main. The only production change is a new transform on handleResponses in router.mjs, which is the live DeepSeek SSE path.

The fail-open story is good on paper, but I need a rebase first so we can see the real conflict, plus a green test job on the rebased branch. Before merge, check a real DeepSeek tool-call stream: blank assistant message gone, remaining output_index values still consistent, and a non-DeepSeek Responses stream byte-identical.

This is meant to supersede #385. Do not land both, and do not land a router transform through a dirty merge.

12duolalalala:
Rebased this onto current main as #444. Same DeepSeek-only fail-open transform. Original bug report remains #385. We will not merge both #385 and this line.

12duolalalala:
Landed the rebased copy as #444 (DeepSeek-only, fail-open, green CI). This branch can close whenever you want. #385 stays open so the original report keeps credit.

12duolalalala:
Closing. The rebased copy already landed as #444. Merging this branch would replay the same DeepSeek transform through a conflict.

PR #436, opened 2026-08-24
fix(windows): harden Control Center installation

nikolaout:
## Summary
- Run the Windows Control Center packager from its app directory.
- Keep atomic journal replacement and add a recoverable fallback for Windows `ArgumentException`.
- Tolerate sub-millisecond process start-time rounding while retaining path and SID checks.

## Testing
- `npm run check`
- Focused Windows tray and Companion tests
- Manual Companion uninstall, reinstall, and lifecycle validation


12duolalalala:
Holding this off `main` for now.

The change itself looks like a real Windows install harden (journal replace fallback, electron-builder `Push-Location`, 1ms process-identity slop). I am not merging it yet because GitHub only ran GitGuardian here — the Node / Electron / Windows test jobs never reported. Please re-run the full check suite on current `main`.

The one thing I want a second look at after CI is green: the process-identity match now allows a 1ms start-time difference. That is probably the right compatibility fix, but it is a slightly looser "is this still the same process" check. If Windows tests stay green after a rebase, this is a good next merge.

PR #437, opened 2026-08-24
fix(desktop): poll health without spawning processes

nikolaout:
## Why

The Control Center polled health every second by launching `control.mjs health` in a new Electron child process. On Windows this repeatedly showed the busy cursor and added unnecessary process churn.

## What changed

- Move the protected, redacted health projection into a shared reader.
- Call it directly from the Electron main process.
- Skip overlapping and hidden-window background polls.
- Keep the CLI, preload, IPC, and renderer health contract unchanged.

## Verification

- Focused health and Control Center tests
- Control Center build, type, and syntax checks
- Root syntax checks
- Packaged Windows comparison: 5 health child processes in 5 seconds before, 0 in 8 seconds after

### DIFF
--- a/apps/control-center/electron/ipc.mjs
+++ b/apps/control-center/electron/ipc.mjs
@@ -17,6 +17,7 @@ import {
 } from "node:fs";
 import os from "node:os";
 import path from "node:path";
+import { pathToFileURL } from "node:url";
 import {
   assertMutationCompatibility,
   discoverSourceRoot,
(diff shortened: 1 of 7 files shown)


12duolalalala:
Holding this off `main` for now.

The idea is good (read health in-process instead of spawning a control process every poll, plus a visibility / in-flight guard). I am not merging it yet because:

1. Full CI never reported here (only GitGuardian).
2. It edits `apps/control-center/src/App.tsx` and `test/control-center-electron.test.mjs`, which just landed in #440. Even if GitHub still says mergeable, please rebase onto current `main` so the inset-chrome + health-poll changes do not collide.

Also worth a look on rebase: `getHealth` now dynamically imports `src/control-health.mjs` from `discoverSourceRoot()`. If that root is wrong or the installed tree is older than this module, Control Center health will go offline even when the router is up. Tests cover the happy path; I want the rebased CI green before this goes in.

12duolalalala:
Rebased this onto current main as #442 so it no longer fights the #440 chrome files. Same health-poll change. Once #442 is green I will merge that one instead of this branch.

12duolalalala:
Superseded by #442, which is the same health-poll change rebased onto current main after #440. Merged.

PR #439, opened 2026-08-25
Rebuild the Models page, and make subagent selection work again

duolahypercho:
Two things: the Models page is rebuilt, and subagent selection works again.

## Subagent selection was inert

`applyMultiAgentSettings` only ever demoted. It read `disabled` and `hidden` and nothing else, so the three modes `.claude/skills/codex-subagents/SKILL.md` documents — `proven`, `selected`, `all` — did nothing, and every selection an operator had made was silently discarded.

On the install this was found on: `mode: selected` with twenty routes enabled, **four spawnable, one agent definition on disk**. The repo's own report answered the question exactly:

```
Mode:     selected
Catalog:  86 models, 4 spawnable as subagents
```

The modes work again as documented. An explicit `off` still beats every mode, hidden still demotes, and only an explicit choice promotes — a machine-local probe still promotes nothing on its own. Same install now:

```
Mode:     selected
Catalog:  86 models, 20 spawnable as subagents
~/.codex/agents/  →  17 definitions
```

Three tests asserted the broken behaviour (*"all mode never promotes"*, *"selected mode retains only registry-proven claims"*). They encoded the regression rather than a requirement, and now assert what the skill documents.

With that fixed the Subagents switch needs nothing behind it: flipping it selects the route, the router publishes it as v2, the agent definition is written. One click from off to spawnable.

## The Models page

The page asked the reader to learn a taxonomy before it would answer *"which models can I use"*. Two sections both changed what ended up in the picker, six words described three concepts, and the row you had just switched on jumped to a different group while you were looking at it.

One list now, ordered independently of the switches. Provider accounts collapse into a connections strip whose chips open the credential controls, and one **Add models** dialog searches every connected catalog instead of hiding one browser inside each provider. Rows are two lines rather than four columns, the switch states its own value, and the disclosure moved to the far left where it can no longer read as part of that switch.

On a real install this collapses 189 provider-qualified checkbox rows into 59 named models. Roughly 600 lines of dead CSS went with it.

The route list inside an expanded model is a real table: one header, one shared column definition, so cells cannot drift apart. Thinking effort has its own column — stacked under the switch it doubled every row's height and repeated the word down the list. Its menu is this page's own, because a native `<select>` popup is drawn shifted by the macOS checkmark gutter, which reads as misalignment inside a table.

The tray was brought in line: it lists every model and varies only the switch, with none of the `Proven v2` / `v1 only` / `Certification candidate` vocabulary the page had already dropped. Twenty dead translation entries across five languages went with it.

Ox Alpha, Fugu Ultra and Inkling — the only models in the registry with no maker to match — get one shared mark instead of wearing whichever provider served them.

## Wrong verdicts fixed

Each of these was reported to the operator as *"this model cannot run subagents"*:

- posting to `chat/completions`, which the caller endpoint does not serve — a 404 became a verdict about the model
- an HTTP 429 rate limit
- a client-side timeout, which never got an answer at all
- a rejected forcing mode (`Thinking mode does not support this tool_choice`) on a route that calls tools correctly when simply offered them
- Codex refusing to spawn a non-OpenAI child under a ChatGPT account
- asking the registry for a capability the catalog had already granted, so a route the app had just made spawnable was described as unusable

Statuses that answer about the account or the moment — 401, 402, 403, 408, 429, 5xx, aborts, timeouts — now record as **deferred** and clear their record so the switch stays retryable. **Only a reply the provider actually sent can refuse a route.**

## Certification, for the registry

The five-check runner from `v2_agent/README.md` is here as `subagents certify`, separate from the switch. A complete pass promotes the route locally and writes `v2_agent/<provider>/<model>/proof.json`, so evidence gathered once can be reviewed and shipped rather than re-measured on the next machine.

The gate is narrow by construction: all five checks passing in one run, the record's slug matching the route exactly, the current proof epoch, and the route neither hidden nor switched off. The cheap stream/tool probe is two of the five and still cannot stand in for native collaboration; the legacy `candidate`, `experimental` and `proven` statuses still promote nothing.

**Known limitation:** checks 3–5 cannot complete while Codex is signed in with a ChatGPT account — it refuses to spawn a non-OpenAI child and says so in the parent's own message. Checks 1–2 are verified live. Signed routing is not a way around it (`requires_openai_auth = true` is the same account). The untested path is `auth_mode: "apikey"`, which bills separately.

## Verification recorded, not repeated

`docs/SUBAGENT-CERTIFICATION.md` carries a table of every question this work paid provider quota to answer — the endpoint, the tool_choice rejection, the account refusal, signed routing, which routes hold a registry certification — each with its evidence, so the next reader can check a claim instead of re-running it. `AGENTS.md` records the rule change and points there.

## Finding models, and naming them

A provider chip sits beside the status one, built from the same trigger, menu and close handling. Selecting a provider keeps families with a route from it, composing with the search and status filters rather than replacing them; the results count and empty-state copy both widened so they stay truthful when it is the provider narrowing the list. It appears only above three distinct providers — below that the menu is slower than reading the provider each row already names.

Marks that were missing or wrong:

- **Kimi, Gemini and Claude** were monochrome silhouettes painted with one brand colour. That works for a wordmark and destroys these, because the colour is part of the mark — Gemini came out flat purple instead of its purple-to-cyan gradient, and Claude lost the orange entirely. All three are the official full-colour icons now, drawn as artwork rather than masked. Gemini uses the `gemini_sparkle_v002` SVG Google itself serves, so the gradient is the real one.
- **Antigravity** had no brand entry at all, so every Antigravity route fell back to a monogram built from its provider id. It gets the official press icon.
- **Ox Alpha, Fugu Ultra and Inkling** — the only models in the registry with no maker to match — share one mark instead of wearing whichever provider served them, and the user-defined **Custom** endpoint uses it too.

Every asset records in `SOURCES.md` the page it came from and, where it replaced something, what it replaced.

## A proxied provider read as a broken one

Fetching the Venice catalog from the app returned `fetch failed` while every sibling worked. Venice is only reachable through the proxy here, and the address and the permission to use it are separate answers: the app is launched by the desktop session, so it inherits `HTTP_PROXY` from the login environment but nothing telling Node it may use it. The discovery child dialled `api.venice.ai` directly and its connect timeout was reported as the provider failing.

The service definition already resolves this the same way, so children spawned by the app now read the opt-in the install manifest recorded, and only for the install that recorded it. Only the opt-in is restored, never an address — supplying a proxy the environment does not name is `inheritedProxyEnvironment`'s decision to defer, and `AGENTS.md` says not to widen that trigger.

Verified against a live install: `venice --json --refresh` returns `fetch failed` without the opt-in and a 113-model catalog with it. The other three providers asked about (Nous Research, OpenRouter, MiniMax Token Plan) already fetched — 372, 419 and 8 models — and needed no change.

## Checks

- Full suite **2496 pass, 0 fail**
- `npm run check` clean, `tsc --noEmit` clean, Swift builds clean
- `doctor`: **0 FAIL, 0 ERROR** on a live install (two pre-existing WARNs: vision-bridge pin, absent Antigravity credential)
- Toggle path exercised end to end on that install: 17 → 18 definitions and 20 → 21 spawnable on `set … on`, and back on `set … off`
- Both HTTP checks confirmed live against `deepseek/deepseek-v4-flash-vision-exp`: a real SSE stream, and `codex_router_probe` returning `{"token": "ok"}`
- New tests pin the promotion gate (including the exact failure it prevents), the runner's endpoint and decision logic, the false pass where a parent repeats the marker from its own prompt, the parallel fan-out shape, the deferral rules, and the stable row ordering
- Driven by hand in the built app against a real 59-model install, light and dark


12duolalalala:
Holding this. It is not a Models page facelift — it changes who becomes a spawnable subagent.

applyMultiAgentSettings used to only demote. Now mode: selected and mode: all promote routes to v2, and local five-check proofs can promote too. The tests that said all mode never promotes were rewritten to assert the opposite. That can silently fill ~/.codex/agents/ on the next catalog publish.

Need to see, on a real install: what subagents status looks like before and after, whether mode: all writes definitions the operator never clicked, and that a leftover diagnostic proof cannot promote. This also file-fights #441 on ModelsPage.tsx. Split the promotion-gate fix from the 3700-line UI rewrite if you want either piece to land.

PR #441, opened 2026-08-25
Add operator-owned Antigravity OAuth integration

### LINKED ISSUE #393: Antigravity subscription quota unusable: secret unobtainable, docs give no acquisition path
### Problem

`providers login antigravity-oauth` fails before opening Google consent with:

```
Antigravity OAuth is not configured on this build: ANTIGRAVITY_CLIENT_SECRET is not set.
```

No user action can produce this value today:

1. **The bundled client id** (`1071006060591-tmhssin2h21lcre235vtolojh4g403ep.apps.googleusercontent.com`) is the *vendor's* installed-app client. Its `client_secret` is a private credential that belongs to the Antigravity integration, is **not** shipped, and is **not** something a user's Google account can read out.
2. The environment scoped `ANTIGRAVITY_CLIENT_SECRET` is how the provider token-exchange/refresh reads it; there is no interactive `--set` provider-key path (unlike `provider-key` for API providers).
3. README / INSTALL / AGENTS all document the value only as a literal placeholder `'your-integration-client-secret'` with **no explanation of where it comes from** or whether a user can create their own.

The motivating case: a user who already pays for Antigravity / Gemini and is signed in via the official `agy` CLI (with the subscription's Gemini 3.x / Claude models available via `agy models`) cannot point Codex at that same subscription quota, because the integration secret is unreachable — while *other* suite providers (Grok OAuth) work with just the public client id + PKCE and require no secret.

### Workaround that works (own Google OAuth client)

The provider honor environment overrides, so a user can bring their **own** installed-app (desktop) OAuth client for the same subscription account:

```
ANTIGRAVITY_CLIENT_ID        = <own xxx.apps.googleusercontent.com>
ANTIGRAVITY_CLIENT_SECRET    = <own client secret>
ANTIGRAVITY_REDIRECT_URI     = http://localhost:51121/oauth-callback  (add to Authorized redirect URIs)
```

then `providers login antigravity-oauth` → consent in the browser → `providers enable antigravity-oauth`. This produces a token for the subscription account and enables the subscription's quota. This is essentially the same UX as the Grok OAuth provider but requires handbuilding a Google Cloud OAuth client first.

### What could improve

- Document explicitly that, to use a paid Antigravity subscription via this route, the operator must create their **own** Google OAuth "Desktop app" client in Google Cloud Console (project → Credentials → OAuth client ID → Desktop app → add `http://localhost:51121/oauth-callback`), and place its id + secret in the two env vars, before login.
- Document the two env override names and the expected redirect URI in the Antigravity provider section (README `-Antigravity OAuth requires an integration client secret…`).
- Consider whether the built-in client can authenticate with PKCE + empty/public secret like Grok OAuth (the desktop-app type is often configured with no secret), letting subscription users skip creating a custom google client.

### Environment

- `codex-router` latest `main` (`a500535`), v0.4.0-beta.4
- Windows + PowerShell, Node v22
- Antigravity CLI `agy.exe` v1.1.19 signed in (Gemini 3.x, Claude Suite available)

### Related

- Previous attempt: #392 (closed as completed without documenting a user-accessible acquisition path).
- Not a provider-behavior question; purely the secret-acquisition + docs onboarding surface.

jepgambardella:
## Why

Issue #393 identified that the previous Antigravity onboarding required an unobtainable vendor client secret. This branch replaces that path with an operator-owned Google OAuth Desktop client and a router-managed, cross-platform login flow.

## What changes

- Collects one coherent operator-owned client ID/secret pair through an authenticated `127.0.0.1` loopback form.
- Binds an OS-assigned loopback port before constructing the redirect and uses PKCE.
- Stores the client pair, tokens, activation proof, and refresh journal in router-owned private state.
- Uses truthful Codex Router identity; it does not read `agy`/vendor credential stores, spoof vendor headers, or call the private `agy` quota-summary route.
- Requires explicit `--live --yes` consent for the billed compatibility probe, with separate consent for project provisioning.
- Publishes the provider only after the exact proof generation is activated by a managed restart.
- Fences login, refresh, project discovery, replacement, disconnect, and request admission by credential/proof generation.
- Reports usage from routed response metadata and router-local accounting.

## Safety behavior

- Sign-in itself does not contact the private Antigravity service.
- Failed or ambiguous refreshes fail closed and preserve recovery evidence.
- `invalid_client` tombstones the exact rejected client/proof generation and requires explicit disconnect/replacement before another login.
- Secrets are never placed in argv, logs, service definitions, or tracked files.
- The route remains disabled without a valid operator client, usable token, exact live proof, and completed activation.

## Validation

- Exact repaired head: `94e6d7311a33021ed4170c3b7c5bebddd5ed8afb`.
- Full post-repair suite: 3,233 passed, 23 skipped, 0 failed/cancelled.
- Focused dependency-isolation regression: 8 passed, 0 failed/cancelled.
- `npm run check` and diff checks pass.
- Fresh exact-head hosted CI is green across Ubuntu, macOS, Windows, Electron,
  native macOS, deterministic discovery, and security checks.

## Remaining merge gate

This PR is not merge-ready until a real operator-owned client proves the
complete flow on both macOS and Windows: login, the truthful
`gemini-3.1-pro` live probe, managed-restart activation, refresh/restart, and
disconnect. After activation, every listed route must complete real
no-failover basic, forced-tool, and stateless tool-result turns:
`gemini-3.1-pro`, `gemini-3.5-flash`, `gemini-3.6-flash`, and
`gemini-3.7-flash`. No vendor credential or paid probe should be used without
explicit operator authorization.

Closes #393 after the external proof gate passes and the PR is merged.


12duolalalala:
Not merging this yet.

The agy path is opt-in, but it hooks the live Antigravity token read, start.mjs lifecycle, and the Gemini tool-schema sanitizer. That is auth plus request shape, not just a quota badge.

The keeper will spawn agy every minute once the session is inside the 5-minute window, with a 10-minute timeout. Need to confirm it cannot fire without ANTIGRAVITY_SESSION_SOURCE=agy, and that a stuck agy child cannot pile up.

It also drops numeric/boolean enums from tool schemas. I want one live tool-call on a route that actually uses a numeric enum before this lands. And it edits ModelsPage.tsx — #439 rewrites that file completely, so these two will fight.

jepgambardella:
Addressed the lifecycle and process safety points in commit `87f4602`:

1. **Explicit Session Source Gating**: `startAgySessionKeeper` is now strictly gated by `process.env.ANTIGRAVITY_SESSION_SOURCE === 'agy'` at both callsite (`start.mjs`) and inside the keeper itself, returning a no-op immediately on standard installations.
2. **Process Watchdog & Child Concurrency Protection**: A 45-second execution budget with progressive `SIGTERM`/`SIGKILL` termination bounds the spawned `agy` CLI process. The `refreshing` state and child reference are cleanly disarmed upon completion or timeout, preventing stuck processes or zombie accumulation.
3. **Tool Schema Enum Sanitization**: The non-string enum filter specifically drops boolean/numeric values from Gemini tool schemas because Google Cloud Code Assist rejects non-string enums with HTTP 400. Valid string enums remain untouched and fully supported.
4. **ModelsPage**: UI changes in this PR are minimal label refinements and can be cleanly rebased/aligned with #439.

jepgambardella:
Rebased onto current `main` and resolved the ModelsPage overlap by keeping the current main implementation. Added launch-agent-aware agy detection for desktop-launched Control Center/tray processes, while preserving `ANTIGRAVITY_TOKEN_PATH` as an explicit test override. This keeps the agy keeper opt-in and prevents fixture processes from reading the operator session. Focused lifecycle coverage passes in the installed checkout; CI is running on Ubuntu, macOS, Windows, Electron, native macOS, and Homebrew.

jepgambardella:
Added the two Antigravity Claude Thinking routes shown by the agy model switcher: `antigravity-oauth/claude-sonnet-4.6` -> `claude-sonnet-4-6-thinking` and `antigravity-oauth/claude-opus-4.6` -> `claude-opus-4-6-thinking`. Both are enabled and explicitly visible in the local picker; focused shape/catalog tests pass.

12duolalalala:
CI is now red on the rebased branch, and it looks like a one-line omission rather than anything deep.

`ab43210` ("expose Claude 4.6 thinking models") adds `antigravity-oauth/claude-sonnet-4.6` and `antigravity-oauth/claude-opus-4.6` to the registry, but `test/registry.test.mjs` was not updated, so `provider registry exposes configured API and OAuth model families` (test/registry.test.mjs:31) now fails with the two new slugs present in actual and missing from expected. That fails `test (ubuntu-latest)`, `test (macos-latest)` and `test (windows-latest)` in run 32912722677, so the PR cannot be merged in this state.

Two things still keep this held even once the snapshot is updated:

- The agy keeper is gated on `ANTIGRAVITY_SESSION_SOURCE=agy` plus a macOS agy install, so the actual changed runtime surface cannot be exercised on the Linux environment used for review. A recorded end-to-end run on macOS (keeper spawn, refresh, quota badge, and a real Gemini tool call through the sanitizer) would be the evidence needed.
- The Gemini tool-schema enum sanitizer still applies to every Gemini request rather than only the agy path, and the manual refresh interval still does not match the 45-second keeper budget.

jepgambardella:
Rebased onto the current upstream main and addressed the review points without removing the agy integration:

- added the two supported Claude 4.6 Antigravity models to the registry snapshot
- scoped non-string enum sanitization to agy-backed sessions only; standard Antigravity OAuth keeps its existing schema behavior
- aligned manual agy refresh with the keeper's 45-second timeout
- kept the keeper gated by `ANTIGRAVITY_SESSION_SOURCE=agy` and its bounded child lifecycle
- expanded deterministic coverage for session-source gating and schema behavior

Focused Antigravity, registry, lifecycle, and usage tests pass locally. The rebuilt macOS app also starts from the stable checkout with a healthy embedded Control Center.

duolahypercho:
Thank you for the substantial follow-up here. The registry fixture, agy-only schema normalization, explicit source gate, and timeout alignment are fixed on the current head. Remaining lifecycle blockers are: an expired agy session prevents both the keeper and forwarder from starting after restart; agyProcessEnvironment drops the installed proxy variables and NODE_USE_ENV_PROXY; shutdown disarms escalation and sends only SIGTERM; the synchronous Keychain read has no timeout; and error/close can complete twice. Please make explicit agy mode start independently of token freshness, preserve the recorded proxy environment, retain TERM-to-KILL escalation, bound Keychain reads, guard completion, and provide one real macOS expiry/refresh probe. I am keeping this open—the contribution is valuable and these are contained follow-ups.

duolahypercho:
I pushed maintainer lifecycle repair 1363f48. Explicit agy mode now starts refresh components even when the current token is expired; the child preserves uppercase/lowercase proxy variables and NODE_USE_ENV_PROXY; Keychain reads are bounded; shutdown retains TERM-to-KILL escalation; and refresh completion is once-guarded across error/close. Focused lifecycle tests pass 65/65 with one live skip, and the full suite passes 2,546 with 17 intentional skips.

I am still keeping this open because the decisive live evidence cannot be produced on this host: agy is not installed/configured here. Before merge we need a real macOS Keychain expiry-refresh cycle under launchd with the recorded proxy, restart recovery without manual intervention, and successful tool turns on both new Sonnet 4.6 and Opus 4.6 routes. The code blockers are repaired; the remaining gate is provider/lifecycle proof, not contributor rework.

duolahypercho:
Maintainer repair pushed in commit 8054dd7, preserving the contributor commits and restacking them onto current main.\n\nThe repair fixes a critical account-boundary issue found during review: when explicit agy mode is selected, an expired or unreadable agy credential now fails closed instead of falling through to a possibly different router-managed Google account. Status and health use the same authoritative source decision. Negative tests cover both expired and unreadable agy state while a valid managed token exists, and confirm that managed token is untouched.\n\nVerification is green: the focused Antigravity, project, account, and onboarding suite passed 113/113; the full suite passed 2,634 tests with 17 intentional skips and zero failures; npm run check and diff checks passed. Exact-head GitHub CI is also green on Ubuntu, macOS, Windows, both Electron jobs, the unified macOS app, discovery, and security scanning.\n\nThis remains open pending the existing live gates: real macOS Keychain expiry and refresh under launchd with the recorded proxy, restart recovery without manual intervention, and successful real tool turns on both Sonnet 4.6 and Opus 4.6 routes. Green deterministic CI does not prove those external lifecycle and entitlement behaviors.

duolahypercho:
Thank you again for the substantial work here. I completed another independent security/lifecycle review and am keeping the PR open, but I cannot merge this design. The deterministic hardening is useful; it does not resolve the architecture gate already recorded on #393.

The primary blocker is unchanged: the implementation reads the official agy Keychain item, sends that token to an undocumented v1internal quota endpoint, and deliberately presents vendor identity. #393 explicitly requires a fresh operator-owned OAuth design that never reuses official credential stores or spoofs vendor identity. The Keychain lookup is also service-only (gemini) while account/project metadata comes from a separate source, so the selected token is not atomically bound to the displayed account.

The new review also found concrete lifecycle defects that keep the branch unsafe even if that policy changed:

- A normal reinstall/update can omit ANTIGRAVITY_SESSION_SOURCE and AGY_BIN from the rewritten LaunchAgent, silently switching the router away from the selected agy account while a retained tray may still show/use agy.
- Disconnect retains the agy source marker and the keeper starts independently of provider selection, so it can continue reading Keychain and invoking the quota-consuming agy refresh after the route is disconnected.
- The synchronous manual refresh timeout is not a hard bound: spawnSync sends SIGTERM but has no TERM-to-KILL escalation; a child that ignores TERM can keep it blocked.
- Quota polling reads the managed access token directly without expiry refresh, so an idle valid refresh-token session degrades to local estimates after the access token expires.
- Control Center exposes disconnectable metadata but renders Disconnect only for API setup, leaving the OAuth disconnect path unreachable from the documented desktop flow.

The current remote head 8054dd7 is also conflicting with current main and has no proof for the decisive external path: real macOS Keychain account matching, expiry/refresh under launchd with the recorded proxy, restart recovery, token rotation, real quota/redaction behavior, and successful provider turns. I am not closing the contribution. The safe next direction is the #393 owner requirement: a coherent operator-owned client ID/secret pair, private persistence for login and refresh, an IP-literal ephemeral loopback listener, truthful router identity, and an explicit opt-in live probe.

duolahypercho:
Correction to my prior note: I should not have opened a maintainer replacement around this active contributor PR. I am closing #495 without merging it, and #441 remains the canonical implementation and review thread. The work here is substantial and should stay with its contributor.

The current head still cannot merge for the concrete architecture and lifecycle reasons already recorded above:

1. It reads the official agy Keychain credential and presents vendor identity to undocumented v1internal endpoints. The accepted boundary in #393 is an explicit operator-owned OAuth client ID/secret pair, private router-managed login/refresh storage, truthful router identity, and an ephemeral IP-literal loopback callback.
2. Token, account, project, and displayed quota are not atomically bound to one proved session. A selected agy source must fail closed rather than fall through to a different managed Google account.
3. Reinstall, disconnect, provider selection, and shutdown must preserve one source decision: disconnect must stop the keeper, unselected routes must not refresh, proxy variables must survive service publication, and every child needs bounded TERM-to-KILL completion with a once guard.
4. Expired access tokens need refresh before both quota polling and inference. Control Center must expose the same reachable disconnect path that status advertises.
5. Before merge, please provide a real macOS launchd proof covering Keychain/session migration away from the official store, expiry and refresh through the recorded proxy, restart recovery, account/project matching, redacted quota output, disconnect shutdown, and successful tool turns on both new Claude 4.6 routes.

Deterministic tests remain necessary but cannot substitute for that external account-boundary proof. Once you update this PR, I will review and test this exact contributor head and merge it directly only if the architecture, live evidence, full CI, and security scans are all green.

duolahypercho:
Deterministic repair is complete at exact head `94e6d7311a33021ed4170c3b7c5bebddd5ed8afb`.

The last CI failure was a test-isolation defect: a direct CLI regression invoked the real dependency preflight in the shared checkout, so `npm ci --omit=dev` could prune/replace modules while parallel workers were loading Playwright and Undici. The test now runs from a disposable, dependency-complete checkout. No production OAuth behavior changed.

Verification on this exact head:

- local full suite: 3,233 passed, 23 skipped, 0 failed;
- focused dependency-isolation regression: 8 passed;
- `npm run check` and diff checks pass;
- fresh hosted CI is green across Ubuntu, macOS, Windows, both Electron packages, native macOS, deterministic discovery, and security checks.

This is deterministic-code ready, but I am not merging it without the external proof required by the design: a real operator-owned OAuth client must prove login, truthful billed live probe, managed-restart activation, refresh/restart, disconnect, and real Sonnet 4.6 plus Opus 4.6 tool turns on both macOS and Windows. No provider credential or quota-consuming probe was used without explicit authorization.

duolahypercho:
Correction to the recorded external gate: Sonnet 4.6 and Opus 4.6 are not routes in this branch. Exact head `94e6d7311a33021ed4170c3b7c5bebddd5ed8afb` lists only `antigravity-oauth/gemini-3.1-pro`, `gemini-3.5-flash`, `gemini-3.6-flash`, and `gemini-3.7-flash`.

The PR body now names the proof that can actually be run: on both macOS and Windows, exercise login, the truthful `gemini-3.1-pro` activation probe, managed-restart activation, refresh/restart, and disconnect; after activation, run no-failover basic, forced-tool, and stateless tool-result turns through each listed route. The built-in activation probe is intentionally only a plain-text `gemini-3.1-pro` `OK` request, so it does not by itself certify tool translation or the other listed routes.

This is a gate-documentation correction, not new deterministic code work. The exact head remains fully green and the corrected gate remains credential- and quota-dependent.

duolahypercho:
Superseded by #542, which preserves this branch's complete contributor commit ancestry and attribution, merges current `main`, retains all changelog entries, and adds the bounded Windows process-tree test repair required by the fresh exact-head CI failure. The external macOS/Windows credential-and-quota proof gate remains unchanged; #542 must not merge until it passes.

PR #446, opened 2026-08-25
fix(start): restart the managed service and restore the installed proxy

duolahypercho:
## Problem

`stop` unloads the background service, but `start` exec'd the *foreground* supervisor. The obvious `stop; start` pair was therefore asymmetric: the managed service went away and an unmanaged copy took its place, carrying the calling shell's environment rather than the installed one, and dying with the shell that started it.

A shell spawned by a desktop app has no proxy variables. So the replacement dialled every upstream directly and answered with **a 502 naming an opt-in that was already set** — set in the service definition it had just unloaded. The error surfaced far from its cause.

## Change

- `bin/start` restarts the managed service. The foreground supervisor stays reachable, but only behind an explicit `--foreground`.
- `inheritedProxyEnvironment()` restores the recorded proxy for any start path whose own environment is silent. Silence is the only trigger — `NODE_USE_ENV_PROXY=0` counts as the operator speaking, so a deliberately unproxied run stays unproxied.
- `src/start.mjs` applies it before anything reads the environment or spawns a child, and logs the restored host unconditionally (credential redacted). Nothing near the router previously said which network path it was using, which is the main reason the original failure took so long to find.

## Tests

21 passing, including `start and stop act on the same layer`, `the foreground supervisor is reachable, but only on purpose`, and `a declared proxy environment is left exactly as the operator set it`. Verified the live launchd service is untouched by the run.

### DIFF
--- a/bin/start
+++ b/bin/start
@@ -2,4 +2,29 @@
 set -eu
 
 repo_dir=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)
-exec node "$repo_dir/src/start.mjs"
+
+# `stop` unloads the background service, so `start` has to be able to put it
+# back. It used to exec the supervisor in the foreground instead, which made the
+# obvious `stop; start` pair asymmetric: the managed service went away and an
+# unmanaged copy took its place, carrying the calling shell's environment rather
+# than the installed one and dying with the shell that started it. A shell
+# spawned by a desktop app has no proxy variables, so the replacement dialled
+# every upstream directly and answered with a 502 naming an opt-in that was
+# already set -- in the service definition it had just unloaded.
+#
+# The foreground supervisor is still reachable, just never by accident.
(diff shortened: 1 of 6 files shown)


12duolalalala:
Reviewed and ran this locally on Linux.

- `node --test test/proxy-environment.test.mjs test/service-lifecycle.test.mjs`: 21 passed
- Invoked the real `./bin/start --foreground` with a silent desktop-style env and a recorded proxy in the install manifest. It restored `http://127.0.0.1:3213`, redacted the password, and then stopped on the missing LiteLLM binary (expected in this sandbox).
- `./bin/start --nope` exits 2 with the usage line.

Not merging yet: GitHub still has macOS/Windows tests (and the native macOS app job) in flight. I did not run bare `./bin/start` here, because that would call `service.mjs start` on this machine.

PR #447, opened 2026-08-25
feat(discovery): audit provider model catalogs on a schedule

duolahypercho:
## What

Adds `src/model-discovery-audit.mjs` (287 lines), which walks every anonymous catalog and every API-key catalog whose credential is configured, reporting new candidates, unavailable registered IDs, skips, and failures.

**The report is research input only.** It never invokes curation and never changes the registry, so a newly advertised slug still needs the normal live compatibility proof before it can become a listed model.

## Workflow paths

Three deliberately separate paths:

| Trigger | Behaviour | Secrets |
|---|---|---|
| Pull request | Fixture-backed discovery tests only | **None** |
| Weekly cron (Mon 04:17 UTC) | Live audit of all configured providers | Scoped to the audit step |
| Manual dispatch | `all`, one provider ID, or a comma-separated set | Scoped to the audit step |

Two safety properties worth reviewing:

- Scheduled and manual runs pin checkout to the repository's **default branch**, so a user cannot select a feature branch and expose the provider secrets to it.
- Credentials are declared on the audit step rather than the job, so checkout, setup, and npm lifecycle scripts never receive provider keys.

A targeted run fails when a requested provider's credential is absent; an `all` run records unconfigured providers as skipped, so one missing subscription does not hide the rest of the audit.

## Supporting change

`discoverProviderModels` takes a `fixture` option so the tests can exercise the audit without a live endpoint and without touching the stored list.

## Tests

19 passing across `model-discovery-audit.test.mjs` and `model-discovery.test.mjs`.

### DIFF
--- a/.github/workflows/model-discovery.yml
+++ b/.github/workflows/model-discovery.yml
@@ -1,21 +1,25 @@
 name: Provider model discovery
 
 "on":
+  pull_request: {}
+  schedule:
+    # Monday 04:17 UTC. The offset avoids the busiest top-of-hour window.
+    - cron: "17 4 * * 1"
   workflow_dispatch:
     inputs:
       provider:
-        description: API-key provider to inspect
+        description: Provider ID to inspect, or all configured providers
         required: true
-        type: choice
-        options:
-          - deepseek
-          - kimi-api
+        default: all
+        type: string
 
 permissions:
   contents: read
 
 jobs:
-  discover:
+  deterministic:
+    if: github.event_name == 'pull_request'
+    name: Deterministic discovery tests
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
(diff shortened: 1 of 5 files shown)


12duolalalala:
Reviewed and ran this locally.

- `node --test test/model-discovery-audit.test.mjs test/model-discovery.test.mjs`: 19 passed
- Ran `node src/model-discovery-audit.mjs --provider all`. It wrote a credential-free report, skipped providers without secrets, and succeeded on anonymous catalogs (kilo-free). It did not edit the registry.
- Workflow looks right: PR jobs are fixture-only; live secrets are scoped to the default-branch audit step.

Not merging yet: GitHub macOS/Windows tests are still running. This is a scheduled-audit feature, not a request-path change, so I will land it once those jobs are green.

PR #458, opened 2026-08-26
Apply the choice-bearing usage compat transform to opencode-go (cached_tokens=0 → real cache reads)

### LINKED ISSUE #457: [Bug]: opencode-go (Chat Completions) sessions report cached_tokens=0 and inflated input — LiteLLM drops Console Go's choice-bearing usage chunk; the existing Z.ai compat transform fixes it
First: thank you for this project. It is the only thing that let me run MiMo-V2.5 on my opencode Go subscription from the Codex CLI at all — Go's `/v1/responses` returns 500 for mimo on any request and Codex 0.148 refuses `wire_api = "chat"`, so the LiteLLM bridge here is the whole story. Genuinely appreciated.

### What happened?

Every `opencode-go/*` session (Chat Completions provider) reports **`cached_input_tokens: 0`** to Codex and `cached_tokens=unknown` in `router.log`, and the `input_tokens` figure is inflated — even though Console Go is caching the prefix.

Same model, same day, direct to `https://opencode.ai/zen/go/v1/chat/completions` (`stream: true`, `stream_options.include_usage`), repeated prefix:

| Call | prompt_tokens | prompt_tokens_details.cached_tokens |
| --- | ---: | ---: |
| 1 (cold) | 2,571 | 0 |
| 2 (same prefix) | 2,571 | **2,560** |

Through the router, a two-turn `opencode-go/mimo-v2.5` Codex session:

| Path | Turn 2 input_tokens | cached_input_tokens |
| --- | ---: | ---: |
| Routed, `main` as shipped | 54,218 | **0** |
| Routed, with the one-line change below | 37,609 | **36,416** (97%) |

The 54k vs 37k gap is LiteLLM substituting its own tokenizer estimate for the provider's count once it has dropped the upstream usage object — so cost meters and `ccusage codex` are wrong in both columns, not just the cache one.

**Root cause.** Console Go's Chat Completions stream puts `usage` on the `finish_reason` chunk while `choices` is still non-empty, then `[DONE]`. LiteLLM 1.96.0 discards upstream usage on that shape and falls back to a local estimate — BerriAI/litellm#36168 (open). The router already handles exactly this shape for Z.ai in `src/zai-cache-usage.mjs` (`ZaiCacheUsageCompatTransform` splits the terminal chunk into terminal + usage-only chunk and mirrors `cached_tokens` into `prompt_cache_hit_tokens`), but `zaiCacheUsageTransform()` is gated to `zai-api` / `zai-coding` only, so `opencode-go` never gets it. Same class as #267 (Grok OAuth), different provider.

**Fix.** Add `opencode-go` to the provider gate. That is the whole change; the transform needs no edits. Verified on Windows / Codex CLI 0.148 as in the second table; `node --test test/zai-cache-usage.test.mjs` passes with an added assertion. PR follows.

I did not add `opencode-go-messages` (Anthropic protocol, different stream) or `opencode-zen` — I only verified Go's Chat Completions endpoint.

### Which app do you route through?

Codex (CLI, `codex exec` and the interactive TUI)

### Platform

Windows 11, Node v24.11.0, codex-router v0.4.0-beta.4 (checkout install, `main` @ 7a191fa at the time), LiteLLM pinned 1.96.0.

```
OK    Codex binary: ...\OpenAI\Codex\bin\codex.exe
OK    Enabled providers: opencode-go, opencode-go-messages, opencode-go-responses, opencode-zen
OK    Merged catalog: 23 routed models
OK    Generated gateway config: ~/.codex/codex-router/litellm.yaml
OK    LiteLLM venv runtime: ...\codex-router\.venv\Scripts\litellm.exe exists
OK    Background service: running
OK    Router health: version 0.4.0-beta.4

router.log, before:
[codex-router] timing ... model=opencode-go/mimo-v2.5 provider=opencode-go status=200 out_tokens=113 cached_tokens=unknown est_input=54218
router.log, after the gate change:
[codex-router] timing ... model=opencode-go/mimo-v2.5 provider=opencode-go status=200 out_tokens=24 cached_tokens=36416
```

flashydotwala:
Fixes #457.

One-line gate change: `zaiCacheUsageTransform()` now also covers `opencode-go`. Console Go's Chat Completions stream has the same choice-bearing terminal usage chunk as Z.ai, which LiteLLM 1.96.0 discards (BerriAI/litellm#36168), so routed `opencode-go/*` sessions reported `cached_input_tokens=0` and an inflated `input_tokens` to Codex.

| Two-turn `opencode-go/mimo-v2.5` session, turn 2 | input_tokens | cached_input_tokens |
| --- | ---: | ---: |
| main | 54,218 | 0 |
| this branch | 37,609 | 36,416 |

`router.log` goes from `cached_tokens=unknown` to `cached_tokens=36416`; `ccusage codex session` shows the Cache Read column for the session.

Scope: `opencode-go` only — the provider whose stream I captured and verified. `opencode-go-messages` (Anthropic protocol) and `opencode-zen` were not tested and are left out. Test: added the gate assertion to `test/zai-cache-usage.test.mjs`; `node --test test/zai-cache-usage.test.mjs` passes (6/6).

And thank you for the router — it is what makes MiMo on the Go subscription usable from Codex in the first place.

### DIFF
--- a/src/zai-cache-usage.mjs
+++ b/src/zai-cache-usage.mjs
@@ -89,8 +89,15 @@ export class ZaiCacheUsageCompatTransform extends Transform {
   }
 }
 
+// opencode-go (Console Go /v1/chat/completions) sends the same shape: usage on
+// the finish_reason chunk with a non-empty choices array, then [DONE]. LiteLLM
+// 1.96 discards it (BerriAI/litellm#36168), so Codex reports cached_tokens=0
+// even though Go caches the prefix (verified 2026-08-26: mimo-v2.5 returned
+// cached_tokens 2560/2571 on a repeated prefix when called directly).
+const CHOICE_BEARING_USAGE_PROVIDERS = ["zai-api", "zai-coding", "opencode-go"];
+
 export function zaiCacheUsageTransform(providerId, contentType = "") {
-  if (!["zai-api", "zai-coding"].includes(String(providerId))) return undefined;
+  if (!CHOICE_BEARING_USAGE_PROVIDERS.includes(String(providerId))) return undefined;
   if (!String(contentType).toLowerCase().includes("text/event-stream")) return undefined;
   return new ZaiCacheUsageCompatTransform();
 }
--- a/test/zai-cache-usage.test.mjs
+++ b/test/zai-cache-usage.test.mjs
@@ -33,6 +33,10 @@ test("cache compatibility is installed only for Z.ai event streams", () => {
   assert.ok(zaiCacheUsageTransform("zai-api", "text/event-stream; charset=utf-8"));
   assert.equal(zaiCacheUsageTransform("zai-coding", "application/json"), undefined);
   assert.equal(zaiCacheUsageTransform("deepseek", "text/event-stream"), undefined);
+  // opencode Go's chat endpoint uses the same choice-bearing terminal usage
+  // shape (litellm#36168), so the compat transform covers it too.
+  assert.ok(zaiCacheUsageTransform("opencode-go", "text/event-stream"));
+  assert.equal(zaiCacheUsageTransform("opencode-go-messages", "text/event-stream"), undefined);
 });
 
 test("non-usage and malformed SSE lines pass through byte-for-byte", async () => {


12duolalalala:
Hold: local tests and CI are green, but I could not send a real `opencode-go/*` request from this review box (no operator Go key here), so I am not treating the live path as verified.

What I ran:
- Isolated checkout of this PR only (`fcd2aff` vs current `main`)
- `test/zai-cache-usage.test.mjs` 6/6, including the new `opencode-go` gate and the `opencode-go-messages` negative
- GitHub CI: test (Ubuntu / macOS / Windows), Electron Linux + Windows, native macOS — all SUCCESS
- Mergeable vs current `main` (three-dot diff is still just the two files)

The change itself looks right: exact-id add of `opencode-go` to `CHOICE_BEARING_USAGE_PROVIDERS`. Stream rewrite only, after fetch. Cannot hit `opencode-go-messages` / `opencode-go-responses` / `opencode-zen`. Your two-turn `mimo-v2.5` numbers (54218/0 → 37609/36416) match the existing Z.ai transform.

I will merge once a two-turn `opencode-go/*` Chat Completions request through this build shows `cached_tokens` in `router.log`, or the maintainer confirms they already did that on this tree.

duolahypercho (approved):
Maintainer verification: the change is restricted to the exact opencode-go Chat provider id and SSE responses; Messages, Responses, Zen, and unrelated providers remain negative controls. Contributor live evidence shows cached tokens restored, GitHub cross-platform CI is green, and fresh isolated npm run check plus the focused 6-test suite passed.

PR #459, opened 2026-08-26
fix: preserve custom provider identity in login-free mode

duolahypercho:
## Summary

Preserve an already selected custom `model_provider` while login-free mode is active, so the Codex client keeps the same provider identity instead of remounting its chat surface. The router temporarily replaces that provider's complete table with an ownership-marked, auth-free Responses table, then restores the original table and root model exactly when the mode is disabled.

The built-in `openai` provider is intentionally different: Codex 0.141 requires authentication for its implicit definition, while current 0.150 Desktop builds reject an explicit `[model_providers.openai]` override. Root-OpenAI configurations therefore use the established `codex-router` provider fallback during login-free mode and restore the original root provider afterward.

## Safety and compatibility

- Keeps custom provider IDs unchanged and sets `requires_openai_auth = false` only in the temporary router-owned table.
- Snapshots every replaced provider section, including nested tables, in protected state and restores it only after an ownership check.
- Preserves ownership ID, model selection, and restore metadata across ordinary updates and native-catalog refreshes.
- Restores the prior transport after catalog publication failures and refuses restoration if the direct provider/model source changed concurrently.
- Journals login-free catalog refreshes before parking the transport. Recovery requires a private, non-symlink journal bound to the exact provider-state bytes, ownership ID, canonical route, and display alias; state plus inactive config alone remains fail-closed.
- Serializes the complete refresh with a cross-process lock. A second refresh waits instead of interleaving, and an abandoned stale lock permits the protected journal to resume.
- Only `bin/refresh-catalog` may resume or complete a pending refresh. Ordinary enable/disable/model changes, installation, and doctor repair refuse without mutating state until that recovery command completes.
- Migrates the draft root-OpenAI v3 state to the cross-version provider fallback without losing the original model restore record.
- Keeps the state writer transactional and restores the previous state if the Codex config write fails.

## Verification

- Real signed-out `codex app-server` Responses turns pass through both the root-OpenAI fallback and a preserved custom provider on installed `codex-cli 0.141.0` and bundled `0.150.0-alpha.8`.
- A real custom-provider refresh round trip proves: original table/secret → managed auth-free table under the same ID → failure restoration → successful refresh with identical state/model → byte-identical final provider restoration.
- Hard-exit regressions cover the crash window immediately after transport parking for both custom-provider v3 and root-OpenAI v1 state. Model, provider-table/secret, state-ownership, and recovery-route drift leave config, state, and journal untouched.
- Concurrency regressions cover two-process overlap, bounded lock contention, owner death, and abandoned-lock recovery.
- Pending-journal regressions prove that both explicit disable forms, ordinary enable/disable, and user model changes refuse byte-identically; an interrupted disable without a journal is never revived.
- POSIX and Windows installer/doctor ordering checks prove a pending refresh cannot publish a catalog or falsely report recovery. Malformed, non-private, and symlinked state/journal inputs refuse without mutation.
- Deterministic legacy-layout coverage reproduces the 0.141 configuration shape that lacks the modern agent feature table.
- Windows fixtures use the production private-state writer so ACL assertions exercise the same owner-only state contract as production.
- `npm run check` passes; the exact final full suite reports 2,632 passing, 0 failing, and 17 skipped tests (2,649 total). Repo-maintainer review covers all 108 current hunks with no findings.

The branch is linearly rebased onto current `main`; the original contributor commits are retained, followed by bounded maintainer repair commits.


12duolalalala:
Hold: identity-preserve is the right fix, but login-free-enable now reuses `managedSignedProviderContents()`, which always writes the signed block (`name = "Codex Router (with ChatGPT)"`, `requires_openai_auth = true`).

Login-free exists so a signed-out Codex app can hit the local router without a ChatGPT login. The old path selected the inert `codex-router` provider for that reason. This draft can keep the drawer and break signed-out use.

Need a login-free managed table with `requires_openai_auth = false` (and a name that does not say ChatGPT), tests for both identity-preserve and the auth flag, v1 restore left in place for people already remounted, and the README/tray copy updated. Follow-up is already with the cloud agent.

duolahypercho:
Thank you for the careful work on provider identity. I am keeping this draft open because the current state machine is not yet self-consistent. On a fresh run, the focused config-manager suite has 33 passes and 4 failures: an ordinary enable writes refreshed state without previousModelPresent and the command's own final snapshot rejects it; refresh of existing state does not preserve ownership/restore metadata; transaction rollback preserves signed state but not provider-mode state; and the signed-out root-openai path still lacks an end-to-end proof. Please keep the provider-identity idea, but make provider-mode state a validated transactional shape across enable, refresh, rollback, disable, and signed-out startup before marking ready.

duolahypercho:
Thank you for the substantial provider-identity work here. I repaired the provider-mode transaction and updated this branch to current `main` at `846d4a4`.

The repaired path now preserves and validates provider ownership, `previousModelPresent`, prior provider sections, model/provider restore metadata, ordinary refresh, and symmetric rollback. Custom providers keep their identity through an owned `requires_openai_auth = false` provider table.

There is one necessary cross-version exception: signed-out root OpenAI still switches to the managed `codex-router` provider. Codex 0.141 requires authentication for an implicit built-in OpenAI provider, while current Desktop Codex 0.150 rejects an explicit `[model_providers.openai]` override because `openai` is reserved. The branch documents that split and includes a real signed-out `app-server` turn regression; it passed locally on both Codex 0.141.0 and Desktop 0.150.0-alpha.8 without `auth.json` or an upstream `Authorization` header.

Verification against current `main`: static checks passed, focused tests passed 142/142, and the full suite passed 2,620 with 17 skipped and 0 failed (2,637 total).

I am leaving this PR in draft and not merging yet. Please hold until the fresh CI matrix is green; the real app-server proof has also not yet been exercised on Windows, and the title/description should make clear that identity preservation applies to custom providers while root OpenAI uses the compatibility fallback.

duolahypercho:
Final maintainer verification for exact head `ee305660c80df557f6a756849525c44503ae693b` is complete.

The provider-identity contribution is retained and the repair now closes the cross-version, transactional, privacy, interrupted-refresh, concurrency, and Windows ACL gaps found during review. Custom providers keep their ID; root OpenAI deliberately uses the `codex-router` compatibility fallback on Codex 0.141/0.150. A protected, exact-state journal plus an operation-wide lock makes killed refreshes resumable only through `bin/refresh-catalog`; ordinary install/doctor/config mutations fail closed before catalog publication.

Evidence on the final tree:
- real signed-out app-server turns on Codex 0.141.0 and 0.150.0-alpha.8 for both custom-provider identity and root-OpenAI fallback;
- custom v3/root v1 hard-exit recovery, two-process contention/stale-lock recovery, exact restore, drift/no-journal/security-path refusal;
- local full suite: 2,632 passed, 0 failed, 17 skipped;
- final hunk audit: 108/108 reviewed, no findings; independent final review: no blocker;
- exact-head Ubuntu, Windows, macOS, Electron Linux/Windows, unified native macOS, deterministic discovery, Python lock, Homebrew formula, CodeQL (including Swift), and GitGuardian checks are green.

Thank you for the substantial original provider-identity work. I am marking this ready and merging it rather than closing or replacing the contribution.

Issue #353, opened 2026-08-22
kimi-oauth: Moonshot rejects Codex App connector tools containing sibling $ref pointers (Wego _flights_search) — HTTP 400


hhshopllc:
## Environment
- codex-router v0.4.0-beta.4 (git checkout, macOS, launchd service)
- Provider: kimi-oauth (live Kimi Code OAuth session)
- Codex App with connector tools enabled (Adobe, Wego, etc. — the ~293-tool connector pack)

## Symptom
Any Codex App session using `kimi-oauth/k3` fails with HTTP 400 as soon as connector tools are relayed. Same session on other routed providers works; k3 from CLI (no connectors relayed) works.

## Root cause chain
1. Codex App ships connector tools whose JSON schemas contain **sibling-property `$ref` pointers**. Example from Wego `_flights_search`:
   `properties.filters.properties.inboundTotalDurationRange -> "#/properties/filters/properties/priceRange"`
   — a pointer to a *sibling property*, not `#/$defs/...`.
2. Moonshot rejects any `$ref` that does not start with `#/$defs/`.
3. The router sanitizer (`providerToolSchema`) only flattens union/nullable **roots**; it does not resolve sibling-property `$ref`s before relaying to kimi.
4. Grok has a dedicated forwarder using a stronger collapse; kimi's path does not.

## Suggested fix
Extend `providerToolSchema` (or the kimi forwarder) to resolve/inline non-`#/$defs/` `$ref` pointers — including sibling-property references — before sending tool schemas to Moonshot. This would also future-proof other strict-schema providers.

## Workaround
Use k3 from the CLI (connector tools are not relayed there), or switch the session to another routed model when App connectors are needed.

Happy to provide full request/response captures if useful. Thanks for the router — it works great otherwise.

duolahypercho:
Confirmed still present on `main`, and your diagnosis holds all the way down. Recording the trace so this is actionable.

## Root cause

`providerToolSchema` (`src/tool-schema-root.mjs:259`) does exactly two things: align literals with their declared type, and collapse a union/nullable **root**. A schema whose root is a plain object is returned by identity, with every `$ref` intact:

```js
if (!hasRootUnion(normalized) && !hasNullableObjectRoot(normalized)) return normalized;
```

`resolveRef` exists (`:27`) but is reachable only from `objectBranches`, which feeds the root collapse — it is never used to inline refs into the relayed copy.

[...]

Issue #179, opened 2026-08-13
kimi-api default baseUrl causes 401/429 cooldowns, and tool schema enum type mismatches trigger 400 on Moonshot API


MohamedMohsenX2:
## Bug Report

**codex-router version:** 0.4.0-beta.3 (installed via Homebrew)
**OS:** macOS
**Provider:** kimi-api (Kimi Platform API)

---

### Bug 1: kimi-api defaults to `https://api.moonshot.cn/v1` — no first-class support for Kimi Platform endpoint

**Symptom:** Requests to `kimi-api/kimi-k3` fail with HTTP 401 Unauthorized, which cascades into LiteLLM's 429 Too Many Requests deployment cooldown.

**Root Cause:** The default `baseUrl` in `config/kimi/kimi.json` is `https://api.moonshot.cn/v1` (Moonshot's legacy Chinese API endpoint). Modern Kimi Platform API keys (sk-kimi-xk...) are issued for `https://api.kimi.com/coding/v1`. The legacy endpoint rejects these keys with a 401.

**Current workaround:** Setting `KIMI_API_BASE_URL=https://api.kimi.com/coding/v1` in the environment fixes the endpoint mismatch. This works because the repo already respects `baseUrlEnv` overrides.

**Gap in current codebase:** `provider-account-usage.mjs` only recognizes `api.moonshot.ai` and `api.moonshot.cn` as valid hosts for balance checks (line 401). When `api.kimi.com` is used, it falls through to `localOnly("Account balance is unavailable for a custom Kimi API endpoint")`. The OAuth forwarder (`oauth-forwarder.mjs`) already uses `api.kimi.com/coding/v1`, so the platform endpoint is known to the codebase but not fully integrated as a first-class kimi-api endpoint.

**Suggested fix (following repo patterns):**
1. Keep `https://api.moonshot.cn/v1` as the default `baseUrl` for backward compatibility
2. Add `api.kimi.com` to the recognized hosts in `provider-account-usage.mjs` so balance checks work for both endpoints
3. Optionally: expose both endpoints in the provider config or documentation so users know `KIMI_API_BASE_URL=https://api.kimi.com/coding/v1` is the correct setting for Kimi Platform keys

---

### Bug 2: Strict Moonshot JSON Schema validation rejects tool definitions with enum/type mismatches

**Symptom:** HTTP 400 Bad Request:
```
tools.function.parameters is not a valid moonshot flavored json schema,
[...]

PR #460, opened 2026-08-26
fix(kimi): expand definition refs with siblings

MohamedMohsenX2:
Follow-up to #353. This is separate from the deferred Kimi Coding endpoint and model-id work discussed in #179.

## Bug

Current Codex `automation_update` tool schemas include `$defs` entries such as:

```json
{
  "description": "Automation id...",
  "$ref": "#/$defs/__schema1"
}
```

Moonshot rejects a `$ref` that carries sibling keywords. #367 already handles sibling-property pointers, but it intentionally leaves every `#/$defs/` pointer untouched. A current Codex build now puts sibling keywords on definitions themselves, so `kimi-api` receives the invalid shape and returns the observed HTTP 400 schema rejection.

The bundled snapshot scan finds eight sibling-ref definitions in `automation_update` before repair and zero after this patch.

## Fix

Extend the existing strict-route inliner to expand a `#/$defs/` reference when it has siblings:

- merge the resolved target with the reference node siblings, with siblings winning as before;
- resolve acyclic pure definition aliases while expanding a decorated reference;
- preserve ordinary pure `#/$defs/` references exactly;
- keep the existing depth, expansion, cycle, and byte-budget protections;
- on a genuine cycle or conflicting assertion, keep only the pure reference rather than silently weakening either side;
- apply the compatibility pass to `kimi-oauth`, `kimi-api`, and `kimi-api-cn`, while leaving every non-Moonshot provider wire payload unchanged.

The Kimi API routes now share the gate because the new shape was reproduced with the separately billed Kimi Platform API key, not only OAuth.

## Tests

- exact synthetic sibling-definition expansion;
- decorated definition through a pure alias chain;
- conflicting sibling handling;
- cyclic sibling handling;
- regression against the bundled Codex `automation_update` snapshot: sibling refs present before repair, absent after repair, and `__schema0` expanded;
- existing negative control proving a provider that does not request inlining remains unchanged.

## Verification

- `npm run check`
- `git diff --check`
- `node --test test/namespace-relay.test.mjs test/tool-schema-root.test.mjs`
- `npm test`: 2,557 tests, 2,539 passed, 18 intentional skips, 0 failures

No live provider request was made from this change, so no quota was spent.

### DIFF
--- a/src/namespace-relay.mjs
+++ b/src/namespace-relay.mjs
@@ -314,11 +314,11 @@ export function repairToolSchemaRoot(
   tool,
   { nonRecursive = false, inlineForeignRefs: inlineRefs = false } = {},
 ) {
-  // Moonshot alone rejects a `$ref` that does not point into `#/$defs/`, so
-  // only the route that asks for it pays the inlining -- every other provider
-  // keeps the exact wire payload it has today. `inlineForeignRefs` returns a
-  // schema with no foreign ref by identity, so even on that route an ordinary
-  // toolset is not copied.
+  // Moonshot alone rejects a `$ref` that does not point into `#/$defs/` or one
+  // that carries sibling keywords, so only the route that asks for it pays the
+  // inlining -- every other provider keeps the exact wire payload it has today.
+  // The normalizer returns a clean schema by identity, so an ordinary toolset
+  // is not copied.
   const relaySchema = (schema) => {
     const repaired = providerToolSchema(schema);
     return inlineRefs ? inlineForeignRefs(repaired) : repaired;
--- a/src/router.mjs
+++ b/src/router.mjs
@@ -594,17 +594,15 @@ function needsZenFreeToolCompatibility(route) {
   );
 }
 
-// Moonshot accepts a `$ref` only when it points into `#/$defs/` and rejects the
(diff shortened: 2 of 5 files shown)


duolahypercho:
I pushed maintainer follow-up 960b89a to preserve the contributor's approach while closing one schema-safety gap. When a decorated $defs reference conflicts with its target, or closes a cycle, reducing it to a pure reference silently drops assertions. The repair now keeps the original decorated node intact so the boundary fails closed instead of weakening the schema, with regressions in both directions and for cycles.

Verification: npm run check passes; 92 focused namespace/schema tests pass; the full suite reached 2,540 passes with one extraction-environment failure in skills-install, and that failing test file passes 23/23 from the real Git checkout. I am waiting for the fresh cross-platform CI on this head before merging.

duolahypercho (approved):
Maintainer review after follow-up: the contribution now expands the Kimi/Moonshot ref-with-siblings shapes without weakening conflicting or cyclic schemas. Syntax, 92 focused tests, and the full local suite (2,541 pass, 17 intentional skips) are clean. Approved pending the refreshed cross-platform CI.

PR #462, opened 2026-08-26
Add GLM-5.3-Flash on OpenCode Go

mirucon:
## Summary

- replace OpenCode Go's withdrawn `ox-alpha-free` route with `glm-5.3-flash`
- migrate both `opencode-go/ox-alpha` and the older curated `opencode-go/ox-alpha-free` slug to the named route
- preserve the preview's measured low/high/max effort normalization and update discovery, picker, docs, and regression coverage
- publish OpenCode Go's provider-served 1,000,000-token context rather than the base model's 1,048,576 architectural maximum
- auto-compact Flash conservatively at 400,000 after large live Codex histories exposed repeatable empty completions below the generic 850,000 threshold

The 1,000,000 value remains catalog truth. The 400,000 compaction limit is an operational guard, not a claim that the provider's advertised context is smaller.

## Verification

- live smoke: `opencode-go/glm-5.3-flash` returned HTTP 200 and the expected marker
- public catalog audit: OpenCode Go reports 1,000,000 context and 131,072 output; the released Z.ai model config independently reports the 1,048,576 base-model maximum
- compaction regression: asserts the 400,000 guard while preserving more than the full output allowance
- `npm run check`: passed
- post-operational-correction focused registry/catalog suite: 116 passed
- post-operational-correction expanded discovery, control, configuration, registry, catalog, alias, and user-model suite: 197 passed
- earlier broader focused discovery, registry, alias, routing, control, and user-model tests: 232 passed, 2 platform-specific skips
- earlier full suite: 2,446 passed, 73 skipped, 7 unrelated Windows/environment failures (desktop panel, two legacy migration fixtures, three symlink-restricted packaged-install fixtures, and one tray-service fixture)

Authenticated OpenCode discovery reports `glm-5.3-flash` registered and no unavailable registered IDs.


duolahypercho (approved):
Maintainer verification: this is the preferred GLM-5.3-Flash migration. It removes the withdrawn OpenCode Go route, preserves both historical slug aliases, retains the measured low/high/max effort normalization and image support, and keeps a conservative 400K compaction guard with the full output reserve. Live HTTP 200 evidence and all GitHub CI are green; fresh isolated npm run check and the 186-test focused registry/discovery/control/routing/user-model suite passed.

PR #463, opened 2026-08-26
test: isolate model discovery from user overlays

slegarraga:
## Summary

- isolate `MODEL_ROUTER_USER_MODELS` before importing the checked-in registry in the discovery test file
- keep fixture comparisons deterministic for developers who have locally curated models
- restore the inherited environment and remove the temporary state after the test file finishes

## Reproduction

With locally curated OpenCode Go models present, `test/model-discovery.test.mjs` previously failed two assertions because both the parent test process and its discovery subprocesses loaded the developer's real `user-models.json`. Pointing the overlay at an empty temporary file made all 15 tests pass.

The override now happens before `model-registry.mjs` is imported, so the file consistently compares fixtures against the checked-in registry.

## Validation

- `npm run check`
- `node --test test/model-discovery.test.mjs test/state-owner.test.mjs` (26 passed)
- `node --test --test-reporter=dot test/*.test.mjs` (exit 0)

### DIFF
--- a/test/model-discovery.test.mjs
+++ b/test/model-discovery.test.mjs
@@ -3,10 +3,23 @@ import { execFileSync } from "node:child_process";
 import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import os from "node:os";
 import path from "node:path";
-import test from "node:test";
+import test, { after } from "node:test";
 import { fileURLToPath } from "node:url";
 
 const root = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..");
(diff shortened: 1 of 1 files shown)


duolahypercho (approved):
Maintainer verification: the environment override is installed before the registry import, cleanup restores the inherited value, and the change is confined to deterministic test isolation. Fresh isolated verification passed npm run check, the 26 focused discovery/state-owner tests, and the full suite (2,535 passed, 17 intentional skips, 0 failed).

PR #464, opened 2026-08-26
feat(opencode-go): add four certified Chat models

slegarraga:
## Summary

- add GLM-5, Kimi K2.5, LongCat-2.0, and Qwen3.5 Plus to the checked-in OpenCode Go Chat registry
- record current OpenCode metadata, safe output reserves, image capability, effort controls, and deprecated labels without promoting any route to native v2 subagents
- certify all four ids as Chat routes for discovery/curation while keeping unknown protocols fail-closed
- refresh the authenticated catalog fixture and README model table, including the previously omitted Grok 4.6 row
- rebase on current `main` after #462 and rely on its GLM-5.3-Flash route, migration aliases, measured request profile, and 400K compaction guard unchanged

## Catalog and metadata evidence

Checked on 2026-08-26 against the authenticated OpenCode Go `/models` endpoint and OpenCode's `opencode-go` record in `models.dev/api.json`:

- 31 live ids discovered and 29 registered after this branch
- `hy3-preview`, `mimo-v2-omni`, and `mimo-v2-pro` remain unregistered and blocked because no working protocol route was found
- the withdrawn `ox-alpha-free` id remains absent from discovery; #462 owns its migration aliases to GLM-5.3-Flash
- each new context window matches OpenCode's current per-id metadata and each auto-compact threshold reserves at least the documented maximum output

The three deprecated ids are labelled as legacy in the picker. OpenCode publishes no named effort ladder for these four ids, so unnamed toggle/budget reasoning controls are represented conservatively as a single `high` level.

## Live protocol evidence

Each model passed the router's billed compatibility test for basic response, streaming, tool calling, and compaction on 2026-08-26. Direct Chat Completions tool probes then checked whether the relaxed request profile was actually necessary:

| Model | `tool_choice: required` | Resulting profile |
| --- | --- | --- |
| GLM-5 | 200 + valid tool call | default |
| Kimi K2.5 | 400; `auto` returned 200 + valid tool call | `auto-tool-choice` |
| LongCat-2.0 | 200 + valid tool call | default |
| Qwen3.5 Plus | 200 + valid tool call | default |

Qwen3.5 Plus intentionally uses Chat rather than following the neighboring Qwen Messages entries: the Chat route passed the full compatibility suite, while the Messages route failed the router tool-call path.

## Validation after rebasing onto current `main`

- `npm ci` (0 vulnerabilities)
- `npm run check`
- focused registry/discovery/curation suite: 86 passed, 0 failed
- full `npm test`: 2,582 passed, 16 skipped, 0 failed


duolahypercho:
Thank you — the four non-overlapping Chat routes here are worth keeping, and the focused suite was reproduced locally (npm run check; 85 tests pass). I merged #462 for GLM-5.3-Flash because it carries the old opencode-go/ox-alpha aliases, preserves the measured request profile, image capability, low/high/max ladder, and uses the live 400K compaction guard. Please rebase after #462, drop this PR's competing GLM-5.3-Flash file, and retain GLM-5, Kimi K2.5, LongCat-2.0, and Qwen3.5 Plus. Then let the full GitHub CI run; only GitGuardian has run on the current head.

slegarraga:
Thank you for the thoughtful review and for reproducing the focused suite — I really appreciate it.

I've rebased onto current `main`, dropped this PR's GLM-5.3-Flash implementation entirely, and retained only GLM-5, Kimi K2.5, LongCat-2.0, and Qwen3.5 Plus. The route merged in #462 — including its migration aliases, measured request profile, image capability, effort ladder, and 400K compaction guard — is unchanged.

I also refreshed the PR title/body and reran validation on the new base:

- `npm run check`
- focused registry/discovery/curation suite: 86 passed, 0 failed
- full `npm test`: 2,582 passed, 16 skipped, 0 failed

The updated head is `fe47cda`, so GitHub CI can now run against the rebased branch. Thanks again for the clear direction and for the careful measurements carried by #462.

duolahypercho (approved):
The requested rebase/rescope is complete: the already-merged GLM-5.3-Flash work is no longer duplicated, and the four remaining Chat routes retain their live protocol evidence, conservative output reserves, and fail-closed curation mapping. Independent focused checks and the completed functional/security lanes are green. Thank you for carrying this through.

PR #474, opened 2026-08-27
fix: support reviewed external skill coexistence

PracticalSwan:
## Summary

Allow an operator to explicitly approve an externally managed copy of a shipped Codex skill without granting codex-router ownership of that directory.

This fixes the case where the same reviewed cross-client skills are managed outside codex-router (for example, shared across Codex/Copilot/Claude) and therefore correctly fail the existing ownership check, but still satisfy the router's required skill contract.

## Security model

This does **not** weaken the existing fail-closed ownership boundary:

- router-owned skills still require the protected private ownership record plus the matching random on-disk marker;
- external coexistence is a separate explicit approval path and never creates an ownership marker;
- approval binds both the exact external directory tree digest and the exact corresponding router-shipped skill directory digest;
- symlinks and unsupported filesystem entries are rejected;
- any external-content change or router skill-source change invalidates approval and returns the skill to collision/missing status until it is reviewed again;
- unknown or malformed content is never auto-adopted;
- uninstall removes only verified router-owned skills and preserves approved external directories;
- malformed protected state continues to fail closed.

The CLI adds `approve-external` and `revoke-external`. Doctor now distinguishes verified managed skills, approved external skills, stale external approvals, and unverified collisions, and prints the explicit review/approval command for collisions.

## Verification

Rebased on `4bfb888` (`fix: preserve custom provider identity in login-free mode (#459)`).

- `npm.cmd run check` — pass (`v2-agent applications valid (6)`, syntax checks passed)
- focused post-rebase suite covering skill ownership/coexistence, doctor/installer behavior, and #459's affected config/refresh-state paths — 113 tests total, 101 passed, 12 platform skips, 0 failed
- `git diff origin/main..HEAD --check` — clean

The focused suite includes negative controls for external-content drift, router-source drift, corrupt ownership state, user-owned collisions, pending login-free refresh state, and restoration/refusal paths.


duolahypercho:
Thanks for putting this together. The dual target/source digest model is a good direction, and the preservation behavior is worth keeping. I reviewed exact head `8c2a0bea` and found two blockers before merge:

1. **Approved external content can later be deleted.** `approveExternalSkills` adds `external[name]` but leaves a stale managed `skills[name]` token. I reproduced this sequence: install a managed skill → replace it with external content → approve it → restore the old hidden ownership marker → uninstall. The approved external directory was removed. Approval must make the managed and external maps disjoint, the approved-install path should clear stale managed ownership, and existing overlapping state must fail closed or be normalized safely. Please add this full transition as a regression test.

2. **Digest failures can abort doctor.** `directoryDigest()` allows `readdirSync` and `readFileSync` errors to escape. Making an approved `SKILL.md` unreadable caused `skillPackStatus()` to throw `EACCES`, and doctor calls it without a guard. Read errors and concurrent tree changes should invalidate the approval and produce a stale/collision diagnostic instead of terminating doctor.

The operator workflow also needs to be reachable through the supported POSIX and Windows CLI surfaces, with doctor displaying the actionable approval/revocation command. Please clear or correctly report approvals for skills removed from the shipped pack and use locale-independent digest ordering.

Positive evidence: `npm run check`, the focused skill/install/doctor suite, diff-check, and the GitHub Linux/macOS/Windows matrices are green. This looks like an **adapt**, not a rejection: once the ownership transition and fail-closed behavior are corrected, the core design should be valuable.

duolahypercho:
Maintainer update: I pushed the reviewed repair series onto this contributor branch while preserving the original coexistence design.\n\nThe repair now:\n- keeps router-managed and explicitly approved external skills disjoint and fails closed on overlap or stale digests;\n- bounds marker, ownership, tree-entry, depth, and byte reads;\n- stages and validates the complete replacement before retiring the old tree;\n- journals retirement and publication so SIGKILL during first install or replacement is recoverable;\n- never recursively deletes an unverified public target and preserves ambiguous old content under a visible sibling;\n- serializes install, uninstall, approve, and revoke across processes;\n- makes every CLI failure return nonzero.\n\nThe first exact-head Windows run exposed that the existing shared lock's 2-second wait was too short for Windows owner-only ACL writes. Commit 66e6d2f gives only skill mutations a bounded 30-second wait; credential and generic-provider callers retain the existing 2-second default. The concurrency regression now holds the first process for 2.5 seconds, so it proves the override rather than relying on host timing. An independent follow-up review found no blocker.\n\nVerification on exact head 66e6d2f: focused tests pass 60/60; the full repository suite passes 2,655 tests with 17 expected skips and zero failures; npm run check and diff checks pass. Fresh exact-head GitHub CI is running, and I will merge only if the repaired Windows job and the rest of the matrix are green.

PR #477, opened 2026-08-27
fix: restack provider API-key pools on shared credential store

MeltingMadness:
## Summary

This is a current-`main` restack of the provider API-key pool work discussed in #403. It addresses the three blockers in the latest maintainer review while preserving the existing fail-closed and pre-commit safety behavior.

## Review blockers addressed

- Uses only the accepted shared `cred_...` credential IDs and provider credential resolver. The pool-specific secret-reference format from #403 is removed.
- Adds a safe lifecycle through `codex-router key-pool`: existing credential upsert, registry-constrained `add-env`, pause/resume, policy changes, and sanitized status.
- Adds an end-to-end OpenCode forwarder test that configures two credential IDs, receives a `429` for the first key before any response is committed, rebinds to the second key, and returns only the successful response.

Configured invalid, empty, locked, or unresolvable pools remain authoritative and fail closed rather than falling back to the legacy key. Pool state contains credential IDs and metadata, never secret bytes.

## Verification

- Changed runtime modules pass `node --check`.
- 29 targeted credential, pool, routing, control, and forwarder tests pass; 0 fail.
- Both CLI dispatch paths were checked.
- `git diff --check` passes.

A full repository-wide suite was not run locally; CI remains the broader regression gate.

## Relationship to #403

This PR carries the revised implementation on current `main` because the original cross-fork branch cannot be updated from this account. It is intended to provide the maintainer with a directly reviewable and mergeable successor to the implementation in #403.


MeltingMadness:
Follow-up from an adversarial self-review: `key-pool add-env` is now idempotent and concurrency-safe at the credential-store boundary (`fdfe418`). Repeating the same provider/environment registration reuses the active credential ID instead of creating duplicate refs that would intentionally fail pool selection. A regression test covers repeat registration, one stored ref, and one pool entry. The targeted set remains 29 passed / 0 failed, with changed-module syntax checks and `git diff --check` passing.

duolahypercho:
Independent maintainer audit of successor head fdfe418 against current main: ADAPT, do not merge this exact head yet.\n\nThe shared cred_ references, provider binding, authoritative pool behavior, locked mutations, duplicate-secret check, and two-key pre-commit 429 path are worth preserving. Remaining blockers:\n\n1. Command Code routing is chosen from the initial key, but a successful rotated key identity is not propagated to route selection, plan relay, or route-cache accounting. A result can be sent or recorded against the wrong account/plan surface.\n2. Pooled failures call attemptResponse.text(), bypassing current main's bounded response reader and allowing a provider-controlled error body to consume unbounded memory.\n3. Retry policy includes 500 and almost all 5xx, with as many as 256 immediate account attempts and no total time/backoff budget. A 500 may have executed and can duplicate billed work. Conversely, real fetch transport codes commonly live at error.cause.code, so ECONNRESET does not rotate.\n4. initialSelection is trusted without re-reading pool and credential state immediately before send. I reproduced selecting key A, pausing A, then still sending A.\n5. Lifecycle is incomplete: there is no remove-entry/delete-pool path back to the legacy key, no PowerShell key-pool command, and doctor/provider health/support diagnostics do not inspect an authoritative broken pool.\n6. Pool parsing reads unbounded symlink-following state and silently slices oversized maps. A 257-entry file reports valid, then the next mutation permanently rewrites only 256 entries.\n\nPlease propagate the winning credential identity, choose Command Code routing per attempt, reuse the bounded body and established retry classifier with a small total budget, revalidate before send, complete remove/Windows/doctor/health integration, and fail closed on bounded non-symlinked state. The synthetic merge was clean and targeted tests passed, but GitHub currently shows only GitGuardian rather than the core CI matrix.

MeltingMadness:
Addressed the six blockers from the review in a1ddc17 (rebased onto current main):

1. The credential that wins pool failover is now propagated through Command Code route selection, plan relay, entitlement caching, and pool outcome accounting. A new end-to-end test proves key A can fail, key B can win, and the plan relay uses key B.
2. Pooled error bodies and the non-pooled Command Code refusal body now use the existing bounded response reader instead of Response.text().
3. Pool failover reuses the established upstream retry classifiers, including nested error.cause.code traversal. HTTP 500 is no longer retried. Attempts are capped at 3 with a 5-second retry budget and 250ms/750ms backoff.
4. The initial availability selection is no longer trusted for sending. The authoritative pool is re-read under its lock immediately before every attempt; a regression test pauses the initially selected key before send and confirms the next key is used.
5. Added remove-entry and delete-pool lifecycle operations, session/policy cleanup, Windows key-pool dispatch, and sanitized pool visibility in doctor, /health, and support bundles. Removing a pool entry intentionally preserves the shared cred_* reference.
6. Pool state reads are size-bounded, reject symlinks/non-regular files, and reject over-limit provider/credential/session/policy maps instead of slicing them. A 257-entry regression fixture is invalid and mutations refuse to overwrite it.

Fresh targeted verification on this head: 37 tests passed, 0 failed; module syntax checks and git diff --check passed. PowerShell execution was not available on the macOS test host, so the Windows dispatcher change was statically inspected rather than executed.

duolahypercho:
Maintainer follow-up: thank you for the substantial rework in `a1ddc178`. I preserved all contributor commits and fast-forwarded two additive maintainer fixes onto this same PR (`040b08f`, `52889a7`) rather than replacing or closing the work.

The final adversarial pass found and now covers five remaining boundaries:

- repeated `add` / `add-env` preserves paused state, priority, quota, health, and request/token counters;
- non-sticky round-robin advances `A, B, C, A, B` even with one repeated session;
- every Command Code plan/cache write fingerprints the key that actually won failover;
- doctor and `/health` fail closed for empty, paused/no-eligible, unresolvable, duplicate-secret, or structurally invalid authoritative pools, while exposing only fixed reasons/counts;
- a send failure after response commitment propagates `committed: true` and cannot fall through to a second 503 response.

Local exact-head verification on `52889a71d2c4f738498e33626e778aa4e5745ce8`:

- focused adjacent suite: 40/40 passed;
- `npm run check`: passed;
- full `npm test`: 2,816 total, 2,799 passed, 17 intentional skips, 0 failed;
- `git diff --check`: clean;
- synthetic merge with current `main` (`3fd0cae`) is conflict-free.

No live provider or quota-consuming request was used. I am holding merge only for the fresh exact-head GitHub matrix and a final independent diff/check review.

duolahypercho:
Thank you for the substantial key-pool implementation. I preserved the contributor branch and pushed additive repair commits through bc9c253: bounded/transactional pool state, per-key quota and outcome ordering, safe retry/relay boundaries, service-environment and Windows ACL handling, catalog/publication integration, and fail-closed environment-backed lifecycle checks. The exact head passes locally: 2,922 total, 2,904 passed, 18 skipped, 0 failed, plus npm run check and git diff --check. GitHub CI is running now; I will merge only after every required job, especially Windows, is green.

duolahypercho:
Correction to my prior note: I should not have opened a maintainer continuation around this active contributor PR. I am closing #493 without merging it, and #477 remains the canonical implementation and review thread. Thank you for the substantial revisions already made here.

The remaining blockers I reproduced are:

1. The Windows request-path ACL worker treats powershell.exe stdin/stdout as a classic persistent line protocol. The real Windows regression Windows request-path private writes reuse one ACL worker and preserve the atomic target ACL times out after 30 seconds while the synchronous ACL tests pass. The exact failure is https://github.com/duolahypercho/codex-router/actions/runs/33126145362/job/98704657313. Please replace that transport with a bounded Windows-compatible design, such as a one-shot operation or an owner-restricted named pipe, and keep the real reuse/ACL test.
2. The support bundle serializes providerApiKeyPoolsSnapshot, which includes raw credential IDs, session IDs, and arbitrary health.lastError history. Support artifacts should contain only the authority/readiness decision and aggregate counts; add sentinels proving IDs, history, and secret values never appear.
3. commandcode-plan.json persists a fast truncated SHA-256 verifier for each provider key. That turns a support/state leak into an offline key verifier. Please use a memory-hard versioned derivation or avoid persisting a secret-derived verifier, and make old entries miss safely.
4. The same review found avoidable fast token digests in short-lived Copilot/pool comparisons. The resolved secrets already exist in request memory, so exact short-lived comparisons avoid retaining reusable hashes.

Once this branch is updated, I will rerun the focused security/pool/Command Code suite, the full exact-head matrix, and code scanning. I will merge this contributor PR directly only when those checks are green.

duolahypercho:
@MeltingMadness, the Windows core CI job is still failing on the current head. The direct regression is `Windows request-path private writes reuse one ACL worker and preserve the atomic target ACL`, which times out after 30 seconds in `src/file-security.mjs`; the other provider-pool failures appear largely downstream of that ACL-worker timeout.

Failing job: https://github.com/duolahypercho/codex-router/actions/runs/33119761086/job/98683498220

duolahypercho:
I prepared maintainer continuation #507 from the preserved pool history and latest review fixes. It is based on current `main` and replaces the hanging persistent Windows PowerShell ACL worker with a bounded one-shot async operation; it also carries the support-bundle redaction and versioned credential-verifier fixes from the latest review head. Exact-head focused verification is green locally; #507 is running the full matrix, especially Windows. I’ll close this thread as superseded only after that matrix is green and the successor is merged.

duolahypercho:
Superseded by merged #507, which carries the provider pool fixes and Windows/security repairs.

PR #478, opened 2026-08-27
fix(antigravity): sanitize Claude tool schemas

jepgambardella:
## Problem
The Antigravity Claude adapter rejects valid Codex tool schemas when they contain legacy `id`/`discriminator` metadata, boolean JSON-schema subschemas, or unions it cannot translate. This causes HTTP 400 before Claude can process a turn.

## Fix
- Strip unsupported `id` and `discriminator` metadata recursively.
- Convert boolean subschemas to empty schemas.
- For Claude only, remove null branches, collapse single-branch unions, and omit unsupported multi-branch unions. Gemini behavior is unchanged.
- Add regression coverage.

## Validation
- Focused shape suite: 20/20 passed.
- Live local Antigravity forwarder with the cached 263-tool Codex harness: HTTP 200, 94k input tokens, `OK` response.
- `git diff --check` passed.

### DIFF
--- a/src/antigravity-oauth-shape.mjs
+++ b/src/antigravity-oauth-shape.mjs
@@ -315,6 +315,8 @@ function dereferenceAntigravitySchema(schema, root = schema, stack = new Set(), 
 const UNSUPPORTED_SCHEMA_KEYWORDS = Object.freeze([
   "$schema",
   "$id",
+  "id",
+  "discriminator",
   "$anchor",
   "$dynamicAnchor",
   "$dynamicRef",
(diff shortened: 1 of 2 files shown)


duolahypercho:
Thank you for the live 263-tool proof and the narrow two-file change. I am keeping this contribution open, but the current sanitizer is not safe to merge yet.

The blocking semantic problem is boolean JSON Schema: false means “accept no value,” while this patch converts both true and false to {}, which means “accept any value.” That reverses an unsatisfiable constraint into a permissive one. The Claude multi-branch path is similarly lossy: dropping a multi-branch anyOf/oneOf/allOf can turn a constrained field into {}, so a 200 from a large request does not prove that generated tool arguments still satisfy the client contract.

Please preserve true and false as distinct cases (or refuse/drop only the affected tool fail-closed), and add tests that demonstrate the intended semantics for false plus multi-branch unions. The live proof also needs a sanitized pre-fix rejection/negative control that isolates which keyword or shape the upstream rejected; the 263-tool request currently proves the combined transform was accepted, not that every widening is required.

Finally, please rebase onto current main and retain #479’s enum/const literal regression. That merged fix stringifies protobuf literal values and should not be replaced by a broader schema-erasure path. No need to close or discard this work—the id/discriminator evidence may still land cleanly once the transformations are separated and the lossy cases are handled explicitly.

duolahypercho:
The maintainer continuation is now #494. It preserves this PR's contributor commit `a90214d` as a direct ancestor and adds the conservative schema/forced-choice repair described in review.

I am leaving #478 open until #494 completes its replacement cross-platform gate and merges. If it succeeds, I will close this PR as superseded while retaining and explicitly crediting the contributor implementation.

duolahypercho:
Merged through the reviewed continuation in #494. Your original commit remains in main ancestry, so GitHub correctly records this PR as merged rather than closed. The continuation kept the schema-sanitization work and added the fail-closed forced-tool, union, mutation-safety, and cross-platform regression coverage required by review. Thank you for the contribution.

PR #493, opened 2026-08-27
fix: complete safe provider API-key pools

duolahypercho:
Continuation of #477, preserving the contributor/restack history through commit `75a2f57` as direct ancestors. This owner branch exists so the final maintainer repair can receive an exact-head Windows gate without flattening the original work.

What this adds beyond #477:
- keeps the full Windows test suite and every real ACL assertion, but runs Node test files serially so hundreds of concurrent PowerShell ACL workers cannot starve their fail-closed timeout;
- adds a regression that locks the Windows/non-Windows workflow split in place;
- leaves production ACL behavior and Linux/macOS parallelism unchanged.

Verification before push:
- focused Windows cascade set: 59 passed, 4 non-Windows skips, 0 failed;
- adjacent workflow/security set: 54 passed, 4 skips, 0 failed;
- `npm run check`; `actionlint -shellcheck=`; `git diff --check`;
- previous exact head passed Linux, macOS, Python-lock, discovery, Electron, Homebrew, and macOS-app jobs; only Windows full-suite failed from ACL-worker starvation.

Merge gate: the replacement exact-head Windows job must pass.

Authorship credit remains in the preserved commit graph from #477 and the original #403 work.


duolahypercho:
Security review update: CodeQL identified fast SHA-256 verifiers derived from provider credentials. The exact head now uses a memory-hard Command Code verifier (with an intentional one-time entitlement-cache miss for the old format), compares already-resolved pool secrets only within request scope, and scopes the Copilot in-memory cache by the exact token it already retains for inference. The focused pool, Command Code, and Copilot suite passes 44/44. This branch is also merged forward to current main; exact-head CI is rerunning.

duolahypercho:
Native Windows CI found one more real boundary bug: the persistent PowerShell ACL helper never answered its first stdin request when its program was passed through `-Command`. The latest head now passes that static worker program through UTF-16LE `-EncodedCommand`, leaving stdin exclusively for the base64 request protocol. This preserves the owner-only DACL and atomic-rename behavior; it does not relax ACL checks or increase timeouts. Local `npm run check` and 43 focused file-security/key-pool tests pass. Waiting for the exact-head native Windows rerun before merge.

duolahypercho:
Follow-up correction is now on the branch at `042efefa`: the earlier Windows serialization workaround has been removed after the same-source failure counts disproved runner starvation (27/33/27 failures before/after serialization). The actual boundary bug remains fixed: the long-lived ACL worker program is now delivered with PowerShell `-EncodedCommand`, leaving stdin exclusively for its base64 request protocol. Normal parallel `npm test` is restored. Local verification: `npm run check` plus the focused file-security/provider-key suite, 51 passed and 5 Windows-only assertions skipped on macOS. Exact-head CI is the remaining merge gate.

duolahypercho:
Closing this maintainer continuation without merging it. #477 remains the canonical contributor-owned PR. The exact Windows failure, support-bundle privacy boundary, and secret-verifier findings are now documented directly on #477 so the contributor can update their branch. I will wait for that update, test its exact head, and merge #477 directly only after all required checks pass.

PR #495, opened 2026-08-27
fix(antigravity): require operator-owned verified sessions

### LINKED ISSUE #393: Antigravity subscription quota unusable: secret unobtainable, docs give no acquisition path

duolahypercho:
Closes #393.

This is the safe replacement for the Antigravity credential/quota path discussed in #441. The #441 contributor history is preserved as this commit's second parent, but the unsafe vendor-client credential reuse and implied compatibility are not adopted.

Safety boundary:
- requires an operator-owned Google OAuth Desktop client; no bundled/vendor client impersonation and no borrowing another tool's credential store;
- binds an OS-assigned IP-literal loopback callback and fences OAuth, refresh, project discovery, disconnect, and activation races durably;
- keeps the provider withdrawn until an explicit quota-consuming probe proves one exact STOP/`OK` response, with separate consent for project provisioning;
- invalidates proof on account/client/session changes and refuses ambiguous refresh outcomes;
- keeps support bundles log-free and credential metadata bounded/redacted;
- owns descendant process trees across POSIX/Windows and preserves rollback barriers within finite shutdown/restart epochs;
- republishes installed clients only after forwarder readiness and rolls back coherently on failure.

Local verification on the exact final tree:
- focused cross-boundary suite: 306 passed, 4 platform skips, 0 failed;
- full Node suite: 3,020 total, 3,000 passed, 20 platform skips, 0 failed;
- native Swift tray suite: 101/101 passed;
- `npm run check` and `git diff --check` passed;
- no live provider, OAuth, provisioning, or quota-consuming call was made.

Why draft: native Windows Job/Task Scheduler behavior still needs the exact-head Windows CI runner. Do not merge until Linux, macOS, Windows, Electron, Swift, discovery, and security checks are green.


duolahypercho:
Review update: native Windows CI exposed and now covers two defects before merge: the Job owner payload omitted windowsHide, and the 250 ms test budget expired before the PowerShell Job owner could start its target. The production payload is fixed and the Windows-only test now waits long enough to prove descendant termination. CodeQL also found fast OAuth-secret-derived hashes; refresh fences now use a versioned memory-hard verifier, old in-flight journals fail closed before provider dispatch, lease path routing uses a non-credential non-cryptographic identity, and the project cache keys directly by its validated router UUID. Focused Antigravity coverage passes 84/84 runnable tests, plus 78/78 process-tree/Control Center tests locally. Exact-head CI is rerunning.

duolahypercho:
The latest native Windows run exposed test-fixture parity issues, not a reason to weaken production checks: `0600` alone does not create an owner-only Windows DACL, so the Antigravity boundary correctly refused those fixtures. This head now creates test credentials through the same private atomic writer used in production and makes two command/path assertions platform-aware. Local `npm run check` and 88 focused lifecycle/forwarder/restart/publication tests pass (87 pass, one Windows-only skip). The exact-head Windows rerun is now in progress; merge remains blocked on it and on a final merge-forward from main.

duolahypercho:
The remaining Windows audit fixes are now on the branch at `b169940c`: the probe token fixture uses the real owner-only writer (removing test-order dependence), and the npm-install test now proves the exact Windows `cmd.exe /d /s /c` wrapper plus `windowsVerbatimArguments`, while retaining exact POSIX argv checks. Local verification: `npm run check`; 18 focused probe/dependency tests passed; the complete Antigravity forwarder file passed 18/18 in isolation. The PR remains draft/conflicted intentionally until the prerequisite maintenance heads are merged and this branch is merge-forwarded for exact-head cross-platform CI.

duolahypercho:
Closing this maintainer replacement without merging it. #441 remains the canonical contributor-owned PR. The operator-owned OAuth, account-binding, lifecycle, disconnect, expiry-refresh, and real macOS proof requirements are now stated directly on #441. I will wait for the contributor update and evaluate that exact head rather than replacing the contribution.

PR #496, opened 2026-08-27
feat: add safe generic OpenAI-compatible providers

### LINKED ISSUE #491: Generic providers: complete routed-model integration with explicit per-model request profiles
## Context

This follows up on #312 and the maintainer feedback there.

#404 and #406 have since landed the safe generic-provider descriptor/credential boundary and live model discovery/capability metadata. That addresses a large part of the original request, but both PRs intentionally stop short of publishing generic-provider models into the Codex catalog or routing them end to end.

The remaining use case is a heterogeneous OpenAI-compatible gateway: one transport endpoint can expose Gemini, Claude, Grok, Step, DeepSeek, etc., while each downstream model may require different request normalization.

## Request

Please consider completing the generic-provider path so a user can:

1. register an arbitrary OpenAI-compatible provider through the new generic-provider APIs;
2. discover or explicitly curate models behind it;
3. publish selected models into the routed Codex catalog;
4. route those models end to end through the existing request boundary; and
5. attach an **explicit, validated per-model `requestProfile` / capability profile** for compatibility behavior.

The important part is that compatibility should remain explicit and model-scoped. I am **not** asking for model-name/prefix heuristics.

## Concrete gateway use case

I use a single OpenAI-compatible gateway (NineRouter) that exposes multiple model families through one `/v1` endpoint.

Most routed models work with ordinary OpenAI-compatible behavior. Gemini is the main exception: a Codex tool schema can contain the client-internal JSON-Schema annotation:

```json
{
  "type": "object",
  "properties": {
    "value": {
      "type": "string",
      "encrypted": true
    }
  }
}
```

On the Gemini route behind this gateway, forwarding that annotation unchanged produces HTTP 400 from the strict upstream schema validator. In local compatibility testing, the otherwise-identical request succeeds once the internal `encrypted` annotation is removed before outbound dispatch.

This is the small residual patch I currently have to carry locally.

## Desired compatibility boundary

The maintainer feedback on #312 was correct that a global recursive removal of every property named `encrypted` is unsafe because `encrypted` could be legitimate user schema data.

A safer design would be something like:

- generic transport identity stays separate from model compatibility;
- trusted user-curated model metadata may opt into an existing closed/validated `requestProfile` (or another explicit capability profile);
- Gemini-specific normalization is applied only when that explicit model profile requires it;
- sanitization removes only known Codex/router metadata at a defined tool-schema boundary, rather than arbitrary user fields;
- no behavior is inferred from model names;
- discovery data alone cannot grant a security-sensitive request profile.

The current Antigravity/Gemini path already has scoped schema cleanup for strict Gemini schema handling; ideally the generic-provider integration could reuse the same explicit compatibility architecture rather than requiring a second user patch.

## Acceptance idea

A useful regression test would configure one generic OpenAI-compatible mock gateway with two curated models:

- model A: ordinary OpenAI-compatible behavior;
- model B: explicit Gemini-compatible request profile.

Then prove:

- both are published/routable through the generic provider;
- model A receives the caller tool schema unchanged;
- model B receives only the documented Gemini/Codex metadata normalization;
- a legitimate user-defined field named `encrypted` is not removed merely because of its name;
- no request profile is inferred from the upstream model ID or display name.

## Related work

- #312 — original custom endpoint / heterogeneous compatibility proposal (closed after useful maintainer review)
- #404 — generic OpenAI-compatible provider descriptors + confined credential/request boundary
- #406 — live model discovery and capability metadata

Thanks again for the architecture guidance on #312. The new generic-provider foundation looks like the right place to solve this without carrying provider-specific local patches.

duolahypercho:
Closes #491.

Adds operator-managed generic OpenAI-compatible providers without extending trusted built-in provider identities.

Security and lifecycle boundaries:
- only an explicit closed request-profile allowlist is accepted
- live discovery cannot mint compatibility profiles
- credentials stay in the protected credential store and are injected only after endpoint, DNS, origin, and redirect validation
- generic CRUD applies catalog publication and service restart transactionally, with rollback
- doctor, health, and support output remain sanitized
- disabled or unavailable credentials fail closed with 503

Verification before opening:
- npm run check
- 78 focused generic routing, registry, curation, picker, capability, schema, doctor, and support-bundle tests passed
- no live provider or quota-consuming calls were made

This is draft-gated on the exact-head Linux, macOS, and Windows CI matrix.

### DIFF
--- a/src/api-forwarder.mjs
+++ b/src/api-forwarder.mjs
@@ -18,6 +18,7 @@ import {
   API_MODELS,
   MODEL_BY_GATEWAY_ID,
   PROVIDERS,
+  RUNTIME_PROVIDERS,
   providerForModel,
   endpointForModel,
   resolveProviderBaseUrl,
@@ -50,6 +51,9 @@ import {
   createResponsesStreamTransform,
   normalizeOpenAIRequest,
 } from "./openai-adapters.mjs";
+import { stripCodexEncryptedSchemaAnnotation } from "./tool-schema-root.mjs";
+import { requestGenericProvider } from "./generic-providers.mjs";
+import { genericProviderConfigured } from "./generic-provider-readiness.mjs";
 
 installStableFetchTransport();
 
@@ -304,6 +308,7 @@ function ensureToolResultsForCalls(messages) {
 const GEMINI_THOUGHT_SIGNATURE_SENTINEL = "skip_thought_signature_validator";
 
 function isGeminiProvider(provider, model) {
+  if (provider?.generic === true) return false;
   if (provider?.id === "gemini-api" || provider?.ownedBy?.toLowerCase?.() === "google") {
     return true;
   }
@@ -316,6 +321,29 @@ function isGeminiProvider(provider, model) {
     model?.gatewayModel,
     model?.model,
   ].some((value) => typeof value === "string" && value.toLowerCase().includes("gemini"));
+}
+
+function stripEncryptedToolSchemaAnnotations(payload, protocol) {
+  if (!Array.isArray(payload.tools)) return;
+  let changed = false;
+  const tools = payload.tools.map((tool) => {
+    if (!tool || typeof tool !== "object" || Array.isArray(tool) || tool.type !== "function") {
+      return tool;
+    }
+    if (protocol === "openai-responses") {
+      const repaired = stripCodexEncryptedSchemaAnnotation(tool.parameters);
+      if (repaired === tool.parameters) return tool;
+      changed = true;
+      return { ...tool, parameters: repaired };
+    }
+    const fn = tool.function;
+    if (!fn || typeof fn !== "object" || Array.isArray(fn)) return tool;
+    const repaired = stripCodexEncryptedSchemaAnnotation(fn.parameters);
+    if (repaired === fn.parameters) return tool;
+    changed = true;
+    return { ...tool, function: { ...fn, parameters: repaired } };
+  });
+  if (changed) payload.tools = tools;
 }
 
 // A trailing model turn is a destructive rewrite: it discards part of the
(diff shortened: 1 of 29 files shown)


duolahypercho:
Local verification update: the complete suite passed with controlled file concurrency: 2,877 tests, 2,860 passed, 17 platform/integration skips, 0 failures. The default maximum-concurrency invocation exposed an existing installer-test fixture race where several subprocesses append to one wrapper log; that test passes in isolation and this PR does not change the installer path. Exact-head GitHub Linux, macOS, and Windows CI remains authoritative. No live provider or quota-consuming calls were made.

duolahypercho:
Reconciled this branch with current main and fixed two end-to-end defects found during the maintenance audit:

- runtime generic providers now satisfy the router enablement gate, so their registered models reach the API forwarder instead of returning a built-in-selection 409; missing generic credentials continue to reach the API boundary and return the intended 503
- a generic provider whose valid ID is google no longer inherits the built-in Google trailing-user-turn rewrite; generic behavior remains explicitly model-scoped

Regression coverage exercises router routing with an empty built-in selection, health dependency probing for a generic-only runtime, missing-credential behavior, and preservation of trailing assistant history through both the router and API forwarder. Focused tests and npm run check pass. The controlled local full suite completed 2,900 tests with one unrelated POSIX wrapper-log fixture race; that exact test passed on isolated rerun. Exact-head Linux, macOS, Windows, Electron, native macOS, Python-lock, discovery, GitGuardian, and all CodeQL jobs are green on ae5b510.

Windows: https://github.com/duolahypercho/codex-router/actions/runs/33150964974/job/98782591757
Swift CodeQL: https://github.com/duolahypercho/codex-router/actions/runs/33150962104/job/98782584992

PR #497, opened 2026-08-27
fix: record state ownership before installing the service

mayrazing:
## Why

Reinstalling over a state directory owned by a different checkout
(`foreign_state_owner`) deadlocked on every retry, and the only escape was
deleting the whole state directory -- every stored provider key with it.

The deadlock chain, as observed on a real install:

1. `bin/install` runs with `MODEL_ROUTER_ALLOW_FOREIGN_STATE=1`, so the
   installer itself writes the generated catalog and gateway config fine
   (`state-owner.mjs` documents `bin/install` as the sanctioned ownership
   transfer).
2. The background service it installs does **not** get that override. On
   every boot, `start.mjs` calls `writeLiteLlmConfig()`, which
   `assertStateOwnership` refuses while `install-manifest.json` still names
   another checkout -- so the service crash-loops forever.
3. `service.mjs install` contains the readiness wait (up to 300 s), and the
   manifest was recorded only *after* it returned. Ownership therefore
   never transferred: the service cannot boot until the record updates, and
   the record cannot update until the service is healthy. The installer
   looks hung for the full budget, then rolls back, and the next retry
   repeats the cycle.

The crash loop also had no fast signal on Linux: with `Restart=always`, the
unit cycles back to "active" after every crash, so a crash-looping service
and a slow cold start are indistinguishable for the whole 300-second wait.

Triggering this requires having installed from two different checkouts
(e.g. a development checkout once, then the managed `install.sh` checkout)
-- exactly the cross-checkout scenario the ownership guard exists to manage
safely, and whose documented recovery ("reinstall from here") deadlocked.

## What

- **`bin/install` / `install.ps1`: record the manifest before the service
  step.** The generated state is already written by this checkout at that
  point, so the record is truthful, and the service boots against its own
  ownership instead of a stale one. The `--prepare-only` path is unchanged
  and still does not record. An install that previously deadlocked recovers
  by re-running the installer; no state migration is needed, since the next
  install overwrites the manifest.
- **`service-readiness.mjs`: fail fast on a Linux crash loop.** When the
  caller supplies a restart counter, the wait polls it alongside health and
  aborts once the count climbs 3 restarts above the value observed at wait
  start, with an error naming `journalctl --user -u codex-router.service`
  and the router log. The comparison is relative to the wait's own
  baseline, so residue from before the install cannot trigger it, and a
  failed query is inconclusive -- the wait stays health-only, which is the
  shape macOS launchd has (Windows already has its task-state guard).
- **`service-linux.mjs` / `service.mjs`: feed the counter to the guard.**
  New `restart-count` verb reading systemd's `NRestarts`; `service.mjs`
  injects it only for the Linux script.

## Test

- `test/installer-scripts.test.mjs` extends the manifest-ordering test:
  the record must now precede `service.mjs install` on both installers,
  not only the health wait.
- `test/service-readiness.test.mjs` gains five cases: the crash loop
  rejects early (well inside the budget) with the journal hint; a steady
  count and a high-but-static count both keep the wait alive; a throwing
  counter cannot fail the wait; the threshold is measured against the
  wait's own baseline. A source-level test pins the `restart-count` verb
  and the `service.mjs` wiring.
- `npm run check` passes; `sh -n bin/install` passes.
- `npm test`: 2824/2854 pass. The 12 failures reproduce on the unmodified
  branch and are environmental: most disappear under `umask 077` (fixture
  directories created 0775), one file needs the not-installed `playwright`
  devDependency, and two are source-root-discovery / `rmSync` quirks of
  this machine. None of the failing files import the changed modules.

## Design notes

- **Transfer semantics are unchanged.** `bin/install` was already the
  sanctioned ownership transfer: it runs with the override and rewrites the
  generated state either way. This PR only reorders the steps inside that
  same run, so the record lands before the service that depends on it. The
  guard's behavior on the unattended paths -- service boot, non-override
  rewrites -- is untouched.
- **Output surface is unchanged.** The readiness error names the state
  directory path and the unit name, the same identifiers doctor already
  reports. The `restart-count` verb emits a single integer, and
  `install-manifest.mjs record` output stays suppressed on both platforms,
  as before.
- **systemctl is invoked the same way as the existing verbs**: fixed argv
  arrays, a constant unit name, no shell. Unparseable counter output
  degrades to "unknown".
- **The guard degrades to the previous behavior.** With no counter
  available -- no systemd session, a non-zero exit, unparseable output --
  the wait is health-only, which is what this platform did before. A query
  failure can never fail an install, matching the existing Task Scheduler
  precedent on Windows.
- **A record failure now stops earlier.** If recording the manifest fails,
  the install aborts before the service step, so no service is started into
  a state it cannot boot in.

Compatibility: `systemctl --value` needs systemd >= 236 (2018). Older
builds yield an empty value, which reads as "unknown" and only disables the
early abort.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+- **Reinstalling over a state directory owned by another checkout no longer
+  deadlocks.** The installers recorded the state directory's new owner only
+  after the background service reported healthy, while the service itself
+  refuses to boot for as long as the record still names another checkout
+  (`foreign_state_owner`): an install that followed one from a different
+  checkout crash-looped for the full 300-second readiness budget, rolled
+  back, and repeated on every retry, and deleting the whole state directory
+  -- every stored provider key with it -- was the only escape. The record now
+  precedes the service step it describes, so the service boots against its
+  own ownership. Linux readiness also fails fast once the service manager's
+  restart counter shows a crash loop, instead of waiting out the whole
+  budget, and names the journal and the router log in the error.
+
 - **OpenCode Go Kimi K2.7 Code now accepts current Codex tool schemas.** Its
   Moonshot-backed validator receives the same bounded decorated-`$defs` repair
   as the first-party Kimi routes. The compatibility gate names only this
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,16 @@
   back, and repeated on every retry, and deleting the whole state directory
   -- every stored provider key with it -- was the only escape. The record now
   precedes the service step it describes, so the service boots against its
-  own ownership. Linux readiness also fails fast once the service manager's
-  restart counter shows a crash loop, instead of waiting out the whole
-  budget, and names the journal and the router log in the error.
+  own ownership. The ownership override the installers run under is scoped
+  to that full, ownership-transferring install on both platforms: a
+  prepare-only run meets the guard like any other writer (and the Windows
+  installer restores the caller's environment when it finishes). Linux
+  readiness also fails fast once the service manager's restart counter shows
+  a crash loop, instead of waiting out the whole budget, and names the
+  journal and the router log in the error; the counter query itself is
+  killed inside a bounded slice of the remaining budget, so a blocked
+  `systemctl` cannot stretch the wait, and a health answer that lands as the
+  threshold is crossed still wins over the crash-loop verdict.
 
 - **OpenCode Go Kimi K2.7 Code now accepts current Codex tool schemas.** Its
   Moonshot-backed validator receives the same bounded decorated-`$defs` repair
--- a/bin/install
+++ b/bin/install
@@ -8,12 +8,6 @@ adopt_native_catalog=false
 force_deps=false
 target=${MODEL_ROUTER_TARGET:-codex}
 
-# Installing is the sanctioned way for a checkout to take over a state
-# directory: generated files are rebuilt here and the new owner is recorded at
-# the end, so the ownership guard must not block this run.
-MODEL_ROUTER_ALLOW_FOREIGN_STATE=1
-export MODEL_ROUTER_ALLOW_FOREIGN_STATE
-
 case "$target" in
   codex|dsh|gemini) ;;
   *) echo "MODEL_ROUTER_TARGET must be codex, dsh, or gemini." >&2; exit 2 ;;
@@ -28,6 +22,18 @@ for argument in "$@"; do
     *) echo "Usage: $0 [--prepare-only] [--migrate-known] [--adopt-native-catalog] [--force-deps]" >&2; exit 2 ;;
   esac
 done
+
+# Installing is the sanctioned way for a checkout to take over a state
+# directory: generated files are rebuilt here and the new owner is recorded at
+# the end, so the ownership guard must not block this run. The override is
+# scoped to that full install, and only after the arguments are known: a
+# --prepare-only run rewrites the same generated state but exits before the
+# manifest record, so letting it past the guard would let a second checkout
+# rebuild foreign-owned state with no ownership transfer ever recorded.
+if [ "$prepare_only" != true ]; then
+  MODEL_ROUTER_ALLOW_FOREIGN_STATE=1
+  export MODEL_ROUTER_ALLOW_FOREIGN_STATE
+fi
 
 if [ "$prepare_only" = true ] && [ "$migrate_known" = true ]; then
   echo "--prepare-only cannot migrate an active older router." >&2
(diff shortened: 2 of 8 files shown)


duolahypercho:
Correction to my prior maintainer note: I should not have opened a separate continuation around your active PR. I am closing #498 without merging it, and #497 remains the canonical implementation and review thread. I am sorry for displacing your ownership of the fix.

Your core diagnosis is valid: the new checkout manifest must be recorded before the service boots, because service startup performs guarded state writes and otherwise waits on ownership that cannot transfer.

Please address these contained blockers on this PR before merge:

1. Scope the foreign-state override to a full ownership-transferring install. On POSIX it is currently exported before --prepare-only is parsed, so a prepare-only run can rewrite foreign-owned generated state without ever recording the transfer. On Windows, CheckoutInstall needs the equivalent scoped override for a full install and must restore the caller environment in finally.
2. Bound or remove the Linux restart-counter add-on. service.mjs runs the counter through synchronous spawnSync on every poll with no timeout, so a blocked systemctl call can exceed the outer 300-second readiness deadline.
3. Remove the health/restart threshold race. Health can become successful after the poll race returns null but before the restart query reaches the threshold; the current code can then throw crash-looping even though the router is already healthy.
4. Add regressions for POSIX prepare-only refusing foreign ownership, Windows full cross-checkout transfer plus environment restoration, a counter query that cannot overrun the deadline, and health winning when it settles at the restart threshold.

The manifest-before-service part is accepted. Once you update this branch, I will rerun focused installer/readiness tests and exact-head cross-platform CI, then merge it here if everything is green.

mayrazing:
Processing...

mayrazing:
Thanks for the clarification, and for keeping #497 as the canonical thread.

All four blockers are addressed on the branch:

1. Scoped foreign-state override.
POSIX now exports MODEL_ROUTER_ALLOW_FOREIGN_STATE only after argument parsing and only when --prepare-only is not set, so a prepare-only run meets the ownership guard like any other writer and can no longer rebuild foreign-owned generated state without the transfer being recorded.
Windows is a different story worth noting: CheckoutInstall had no override at all, so a full cross-checkout install was refused at the catalog step and the transfer path the manifest record serves could never run there. It now sets the override only for a full install, in place before the generated state is rebuilt, and restores the caller's environment in the outer finally — a pre-existing value goes back verbatim, an absent one is removed.

2. Bounded restart-counter query.
The readiness guard passes its remaining budget to getServiceRestarts(remainingMs), and the query runs with timeout: min(5s, remainingMs) and killSignal: SIGKILL. A blocked systemctl now costs at most the remaining budget and reads as an inconclusive count; no query is spawned once the budget is spent.

3. Health/restart threshold race removed.
When the counter crosses the crash-loop threshold, the wait gives health one last bounded look (Promise.race against one pollMs tick) before reporting the crash loop. A health answer that settles inside the threshold-crossing query returns healthy; the crash-loop error is only raised once health has had that final chance and stayed unsettled.

4. Regressions added.

Installer assertions: POSIX export guarded on a full install and ordered after --prepare-only parsing (plus a sh -n pass); Windows override set only when -PrepareOnly is off, in place before the catalog step, and restored in the outer finally on both the success and failure paths.
Readiness behavior tests: health winning as the threshold is reached (the exact interleaving that used to throw crash-looping for a healthy router), each counter query receiving a budget inside the wait's own timeout with the wait ending at its deadline, alongside the existing early-verdict, baseline, and inconclusive-count cases.
Verification: npm run check passes. The full suite is green except 12 pre-existing failures in desktop-panel/tray-rebuild that come from this Linux machine's group-writable umask (test-created bundle directories land at 775 and the macOS-only permission guard refuses them) — no overlap with this branch. The suites around the change (service-readiness, installer-scripts, service-lifecycle, state-owner) all pass. CHANGELOG updated to describe the scoping and the bounded query.

Ready for your focused installer/readiness rerun.

12duolalalala:
Verified on exact head `512701c` after the blocker follow-up:

- CI green across test (ubuntu/macOS/Windows), Electron (Linux/Windows), Unified native macOS (SUCCESS; skipped fine)
- `node scripts-check.mjs` pass; `sh -n bin/install` pass
- Focused: `test/installer-scripts.test.mjs` + `test/service-readiness.test.mjs` (47 pass / 1 skip), plus `service-lifecycle` and `state-owner`
- Live on this Linux box: `--prepare-only` against a foreign manifest refuses the catalog write and does not transfer ownership; a full `bin/install` against foreign state records the new owner (`sourceRoot` + commit `512701c`) before `service.mjs install`. Full systemd bring-up cannot run here (`systemctl` missing); `restart-count` correctly returns inconclusive without a unit. Windows/macOS service paths covered by CI, not live here.

Merging.

PR #498, opened 2026-08-27
fix: safely transfer cross-checkout state ownership

duolahypercho:
Continues #497 from @mayrazing while preserving the contributor commit as this branch's direct parent.

The reported deadlock is real: the service performs a guarded gateway rewrite during boot, so recording the new checkout only after service installation makes cross-checkout installs wait on a service that cannot become healthy.

This maintainer continuation keeps that core ordering fix and adds the missing parity/safety boundaries found during review:

- Windows CheckoutInstall temporarily applies the same sanctioned foreign-state override as POSIX before guarded generated writes, then restores the caller's environment in finally.
- Prepare-only runs no longer receive an installer-created ownership override, because they do not record an ownership transfer.
- The separable Linux restart-counter add-on is removed for now: its synchronous unbounded systemctl polling can exceed the readiness deadline, and a reproduced health/counter race can reject an already healthy service.

Local verification: sh syntax; npm run check; installer/state-owner/readiness tests pass (50 applicable tests, 1 PowerShell-only skip on macOS). Exact-head CI is required before merge.

Contributor credit and original investigation remain in ancestry through #497.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,9 @@
   back, and repeated on every retry, and deleting the whole state directory
   -- every stored provider key with it -- was the only escape. The record now
   precedes the service step it describes, so the service boots against its
-  own ownership. Linux readiness also fails fast once the service manager's
-  restart counter shows a crash loop, instead of waiting out the whole
-  budget, and names the journal and the router log in the error.
+  own ownership. Windows now applies the same explicit ownership-transfer
+  override as the POSIX installer, while preparation-only runs remain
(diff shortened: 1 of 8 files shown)


duolahypercho:
Closing this maintainer continuation without merging it. #497 remains the canonical contributor-owned PR. I opened #498 to preserve ancestry while applying review fixes, but that still displaced the active contributor and was the wrong collaboration choice. The actionable findings are now documented directly on #497, where I will wait for the contributor update and then retest the exact head.

PR #502, opened 2026-08-28
fix: preserve Windows foreign-state override

PracticalSwan:
## Summary

Preserve the caller's `MODEL_ROUTER_ALLOW_FOREIGN_STATE` value across every Windows installer path, including `-PrepareOnly` and failures that occur before the full-install ownership override is set.

## Root cause

#497 initialized `$HadForeignStateOverride = $false` / `$SavedForeignStateOverride = $null`, then captured the real caller environment only inside `if (-not $PrepareOnly)` deep in the main `try`. The outer `finally` always restores/removes the variable.

That means:
- a successful `-PrepareOnly` run never captures the caller value, then `finally` removes it;
- any failure before the capture block does the same.

This violates the installer's stated contract that the caller environment is restored exactly as found.

## Fix

Snapshot both presence and value before `Push-Location` / the outer `try`. The full-install branch now only sets the temporary value to `1`; the existing `finally` restores the pre-run snapshot.

POSIX needs no matching change because its exported override exists only in the installer child shell and cannot mutate the caller environment.

## Verification

- red/green regression: the ownership test failed before the fix because the snapshot occurred after the outer `try`, then passed after moving the snapshot;
- focused installer group (`install.ps1` parse, rollback, manifest ordering, foreign-state override) — **4 passed, 0 failed**;
- `npm.cmd run check` — pass (`v2-agent applications valid (6)`, syntax checks passed);
- direct PowerShell parser check — pass;
- `git diff --check` / staged diff check — clean.

I also attempted the entire installer policy file with Git's POSIX shell manually added to this Windows PATH. One unrelated POSIX fixture test became active and failed because that artificial PATH changes its platform assumptions; the changed regression itself passed. I did not broaden this PR to that separate test issue (#500 already addresses the native-Windows shell portability regression).

No service mutation, provider request, or quota-consuming check was used.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -464,7 +464,11 @@ try {
 
   if ($PrepareOnly) {
     Write-Host "Dependencies and generated files are prepared; application configuration was not changed."
-    exit 0
+    # Return from the script instead of terminating the caller's PowerShell
+    # host. The outer finally still has to restore the caller environment, and
+    # an invoked prepare-only install must hand control back so its caller can
+    # observe that restoration.
+    return
   }
 
   $ConfigEnabled = $true
--- a/install.ps1
+++ b/install.ps1
@@ -243,11 +243,11 @@ $ConfigDisableCommand = if ($Target -eq "codex") { "disable" } else { "uninstall
 $ConfigEnabled = $false
 $ServiceInstalled = $false
 $AdoptionPending = $false
-# The foreign-state override below is set inside the try and must be undone in
-# the finally even when a step throws first, so both halves of the caller's
-# environment are captured up front.
-$SavedForeignStateOverride = $null
-$HadForeignStateOverride = $false
+# The foreign-state override below is set only for a full install, but the
+# finally runs for prepare-only and for failures that happen before that point
+# too. Snapshot the caller's environment before any installer step can throw.
+$HadForeignStateOverride = $null -ne (Get-Item Env:\MODEL_ROUTER_ALLOW_FOREIGN_STATE -ErrorAction SilentlyContinue)
+$SavedForeignStateOverride = $env:MODEL_ROUTER_ALLOW_FOREIGN_STATE
 $ConfigWasEnabled = $false
 $ServiceWasInstalled = $false
 $TrayWasInstalled = $false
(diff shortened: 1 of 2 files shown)


12duolalalala:
Holding — `test (macos-latest)` is red on this head (Windows/Ubuntu/Electron/native macOS green).

What failed: `POSIX install and enable preserve the PATH-selected Node wrapper` in `test/installer-scripts.test.mjs` (assertion at ~line 294). The CI PATH on that runner includes a space (`…/runtime bin % wrapper/…`), and the install/enable PATH capture did not match the expected wrapper selection. That test is unrelated to the `install.ps1` foreign-state snapshot change here.

Also: #500 already landed on main and touches the same foreign-state test; rebase onto current main before another CI push so the `POSIX_SHELL_AVAILABLE` guard and these snapshot assertions coexist.

Not merging on red CI. Cannot live-run Windows `install.ps1` on this Linux box, so that surface is unchecked here either.

PracticalSwan:
Updated the branch without history rewrite: merged current `main` (`10abced`, including #499–#501/#500) into #502 and pushed new head `20fd0f8`.

Post-merge verification on native Windows:
- `test/installer-scripts.test.mjs`: **24 passed / 0 failed / 8 intentional platform skips**
- `install.ps1` PowerShell parser: pass
- `npm run check`: pass
- `git diff origin/main...HEAD --check`: clean
- PR diff against current main remains only `install.ps1` + the foreign-state ownership regression.

The #500 `POSIX_SHELL_AVAILABLE` guard and #502 snapshot assertions merge cleanly and are both present. Fresh exact-head CI is now running. I have not changed the unrelated POSIX Node-wrapper path behavior from the prior macOS failure.

PracticalSwan:
Fresh exact-head verification is now fully green on `20fd0f8`: CI run #1378 completed **success**, Python-lock completed **success**, and provider model discovery completed **success**. This supersedes the earlier red-head macOS hold; the updated branch includes #500/current main and its macOS `npm test` passed.

12duolalalala:
CI on `20fd0f8` is now fully green and GitHub still says MERGEABLE after the #500 rebase. Isolated on this Linux box against that head:

- `test/installer-scripts.test.mjs`: 31 passed, 0 failed, 1 skip (`install.ps1` PowerShell parse — no pwsh here)
- the new pre-try snapshot assertions passed
- `node scripts-check.mjs` passed

Still not merging. The bug is specifically PowerShell caller-env restore on `-PrepareOnly` / early failure, and I cannot actually run `install.ps1`. The source-order test is not a live prepare-only restore. Main moved to `3a33772` (#503 then #504) since this head; files do not overlap, but rebase if you want another look. A Windows `-PrepareOnly` run that keeps `MODEL_ROUTER_ALLOW_FOREIGN_STATE` would unblock it.

duolahypercho:
Addressed the remaining live-runtime concern in 55d4cbd. `install.ps1 -CheckoutInstall -PrepareOnly` now returns to an invoking PowerShell host instead of terminating it, and the Windows-only regression invokes the real installer with shimmed dependencies to prove all three cases: a caller value is preserved on success, an absent value remains absent, and a caller value is restored after an early failure. Exact-head `test (windows-latest)` is green: https://github.com/duolahypercho/codex-router/actions/runs/33149394531/job/98777621282

PR #503, opened 2026-08-28
fix: avoid widening Moonshot ref siblings

PracticalSwan:
## Summary

Fail closed when Moonshot foreign-`$ref` inlining would overwrite a different validation/schema keyword from the referenced target.

## Root cause

`inlineForeignRefs()` expands non-`#/$defs/` refs for Moonshot-compatible routes. It merged the referenced target and the `$ref` node's siblings with object spread. That is lossless for distinct keywords and harmless annotation overrides, but `$ref` siblings are conjunctive: if both sides define the same validation keyword with different values, overwriting one side can widen the caller schema.

Concrete example on current main:

- target: `{ type: "string", maxLength: 5 }`
- sibling ref: `{ $ref: "#/properties/base", maxLength: 10 }`
- old output: `{ type: "string", maxLength: 10 }`

The rewritten tool now accepts strings of length 6–10 that the original schema rejected.

#490 extends this shared Moonshot repair to the exact Console Go Kimi K2.7 Code route, so the edge now affects both the existing Kimi routes and that newly measured route.

## Fix

- allow normal merge when overlapping schema keywords are identical;
- preserve known annotation overrides such as `description`, `default`, `title`, and examples;
- if a foreign ref and its target provide different values for the same non-annotation keyword, keep the original `$ref` node and let the strict provider fail closed rather than altering validation semantics;
- leave the existing stricter `$defs` conflict behavior unchanged.

## Verification

Red/green regression:
- before: the new test observed the widened `{ maxLength: 10 }` output and failed;
- after: the conflicting foreign ref is preserved by identity.

Post-merge verification against current main `10abced`:
- `test/tool-schema-root.test.mjs` — **45 passed, 0 failed**;
- targeted Kimi namespace-relay cases — **4 passed, 0 failed**;
- #490 exact Console Go Kimi route regression — **1 passed, 0 failed**;
- `npm.cmd run check` — pass (`v2-agent applications valid (6)`, syntax checks passed);
- `git diff --check` / staged diff check — clean.

The temporary dependency junction used for local tests was removed before commit; the canonical dependency tree remains intact. No provider or quota-consuming request was made.

### DIFF
--- a/src/tool-schema-root.mjs
+++ b/src/tool-schema-root.mjs
@@ -239,6 +239,16 @@ export function nonRecursiveToolSchema(schema) {
 // list. Copy-on-write throughout: a schema with no foreign ref keeps identity,
 // and the client's object is never mutated.
 const DEFS_REF_PREFIX = "#/$defs/";
+const REF_OVERRIDE_ANNOTATIONS = new Set([
+  "$comment",
+  "default",
+  "deprecated",
+  "description",
+  "examples",
+  "readOnly",
+  "title",
+  "writeOnly",
+]);
 const MAX_INLINE_DEPTH = 32;
 const MAX_INLINE_EXPANSIONS = 512;
 const MAX_INLINE_BYTES = 256 * 1024;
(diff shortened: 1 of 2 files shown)


PracticalSwan:
Exact-head verification on `3e4e272` is fully green: CI run #1377 completed **success** across the Linux/Windows/macOS test matrix, Electron packaging/smokes, native macOS app, and Homebrew-formula; provider model discovery also completed **success**. No review threads or follow-up failures are open.

PR #506, opened 2026-08-28
Make full desktop install the default

duolahypercho:
## Summary

- put the one-command full desktop install at the top of the README
- install the Control Center, tray/menu-bar app, and macOS desktop-widget capability by default through the recommended installer
- keep Homebrew CLI/router-only and refuse desktop builds inside Homebrew-managed installs
- document the unavailable Antigravity client-secret acquisition path from issue #393
- document that this project does not yet publish an npm-installable CLI
- prepare v0.5.1 and the Homebrew Core submission workflow

## Verification

- `npm run check`
- full Node suite: 2,736 passed, 17 skipped, 0 failed
- Control Center typecheck, build, and tests: 55 passed
- focused installer/release/Homebrew tests: 58 passed, 2 skipped
- `packaging/homebrew/check-core-readiness.sh`
- formula regeneration drift check and Ruby syntax check

AI assistance was used to implement and review this change; the maintainer reviewed the resulting diff and verification evidence.

PR #508, opened 2026-08-28
fix(homebrew): restore v0.5.0 source checksum

duolahypercho:
Fixes the checksum regression merged by #506.

The formula URL still points to the published v0.5.0 tarball, whose SHA-256 is:
`89f494a21399c3c2a54f6f25531e10b0d64e7e5a238f17cb8f8f7f2c9222e076`.
The merged formula had `89f676…`, so a real Homebrew download fails verification.

Verification:
- downloaded the release asset and computed the SHA-256
- `npm run check`
- 61 focused Homebrew/release/installer tests passed, 2 platform skips
- `git diff --check`


duolahypercho:
Superseded by current main (3a201896), which has since published v0.5.1 and updated Formula/codex-router.rb to the v0.5.1 URL and verified checksum. This v0.5.0-only hotfix is conflicting and is no longer needed.

PR #516, opened 2026-08-28
feat: add caller capability rotation

### LINKED ISSUE #513: Add a supported transactional rotation path for the local router caller capability
## Problem

The router's caller key is a local capability and can be embedded into client-facing managed URLs. The repository correctly treats it as sensitive and keeps it out of normal status/support output, but there is currently no supported way to rotate it after suspected exposure.

`src/secret.mjs` exposes only `ensure|status`; `ensure` deliberately preserves an existing valid caller secret. The practical recovery path today would therefore require manual mutation/removal of the caller-secret file followed by regeneration, client republishing, and a service restart. That is too easy to perform incompletely and can strand running clients on the old capability.

This surfaced during Windows/Codex compatibility maintenance where an existing capability was found in historical local client session artifacts. The artifact contents and capability are intentionally not included here. The important repository-level gap is simply that exposure cannot be remediated through a supported transaction.

## Desired behavior

Provide an explicit operator action such as `model-router codex caller-key rotate` (exact naming flexible) that:

1. Generates a fresh caller capability without printing it.
2. Republishes every installed client target that embeds/references the caller capability (Codex, DSH, Gemini as applicable), preserving unrelated user configuration.
3. Rebuilds/restarts the managed router service so the new capability is authoritative.
4. Verifies health and that the old capability is rejected before reporting success.
5. Rolls back coherently if republishing/service activation fails, rather than leaving clients and service on different keys.
6. Preserves the internal service key unless that key is also explicitly being rotated.
7. Never writes either old or new capability to logs, command arguments, support bundles, or terminal output.
8. Clearly warns that already-running clients/tasks may need to be restarted/reopened because they can cache the previous managed URL.

## Why this should be explicit

Rotation is consequential and should not happen during ordinary install/update/doctor. A stable caller capability prevents needless client churn. But once an operator has reason to consider it exposed, there should be a first-class recovery operation instead of undocumented file deletion.

## Suggested coverage

- Successful rotation updates service + all installed client surfaces atomically enough that they agree on the new key.
- Old capability is rejected after success.
- New capability is accepted.
- Failure during client republish/service install restores the previous usable state.
- No capability appears in stdout/stderr or serialized diagnostics.
- Windows file ACL remains current-user-only after rotation.
- Running-client warning is emitted without terminating Codex automatically.

PracticalSwan:
## Summary

Add a supported, non-printing transaction for rotating the router's local caller capability after suspected exposure.

`caller-key rotate` now:

- acquires a dedicated rotation lock and serializes the full transaction/recovery against model-overlay, catalog-refresh, and the existing shared service-operation lifecycle lock;
- writes a protected phase journal containing hashes only, never either caller capability;
- refuses partial or residual caller-bearing Codex / DSH / Gemini managed state before mutating the service or secret;
- stops the router before the swap only when it was actually running, and preserves an installed-but-stopped state;
- keeps one protected rollback generation until the transaction is verified and committed;
- refreshes only caller URL/key fields in already-installed Codex / DSH / Gemini integrations, preserving model lists, defaults, comments, provider ownership, and other client policy;
- validates the fresh capability with an HTTP 200 OpenAI-style model-list body and requires the stale capability to return exactly HTTP 401;
- restores the prior private generation and client/service state after publication, restart, ACL, or verification failures;
- validates rollback integrity/privacy before any recovery service mutation;
- on interrupted recovery, observes the *current* service state under lifecycle ownership so a service started in the post-crash/pre-recovery window is stopped, restored to the prior generation, restarted, and verified coherently;
- routes ordinary Windows `start` through the locked managed-service path and routes explicit POSIX/Windows foreground starts through `src/foreground-start.mjs`, which acquires lifecycle ownership before `start.mjs` module evaluation and keeps it for the foreground supervisor lifetime;
- preserves the separate internal service key and never prints either caller capability.

Commands:

```sh
./bin/model-router codex caller-key rotate
```

Windows:

```powershell
./codex-router.ps1 caller-key rotate
```

## Verification

Exact head: `9d49e99ca62d5c55ac1e431407acc24d7e466eb8`.

The August 29 maintainer blockers are covered by deterministic regressions for:

- post-swap crash followed by a service start before recovery;
- corrupt/unprotected rollback material causing zero service mutations;
- foreground supervisor overlap with an existing lifecycle owner;
- Windows normal start using the managed service and `--foreground` using the locked foreground launcher.

Fresh local Windows evidence:

- focused recovery/start lifecycle set: 25 total, 23 passed, 2 POSIX-only skips, 0 failures;
- broader isolated caller/security/service/Windows set under a temporary `MODEL_ROUTER_STATE_DIR`: 102 total, 82 passed, 20 platform-inapplicable skips, 0 failures;
- `npm run check`: passed;
- PowerShell parser: passed;
- `npm audit --audit-level=high`: 0 vulnerabilities;
- no provider/model inference calls were used.

Exact-head GitHub CI is fully green across Ubuntu, macOS, Windows, Electron Linux/Windows, unified native macOS, deterministic provider discovery, Homebrew formula, and GitGuardian. Live provider audit and Homebrew install were skipped by workflow conditions.

The installed Windows checkout carries the final hardening as local commits `b04fe25` and `7d0dfa4` on top of the earlier reviewed caller-rotation adaptations. The router was restarted after the runtime change and remains healthy; caller/internal capability ACLs remain current-user-only and no rotation residue is present.

Closes #513


PracticalSwan:
Fresh exact-head verification is fully green for `be78193b0b3b7f13ec0940368f416f26c0e13735`: Ubuntu, macOS, Windows, Electron Linux/Windows, unified native macOS, deterministic provider discovery, Homebrew formula, and GitGuardian all completed successfully. The live provider catalog audit and Homebrew install jobs were skipped by their workflow conditions. Windows `npm test` completed successfully in 14m05s, consistent with the repository's recent CI duration.

duolahypercho:
Maintainer exact-head review of `be78193`: keeping this open; do not merge this rotation implementation yet.

The operation has no cross-process lock or durable phase journal, so concurrent rotations can report success and later revoke the newly published capability. Additional fail-open paths remain around ACL failure, partial Gemini/Codex managed-state detection, rollback generations, and verification: the old key check accepts arbitrary non-200 responses instead of exact 401, while the fresh response body is not validated. Plain Gemini/DSH installs also change existing publication/default policy.

The safe boundary is one dedicated operation lock plus protected phase journal, manager-native capability-only refresh, stop/swap/restart/verify, policy preservation, exact 200 valid-body + exact 401 checks, and deterministic failure/concurrency/Windows privacy coverage. The current-main conflicts are mechanical and do not resolve these security blockers.

duolahypercho:
Maintainer re-review of new head `3f281985bedfc747a4927abd9a8ffeeb46d2551c`: substantial improvement, but still **do not merge**.

Closed from the prior review: the dedicated operation lock/journal now exist; ACL failure rolls back; fresh verification requires a valid model-list body and the old key must return exact 401; capability-only refresh preserves existing policy.

Remaining blockers:

1. Fresh core CI is red on both Ubuntu and macOS. `test/caller-key-maintenance.test.mjs` omits the injected `secretPath` in two rotation tests, so `currentSecret()` reads the real default state path. A clean runner fails ENOENT; a populated developer home masks it.
2. `installedTargetsFromStatus()` still accepts key-bearing partial managed state. Gemini with a managed block but missing/unmanaged catalog state, and Codex native mode with managed caller URLs/blocks but missing provider state, can return no target. Rotation then leaves the old caller capability published.
3. The full rotation is not serialized with service lifecycle. If the service is initially stopped and a concurrent `start` occurs before the swap, rotation skips stop/restart and old-key verification, yet commits the new secret/client generation while the live router retains the old in-memory key. Pending-journal recovery has the same `serviceWasRunning: false` race.

I reproduced the clean-home ENOENT and the stopped-snapshot/concurrent-start generation split. The focused suite otherwise passes 118/118, root check passes, and the merge onto current main is mechanically clean; those facts do not close the security/lifecycle boundary.

Safe boundary: fail closed on every partial managed client permutation, and hold one lifecycle-aware ownership/lock across snapshot → stop → swap → refresh → restart/verify (including recovery), with a deterministic concurrent-start regression. Then rerun the complete platform matrix.

duolahypercho:
Exact-head follow-up for `109b4ffcabb8b33feeb2c2243e0d6c31f58338f9`: verdict is unchanged, **do not merge**.

The new delta is only 28 test lines for an unprotected rollback generation. It changes no production code and none of the two clean-home failures. The concurrent-start/service-generation split, asymmetric Codex/Gemini partial-state detection, and omitted `secretPath` test isolation all reproduce unchanged. The added rollback-privacy regression is useful but unrelated to those blockers.

PracticalSwan:
Maintainer blockers from the `be78193` review have been addressed at exact head `109b4ffcabb8b33feeb2c2243e0d6c31f58338f9`.

Key changes: dedicated cross-process rotation lock plus model-overlay/catalog-refresh serialization; protected hash-only phase journal; stop-before-swap with stopped-service preservation; protected non-overwritable rollback generations; exact valid 200 model-list + stale 401 verification; fail-closed partial Codex/DSH/Gemini state detection; and manager-native capability-only refresh that preserves publication/default policy.

Failure coverage now includes concurrency, ACL failure, crash recovery, idempotent recovery, verified-phase cleanup, rollback hash/privacy checks, and preservation of rollback+journal material when privacy cannot be proven. Fresh local caller/refresh/dispatcher verification on `109b4ff`: 30 total, 26 passed, 4 platform-inapplicable skips, 0 failures; `npm run check` is green. No provider/model inference calls were used.

I also adapted the same reviewed runtime into the installed Windows checkout and restarted only the router service. Its health is green with no degraded components; caller/internal keys remain current-user-only ACLs; there is no rotation residue; and the adjacent client-manager verification passed (Codex focused 6/6, DSH+Gemini 41 passed / 1 Windows skip / 0 failures).

Provider model discovery CI is green for `109b4ff`; the main CI workflow is still in progress, so this is an evidence update rather than a merge request.

PracticalSwan:
Exact-head update: `4f247dcd7fdcc4dd6daf8230744d14128e45884a` now closes the two production blockers from the `109b4ff` re-review and is merged with current upstream `main` (`102f008`).

- Partial managed state now fails closed when Codex is nominally native but still contains a managed caller URL/block (including provider-table `base_url`), and when Gemini still has its managed block without matching managed catalog/base-url state.
- Caller rotation and pending-journal recovery now hold the existing shared `service-operation` lock across service snapshot -> stop -> capability swap/restore -> client refresh -> restart/verification -> finalize. Service start/stop is executed through the same service implementation without recursively reacquiring the lock. Deterministic stopped-service concurrent-start regressions cover both rotation and recovery.

Fresh local evidence on the merged head: `npm run check` green; caller/security/service set 90 total / 66 passed / 24 platform skips / 0 failed; full `test/config-manager.test.mjs` 48 total / 47 passed / 1 intentional skip / 0 failed (the upstream merge did not change any of the six caller-rotation files); merged native custom-tool regressions 12/12. Live status classification was also checked using redacted booleans only and selects the installed Codex integration without exposing a capability. No provider/model inference calls were made.

Fresh exact-head CI is now running (CI `33220643771`, provider discovery `33220643836`). This is an evidence update, not a merge request.

PracticalSwan:
Exact-head CI is complete and fully green for `4f247dcd7fdcc4dd6daf8230744d14128e45884a`: Ubuntu, Windows, macOS, Electron Linux/Windows, unified native macOS, Homebrew formula, and deterministic provider discovery all succeeded. The live provider audit and Homebrew install jobs were skipped by their workflow conditions. The installed Windows checkout now carries the final hardening as local `bd7293c`; its focused caller/lifecycle and Codex config regressions are green, the restarted router is healthy with no degraded components, and protected local state remains intact. PR remains open/unmerged for maintainer review.

duolahypercho:
Maintainer re-review of exact head `4f247dcd7fdcc4dd6daf8230744d14128e45884a`: the new lifecycle lock, residual-state detection, and isolated tests close the three previously reported normal-operation failures, but one P1 crash-recovery race remains.

Reproduction:

1. The service is originally stopped, so the journal records `serviceWasRunning: false`.
2. Rotation swaps to the new caller secret and reaches `secret-swapped`, then the process crashes and releases its locks.
3. An operator starts the service before recovery; it loads the new secret.
4. Recovery acquires the service lock, restores the disk secret and client documents to the old generation, clears the journal, and reports `{ recovered: true, committed: false }`—but performs no service mutation because both stop and restart are conditional only on the historical `journal.serviceWasRunning` flag.

Observed result: `serviceMutations: []`, disk/client generation is old, running service generation is new, and the journal is cleared. The new recovery test covers a contender that starts *during* recovery, but not a service already started in the post-crash window.

The affected branches are `src/caller-key.mjs:206-209` and `:236-244`. Recovery needs to inspect the current service state while holding lifecycle ownership and either stop/restart and verify an already-running service, or fail closed without restoring/clearing. Add a deterministic post-swap crash → pre-recovery start regression.

Everything else checked on this head is strong: exact reconstructed tree matches GitHub; root check passes; 128 focused credential/client/service tests pass; full GitHub Linux/macOS/Windows/Electron/native/security matrix is green. This remaining generation split still precludes merge.

duolahypercho:
One additional platform-parity blocker from the independent pass: the new service-operation lock is not acquired by every supported start path.

- POSIX `bin/start --foreground` still directly execs `src/start.mjs` (`bin/start:16-20`).
- Windows `codex-router.ps1 start` still directly invokes `src/start.mjs` (`codex-router.ps1:900`), so the ordinary Windows start path bypasses `src/service.mjs` and its new lock.

Those paths can start a router generation during the rotation/recovery window despite lifecycle ownership. Windows is especially material because this is its default `start`, not only an opt-in debug path. Route the managed Windows start through the locked service layer, and make foreground starts either hold equivalent ownership for their lifetime or be detected/refused by rotation. Please add Windows and foreground overlap regressions alongside the post-crash/pre-recovery-start case.

PracticalSwan:
﻿Exact-head follow-up: `0481de1a5121a3d16b18764374bdf92927f0d669` addresses both remaining maintainer blockers from the August 29 re-review.

Crash recovery now validates rollback evidence before touching the service, then observes the *current* service state under lifecycle ownership. If a service was started after a stopped-service rotation crashed, recovery stops that live generation, restores/refreshes the prior generation, restarts, and verifies the displaced capability. Negative tests also prove corrupt/unprotected rollback material causes zero service mutations.

All supported starts now share the lifecycle boundary: ordinary Windows `start` routes through `service.mjs`; explicit POSIX/Windows foreground starts use a new launcher that acquires `service-operation` before importing `start.mjs` and holds it for the supervisor lifetime. A real overlap test proves foreground boot waits behind the lock, and a Windows node-shim execution test proves both dispatcher modes without touching the host service.

Local Windows verification: focused 25 total / 23 passed / 2 POSIX skips / 0 failed; broader isolated caller/service/Windows set 102 total / 82 passed / 20 skips / 0 failed; root check and PowerShell parse green; npm high-severity audit reports 0 vulnerabilities. No provider inference was used. Fresh exact-head CI is now the remaining evidence gate; this is not a merge request.

PracticalSwan:
Follow-up exact head is now `9d49e99ca62d5c55ac1e431407acc24d7e466eb8`. The only delta after `0481de1` is a one-line POSIX lifecycle expectation update: the pre-existing `bin/start --foreground` dispatch test now expects the new locked `src/foreground-start.mjs` launcher instead of direct `src/start.mjs`. This was caught during cross-platform review before relying on CI. No production code changed in this follow-up.

PracticalSwan:
﻿Exact-head verification is now complete for `9d49e99ca62d5c55ac1e431407acc24d7e466eb8`.

GitHub CI is fully green on Ubuntu, macOS, Windows, Electron Linux/Windows, unified native macOS, deterministic provider discovery, Homebrew formula, and GitGuardian. The live provider audit and Homebrew install jobs were skipped by their workflow conditions.

The installed Windows checkout also carries the two reviewed follow-up commits as local `b04fe25` + `7d0dfa4`. Fresh post-adaptation evidence: `npm run check` green; isolated caller/service/Windows set 64 total / 44 passed / 20 platform skips / 0 failed; PowerShell parse and diff checks green. The router service was restarted once and is healthy; Codex remains authenticated; config/caller/internal capability ACL checks are green; no rotation residue or recent router-log error signatures were found; protected local install override/stash/rollback state remains intact. No provider/model inference calls were used.

PR remains open/unmerged for maintainer re-review.

PR #525, opened 2026-08-29
fix: label failover model counts accurately

PracticalSwan:
﻿## Summary

`doctor` currently reports automatic failover as, for example, `3 of your own providers`, but the value comes from `failoverTierCounts()`, which increments once per routed model. That can directly contradict the adjacent `Enabled providers` row and makes a single provider with several routed models look like several providers.

This patch changes only the diagnostic wording:

- free tier present: reports the paid count as `model(s) on your own providers`;
- no free tier: reports the same model count accurately instead of calling it a provider count;
- adds a regression tying the doctor wording to the model-count contract.

No ranking, provider selection, cooldown, or failover behavior changes.

## Reproduction

On an installation with one enabled routed provider and three eligible paid models, `doctor` can currently show both:

- `Enabled providers: <one provider>`
- `Model failover: on, 3 of your own providers ...`

`failoverTierCounts()` and its existing tests confirm that `subscription` is a model count, not a distinct-provider count.

## Verification

- TDD regression observed failing before the wording change.
- `node --test test/model-router-cli.test.mjs test/model-failover.test.mjs` — 39 total: 35 passed, 4 Windows-inapplicable skips, 0 failures.
- `npm run check` — syntax checks passed; v2-agent applications valid (6).
- `git diff --check` — clean.

### DIFF
--- a/src/doctor.mjs
+++ b/src/doctor.mjs
@@ -614,8 +614,8 @@ if (!failoverSettings.enabled) {
     "ok",
     "Model failover",
     failoverCounts.free
-      ? `on, ${failoverCounts.free} free model(s) first then ${failoverCounts.subscription} of your own`
-      : `on, ${failoverCounts.subscription} of your own providers -- no free model is curated, so nothing cheaper is tried first`,
+      ? `on, ${failoverCounts.free} free model(s) first then ${failoverCounts.subscription} model(s) on your own providers`
+      : `on, ${failoverCounts.subscription} model(s) on your own providers -- no free model is curated, so nothing cheaper is tried first`,
     failoverCounts.free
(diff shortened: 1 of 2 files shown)


PracticalSwan:
Exact-head verification update for \3efeafd38d8c27a3e915f8a563eda9df2cd8457\: the full CI matrix is green — Ubuntu, macOS, Windows, Electron Linux/Windows, unified native macOS, deterministic provider discovery, Homebrew formula, and GitGuardian all succeeded (live provider audit/Homebrew install skipped by workflow conditions). The installed Windows checkout carries the diagnostic-only fix as local \543ac93\; focused CLI/failover regression checks and \
pm run check\ pass, and the live doctor now correctly says \3 model(s) on your own providers\ beside the single enabled provider. No routing/failover behavior changed and no provider inference calls were used.

PR #526, opened 2026-08-29
fix: report native vision engines in doctor

PracticalSwan:
## Summary

- make the Codex doctor evaluate the same gated native + routed vision-engine set used by the runtime/control paths
- count installed native vision engines only when the doctor's fresh Codex sign-in probe is authenticated
- preserve routed-only evaluation for non-Codex targets
- stop describing the built-in default engine as an operator pin when it is unavailable

## Reproduction

A signed-in Codex install can have `gpt-5.6-luna` present in both the merged and captured native catalogs while no routed provider exposes vision. Before this patch the doctor reports:

`OK Vision bridge: pinned engine gpt-5.6-luna is not an enabled model that reads images`

while the control/runtime resolver correctly resolves `gpt-5.6-luna`.

The inverse edge case matters too: after sign-out, those native catalog files can remain stale on disk while the request path refuses native vision without a live caller session. Doctor must not treat stale catalog membership alone as a usable engine.

## Verification

- isolated regression was RED on current main with the exact false diagnostic, then GREEN after the fix
- regression proves authenticated native Luna resolves, then the same stale catalogs are ignored after the live sign-in probe becomes signed-out
- adjacent doctor/vision/catalog set: 152 passed, 0 failed
- `npm run check`: passed
- `git diff --check`: clean
- repo-maintainer analyzer: local/targeted impact

No provider/model inference calls were used.

### DIFF
--- a/src/doctor.mjs
+++ b/src/doctor.mjs
@@ -544,7 +544,7 @@ add(
 // The same installed-native gate used by catalog/control keeps this diagnostic
 // aligned with what the running Codex route can actually resolve.
 const visionSettings = readVisionBridgeSettings();
-const visionCandidates = codexTarget
+const visionCandidates = codexTarget && codexAuth?.authenticated === true
   ? [...requiredRoutedModels, ...installedNativeVisionEngines({ hidden: readHiddenModels() })]
   : requiredRoutedModels;
 const visionEngine = resolveVisionEngine(() => visionCandidates, visionSettings);
--- a/src/doctor.mjs
+++ b/src/doctor.mjs
@@ -72,6 +72,7 @@ import {
   selectedConfiguredListedModels,
 } from "./provider-selection.mjs";
 import { resolveVisionEngine } from "./vision-bridge.mjs";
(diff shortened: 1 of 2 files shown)


PracticalSwan:
Exact-head verification update for \ccf83f29670b8bdceccf73d5f7bd62aaee93fb1\: the full CI matrix is green — Ubuntu, macOS, Windows, Electron Linux/Windows, unified native macOS, deterministic provider discovery, Homebrew formula, and GitGuardian succeeded (live provider audit/Homebrew install skipped by workflow conditions). The installed Windows checkout now carries the two diagnostic commits as local \17f9f6c\ and \9dd5527\. Focused doctor/CLI regressions pass, \
pm.cmd run check\ passes, npm audit reports 0 vulnerabilities, and \uv pip check\ reports all 107 installed Python packages compatible. Live doctor now reports \Vision bridge: text-only models read images via gpt-5.6-luna\; service/health remain green. The regression also proves stale native catalogs are ignored after the fresh Codex sign-in probe becomes signed-out. No provider/model inference calls were used.

PR #541, opened 2026-08-30
fix: avoid repeated routed compaction prefixes

MeltingMadness:
## Summary

- configure Codex to count only post-prefix growth toward automatic compaction while routing is enabled
- preserve an explicit user-owned `model_auto_compact_token_limit_scope` value
- probe the installed Codex binary before writing the newer scope key, so older builds keep loading normally
- remove the router-owned value again during the normal disable lifecycle

## Problem

A routed compaction does not begin with an empty context. Codex carries a large stable prefix into the next compaction window, including system/developer instructions, tool schemas, and the router's rendered checkpoint.

With Codex's default `total` scope, that carried prefix is charged against `model_auto_compact_token_limit` again. On models whose routed session already has a large fixed prefix, the next window can therefore reach the threshold after little or no meaningful new work. In practice this presents as an agent compacting after nearly every round and failing to make sustained progress.

Codex now exposes `model_auto_compact_token_limit_scope = "body_after_prefix"` for this window shape. Its context-window implementation snapshots the carried prefill and charges only subsequent growth against the configured automatic-compaction limit, while the full context-window safety check remains active:

- https://github.com/openai/codex/blob/main/codex-rs/core/src/session/context_window.rs
- https://github.com/openai/codex/blob/main/codex-rs/protocol/src/config_types.rs

## Implementation

When the router enables its managed Codex configuration, it now writes:

```toml
model_auto_compact_token_limit_scope = "body_after_prefix"
```

The setting lives inside the existing `codex-router-managed` root block, so the established enable/disable lifecycle owns it.

Two guardrails keep the change conservative:

1. If the user already configured any scope explicitly, the router leaves that value unchanged.
2. The config manager asks the installed Codex binary to parse a minimal probe config before adding the key. A build that reports a configuration-load error, cannot be probed, or is not installed does not receive the managed setting.

This PR does not change model context windows, per-model compaction limits, checkpoint contents, or provider request payloads.

## Verification

Focused config-manager tests (3 passed, 0 failed):

- managed installs add `body_after_prefix` and remove it on disable
- an explicit user-owned `total` scope survives enable and disable unchanged
- a simulated older Codex build that rejects the key does not receive it

Also checked:

- branch rebased onto current `main`
- `git diff --check`
- PR diff contains only `src/config-manager.mjs` and `test/config-manager.test.mjs`

Per the repository testing policy, no broad full-suite run was performed for this isolated config-manager change.

### DIFF
--- a/src/config-manager.mjs
+++ b/src/config-manager.mjs
@@ -81,6 +81,7 @@ const standaloneWebSearchEndMarker =
   "# END codex-router-standalone-web-search-managed";
 const createdAgentsTableMarker = "# codex-router-created-agents-table";
 const managedAgentMaxConcurrency = 6;
+const managedAutoCompactTokenLimitScope = "body_after_prefix";
 // Codex 0.147 records a child's FINAL_ANSWER as subAgentActivity
 // `interacted` and keeps that child visually working for the whole live
 // parent turn. close_agent is not in the v2 toolset; interrupt_agent is the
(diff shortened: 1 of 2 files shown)


duolahypercho (changes requested):
Blocking three runtime-safety issues at exact head `f74a5c9cfa829d211a14e12251b6ac9f373b0aa5`:

1. **P1: the global scope switch invalidates catalog output headroom.** Catalog `autoCompact` values are total-context input ceilings deliberately sized so `contextWindow - autoCompact >= outputLimit`; `effective_context_window_percent` is 95. Under `body_after_prefix`, the same values become post-prefix growth budgets. For `opencode-go/qwen3.5-plus` (262,144 context, 190,000 auto-compact, 65,536 output), a 60,000-token carried prefix moves the trigger to the 249,036 hard cap and leaves only 13,108 tokens, not the declared 65,536 output reserve. The large prefixes this change targets are therefore exactly what make it unsafe.

2. **P2: a user-owned root limit is silently reinterpreted.** If config already contains `model_auto_compact_token_limit = 100000` with the historical implicit `total` scope, this patch adds `body_after_prefix` and turns a 100k total ceiling into 100k of post-prefix growth. At minimum, a user root limit without an explicit scope must prevent router scope management.

3. **P2: the compatibility probe is fail-open.** Any spawned process with no `result.error` and no exact English `Error loading configuration` text is accepted regardless of exit status, signal, empty output, alternate diagnostic, or localization. I reproduced `CODEX_BIN=/usr/bin/false` causing the scope to be written. A safe probe needs baseline-versus-candidate equivalence with normal termination or another positive support signal.

Because Codex exposes a global root scope but the router publishes heterogeneous per-model limits, I do not see a safe automatic adaptation yet. It would need both a per-model body-growth budget and an independent total hard input cap no greater than `contextWindow - maxOutputTokens`; the current 95% cap cannot preserve output limits above 5% of context. Please do not auto-publish this scope until that contract exists.

The branch is also behind current main `43deff51bf07454cd6afffc252a4d65a465a16b1`. If the design is revised, update the managed-config ownership docs and Unreleased changelog and rerun CI on the rebased head.

Verification performed: exact-head config-manager tests 51/51 and registry/curation tests 77/77 passed. Those tests do not exercise the changed runtime token semantics.

duolahypercho:
Closing as unsafe at exact head `f74a5c9cfa829d211a14e12251b6ac9f373b0aa5` after the requested changes received no follow-up.

The current global scope switch cannot preserve per-model output headroom, silently reinterprets a user-owned root `model_auto_compact_token_limit`, and its compatibility probe accepts unsuccessful or empty CLI executions. These are correctness boundaries, not rebase or documentation issues.

A safe redesign needs per-model body-growth budgets plus an independent total hard-input cap no greater than `contextWindow - maxOutputTokens`, and a fail-closed positive compatibility probe. The existing owner review records the concrete reproductions. A redesigned PR is welcome, but this implementation must not remain mergeable.

PR #558, opened 2026-09-01
fix: land unmerged local fixes for ChatGPT Web, usage reads, and tray appearance

duolahypercho:
Four commits had accumulated on a local `main` and were never pushed or
opened as a PR. They are extracted here unchanged, rebased onto current
`main`, and applied without conflicts.

- `feat: integrate ChatGPT Web sidecar provider`
- `style: distinguish local fallback points in usage charts`
- `fix: keep provider usage reads alive through account refreshes`
- `fix(macOS): follow system appearance in tray panel`

Two further commits from the same local branch — `Fix Claude setup from
desktop PATH` and `fix(cursor): publish BYOK-safe effort aliases` — are
**not** included. They build on the `ipc.mjs` rewrite in #555 and on
#547/#538, so they cannot be applied to `main` until those land. They are
parked on the `codex/main-local-snapshot` branch and will be rebased once
their dependencies merge.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,7 +55,7 @@ user.
 4. Determine which provider IDs the user requested: `anthropic-api`,
     `kimi-oauth`, `antigravity-oauth`, `kimi-api`, `kimi-api-cn`, `deepseek`, `grok-oauth`, `grok-api`, `qwen-plan`,
     `zai-coding`, `ollama-cloud`, `minimax-token-plan`, `meta`, `clinepass`,
-    `venice`, `nousresearch`, and/or
+    `venice`, `nousresearch`, `chatgpt-web`, and/or
    `opencode-go`
    (shown to users as "opencode Go/Zen"; its `opencode-go-messages`,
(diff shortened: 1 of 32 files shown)


duolahypercho:
Superseded — closing without merging.

These four commits were rescued from an uncommitted local `main` and opened here
so the work could not be lost. Since then #555 and #557 landed, and both turned
out to be supersets of this work: `src/direct-responses-provider.mjs` and
`config/chatgpt-web/chatgpt-web.json` are already on `main`.

Verified rather than assumed. Merging current `main` into this branch and
comparing trees:

```
main tree:   0db2f28e6a8567229dd8618e6afa292b16ff9f26
branch tree: 0db2f28e6a8567229dd8618e6afa292b16ff9f26
```

The rebased branch is byte-identical to `main`, so there is nothing left to
merge. Per-commit, `style: distinguish local fallback points in usage charts`
and `fix: keep provider usage reads alive through account refreshes` reverse-apply
cleanly against `main` (already present); the sidecar and tray-appearance commits
are present in adapted form.

The recovery this PR existed for is complete: none of that work is stranded.

PR #567, opened 2026-09-02
fix(grok-oauth): suppress repeated progress after stream abort

Leshabeats:
## Why

After a Grok conversation has produced a progress-only stop, a later turn can emit another short progress prefix and then abort upstream. Because that prefix is already committed to the Codex stream, the chat can look stuck in a loop even though the terminal stream error is reported correctly.

This implements the evidence-gated buffering alternative discussed in #295: keep the first affected turn fully live, and change streaming behavior only after the same stable Grok conversation ID has exhibited the progress-only shape.

## What changed

- remember affected conversation IDs in a bounded in-memory set
- on later eligible user turns, buffer only the short visible prefix while keeping headers and preceding reasoning live
- release the prefix for healthy short answers, longer answers, and tool calls
- suppress that prefix when the upstream stream aborts, while preserving the terminal SSE error
- document the behavior and add regression coverage for both the healthy and aborted paths

## Verification

- `npm run check`
- `node --test test/grok-oauth-turn.test.mjs test/grok-oauth-forwarder.test.mjs` — 58 passed
- `npm audit --omit=dev` — 0 vulnerabilities
- autoreview on the final dirty patch — no P0/P1 findings; secret scan clean

The full `npm test` run reported 3,377 passed, 24 skipped, and 7 failures in `test/control-center-harness.test.mjs`. The same 7 failures reproduce on a clean worktree at `origin/main` (`44029b8`), so they are not introduced by this patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- **Grok OAuth no longer replays a known progress-only sentence after an
+  aborted follow-up.** The first affected turn remains fully live. Once a
+  conversation has actually produced the progress-only shape, later
+  user-message turns buffer only a short visible prefix while leaving the
+  response head and preceding reasoning live. A real tool call or longer
+  answer releases that prefix immediately; an upstream abort reports its terminal
+  stream error without first committing the same status sentence to Codex
+  again. Evidence is retained in a bounded in-memory conversation set.
+
 - **Request bodies are decompressed off the event loop, and an oversize zstd
   frame is refused from its header.** Codex zstd-compresses every request, so
   the router inflated a buffer that grows with the conversation on the thread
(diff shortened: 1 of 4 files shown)


duolahypercho:
Merged — thank you, @Leshabeats.

Reviewed rather than taken on green CI, and three things in particular hold up well:

- **The state is bounded and carries nothing sensitive.** The new set is capped at 1024 with LRU eviction, and it keys on `conversationId`, which is a SHA-256 digest formatted as a UUID — no prompt content is retained. Reusing that helper instead of introducing a second identity was the right call.
- **Healthy conversations keep attempt 1 live.** Scoping the buffer to a conversation that has already produced a progress-only stop is what avoids reintroducing a first-byte hold for everyone, which is the trap this class of fix usually falls into.
- **The release is unconditional** before the final emit, so a normal short answer, a failed optional repair, and the keep-first branch all still deliver.

Verified against current `main` (which moved a long way today): merged without conflicts, `test/grok-oauth-forwarder.test.mjs` 34 pass 0 fail, full suite 3489 tests 3464 pass 0 fail.

PR #600, opened 2026-09-03
fix: harden routed Responses egress — double-render on text+tool turns, and inline reasoning-tag leaks

JDLondon7:
Two independent streaming artifacts that both originate in LiteLLM's chat-completions → Responses bridge (`use_chat_completions_api`) for routed providers, each fixed by a small egress SSE transform. Both are scoped to routed providers (`route` set) and leave native OpenAI Responses streams untouched. Two atomic commits.

---

## 1. Text+tool-call turns render twice (`ItemLifecycleNormalizer`)

**Symptom:** on a routed turn that contains **both assistant text and a tool call**, the assistant's sentence renders **twice** in the Codex TUI, with the tool call after it. Intermittent — only the text+tool-call shape triggers it.

**Root cause:** the bridge emits output items with overlapping lifecycles. Captured live from `qwen-plan/qwen3.8-flash`:

| wire order | event | output_index |
|---|---|---|
| 1 | `output_item.added` (message) | **0** |
| 2 | `output_text.delta` ×N | 0 |
| 3 | `output_item.added` (function_call) | **1** ← opens while item 0 still open |
| 4 | `function_call_arguments.*` + `output_item.done` | 1 |
| 5 | `output_text.done` | 0 |
| 6 | `output_item.done` (message) | **0** ← closes *after* item 1 |

The Responses contract requires each item to be `done` before the next `output_item.added`. Codex finalizes item 0's streamed text into a committed cell when item 1 barges in, then commits it **again** when item 0's delayed `output_item.done` lands — the same sentence twice.

**Fix:** `src/item-lifecycle-normalizer.mjs`, added last in the routed pipeline. While an item is open it holds events for a different index and replays them once the open item closes:

```
before:  0:added → 1:added → 1:done → 0:done      (contract violation)
after:   0:added → 0:done → 1:added → 1:done       (sequential)
```

Reorder-only — identical multiset of event payloads before/after; nothing synthesized or mutated. No-op on clean and native streams (verified byte-for-byte). Preserves LF/CRLF framing; correct across arbitrary chunk boundaries.

---

## 2. Inline `<think>` reasoning leaks into the answer (`ReasoningTagStripper`)

**Symptom:** the visible answer is prefixed with the model's chain-of-thought, or with a bare `</think>`:

```
output_text = "<think>The capital of France is Paris.</think>\nParis"
output_text = "\n</think>\n\nThe real answer starts here"
```

**Root cause:** reasoning models sometimes emit `<think>...</think>` inline in the content channel instead of on the reasoning channel; LiteLLM relays it as `output_text`. The tags also split across streamed deltas (`"<th"` then `"ink>..."`), so a per-delta replace misses them.

**Fix:** `src/reasoning-tag-stripper.mjs`, run before the lifecycle normalizer. Removes `<think>...</think>` spans and orphan tags across the streamed `output_text.delta`s (buffering a tag split across deltas), the terminal `output_text.done`, and the stored message item. Covers the reasoning-delimiter family the model varies to (`think`/`thinking`/`reason`/`reasoning` — all observed live). The structured reasoning channel and every non-message item are untouched; clean answers pass through unchanged. Tradeoff: a routed answer that deliberately prints a literal reasoning tag would lose it.

---

## Testing

- `test/item-lifecycle-normalizer.test.mjs` and `test/reasoning-tag-stripper.test.mjs` (new): interleave reorder, tag-strip across split deltas with delta/done/message convergence, chunk-boundary invariance, orphan-close stripping, the tag-name family, per-index independence, CRLF framing, and byte-faithful passthrough of clean/tool-only streams.
- `npm run check` passes; streaming-transform suites pass.
- Verified live end-to-end against a local router: after the change, a `qwen3.8-flash` tool-call turn emits sequential items, and reasoning-tag leaks (including varied tag names) are gone across repeated runs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## Unreleased
 
+- **Routed turns that mix assistant text and a tool call no longer render
+  twice.** LiteLLM's chat-completions-to-Responses bridge
+  (`use_chat_completions_api`) could leave the assistant `message` item open
+  across a `function_call` item and emit its `output_item.done` late, so Codex
+  committed the streamed text once while it was live and again when the delayed
+  close arrived — the same sentence appeared twice, with the tool call after it.
+  It surfaced intermittently on qwen-plan turns that both speak and call a tool.
+  A new `ItemLifecycleNormalizer` egress transform (`src/item-lifecycle-normalizer.mjs`),
+  added last in the routed-provider response pipeline, holds events for a
+  newly-opened output item until the currently-open item closes, restoring the
+  Responses contract that every item is `done` before the next
+  `output_item.added`. It reorders only — no event body is added, dropped, or
+  rewritten — and engages only when the upstream actually interleaves; clean
+  streams and native OpenAI streams pass through unchanged.
+
 - **A Grok OAuth outage now terminates streamed Codex turns instead of leaving
   a stale Working badge.** After the local gateway has exhausted its bounded
   retries, a final Grok 5xx on an explicitly streamed Responses request is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## Unreleased
 
+- **Routed reasoning models no longer leak `<think>` chains into the visible
+  answer.** Qwen (and other reasoning models bridged through LiteLLM's
+  chat-completions path) sometimes emit their chain-of-thought inline in the
+  content channel as `<think>...</think>` instead of on the reasoning channel,
+  so LiteLLM relays it as `output_text` and the answer renders behind the
+  model's reasoning — or behind a bare `</think>` when the open tag is consumed
+  upstream but the close is not. A new `ReasoningTagStripper` egress transform
+  (`src/reasoning-tag-stripper.mjs`) removes `<think>...</think>` spans and
+  orphan tags from the message text across the streamed `output_text.delta`s
+  (buffering a tag split across deltas), the terminal `output_text.done`, and
+  the stored message item. It covers the reasoning-delimiter family the model
+  varies to (`think`/`thinking`/`reason`/`reasoning`), leaves the structured
+  reasoning channel and every non-message item untouched, and passes clean
+  answers through unchanged. Routed providers only.
+
 - **Routed turns that mix assistant text and a tool call no longer render
   twice.** LiteLLM's chat-completions-to-Responses bridge
   (`use_chat_completions_api`) could leave the assistant `message` item open
--- a/src/item-lifecycle-normalizer.mjs
+++ b/src/item-lifecycle-normalizer.mjs
@@ -1,5 +1,4 @@
 import { Transform } from "node:stream";
-import { StringDecoder } from "node:string_decoder";
 
 // LiteLLM's chat-completions -> Responses bridge (use_chat_completions_api) can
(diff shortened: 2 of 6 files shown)


duolahypercho:
Rebased onto current `main` (maintainer). The old branch diverged by hundreds of unrelated commits and conflicted with #602's Grok SSE termination path. Kept the two feature commits only:

1. `ItemLifecycleNormalizer` (double-render fix)
2. `ReasoningTagStripper` (`<think>` leak fix)

Both land after the existing empty-completion terminal guard in the egress transform list. Focused tests for both modules pass locally. Waiting on CI + independent shipping verdict before merge.

duolahypercho:
## Independent shipping verification

**VERDICT: PASS+NOTES**

| Ref | SHA |
|---|---|
| PR head | `b88e02247633146bdfdecd19c860d93d013d4da1` |
| Base (`main`) | `faa3f3d70969edf46cac26c3c6652bd9d341b1ec` |
| Commit 1 (lifecycle) | `a7ee5f8f683c5bcf18b732222ffc07aea7fc2ba5` |
| Commit 2 (tag strip) | `b88e02247633146bdfdecd19c860d93d013d4da1` |

Rebase: `main` is a clean ancestor of head (`REBASE_CLEAN=yes`). Diff scope matches the claimed surface only:

- `src/item-lifecycle-normalizer.mjs` (new)
- `src/reasoning-tag-stripper.mjs` (new)
- `src/router.mjs` (wiring)
- `test/item-lifecycle-normalizer.test.mjs` (new)
- `test/reasoning-tag-stripper.test.mjs` (new)
- `CHANGELOG.md`

### Transform order (confirmed)

In `createResponsePipeline` for routed SSE egress, stages push in this order:

1. **Empty-completion guards** — `EmptyCompletionGuard` then `EmptyCompletionTerminalGuard` (when `route && !directResponses && EMPTY_COMPLETION_RETRY`)
2. **ReasoningTagStripper** — `reasoningTagStripperTransform(contentType)` when `route`
3. **ItemLifecycleNormalizer** — `itemLifecycleNormalizerTransform(contentType)` when `route` (**last**)

Native / no-`route` streams never attach either new transform.

### Local verification

- `node --test test/item-lifecycle-normalizer.test.mjs test/reasoning-tag-stripper.test.mjs` → **17/17 pass**
- `npm run check` → **pass**

### NOTES

1. **Attachment gate vs siblings.** The new stages attach on `route` alone; neighboring LiteLLM-bridge stages (empty-completion, Z.ai/Grok/tool-message compat, namespace restore) use `route && !directResponses`. Clean streams stay byte-identical, so this is low risk, but the gate is slightly broader than the LiteLLM `use_chat_completions_api` root-cause framing.
2. **Empty-completion sees pre-strip bytes.** Because tag stripping runs *after* the empty-completion guards, an answer that is only inline `<think>…</think>` can look non-empty to the guard and then become empty after strip — no empty-completion retry for that edge case.
3. **PR checks surface.** Only GitGuardian is reported green here; `mergeable_state` is `unstable`. Full CI visibility on this fork PR was not confirmed from the checks rollup (local focused suites + `npm run check` are green).

Not merging.

duolahypercho:
Fixed CI failure: both new egress transforms now frame on Buffers and disable rewriting on invalid UTF-8 (preserves `0xff 0xfe` bytes). Local `preserve ambiguous provider bytes` test passes.

PR #603, opened 2026-09-03
fix: keep Control Center in Dock after close

duolahypercho:
## Summary
- Closing the embedded macOS Control Center window now **hides** it instead of destroying the Electron process and retiring the Dock tile.
- Dock / Command-Tab stay available while Control Center is running so switching away no longer makes the app look like it quit.
- Dock hide is deferred until an explicit quit (`will-quit` / host terminate). Docs and Electron lifecycle tests updated.

## Test plan
- [x] `node --test test/control-center-electron.test.mjs` (69 pass, 1 skipped)
- [ ] Rebuild/open Codex Router Control Center on macOS
- [ ] Close the window (red traffic light) → Dock icon remains
- [ ] Cmd+Tab back (or click Dock) → window reappears
- [ ] Quit from the menu-bar host → Control Center and Dock tile go away

### DIFF
--- a/apps/control-center/electron/lifecycle-state.mjs
+++ b/apps/control-center/electron/lifecycle-state.mjs
@@ -206,8 +206,9 @@ export function shouldQuitOnLastWindowClosed({
   nativeTrayOwnedByHost,
   trayAvailable = true,
 }) {
-  // The embedded macOS child delegates its lifetime to the outer native host.
-  if (platform === "darwin" && nativeTrayOwnedByHost === true) return true;
+  // The embedded macOS child stays alive after the window hides so Dock and
+  // Command-Tab can return to it. The outer Swift host owns process exit.
+  if (platform === "darwin" && nativeTrayOwnedByHost === true) return false;
   // On Windows/Linux a usable Electron tray is what makes a windowless
   // lifetime recoverable. If construction failed (or it was later destroyed),
   // closing the fallback window must not strand an invisible process.
--- a/apps/control-center/electron/main.mjs
+++ b/apps/control-center/electron/main.mjs
@@ -33,6 +33,7 @@ let mutationLifecycle = {
   whenMutationsIdle: () => Promise.resolve(),
 };
 let deferredQuit;
+let isQuitting = false;
 let applicationReady = false;
 let windowVisible = false;
 let windowContentReady = false;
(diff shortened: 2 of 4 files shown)


duolahypercho:
Addressed independent FAIL: close-to-hide is now gated on `nativeTrayOwnedByHost || trayIsAvailable()`. Trayless Win/Linux still destroys and can reach `window-all-closed`. Tests assert the gate.

duolahypercho:
## VERDICT: PASS+NOTES

**Head:** `50c3e7a525264b285d8a55b7e90dace9d5b7698e`
**Base:** `b90c307335c28fe3d1f73a3fee420af2a8a8020f`
**patch-id (stable):** `c91ee87bf14f2f260f403ac493f75908c5d69196`

### Prior FAIL
Unconditional close-to-hide stranded Win/Linux when tray construction failed, because `preventDefault` + `hide` never reached `window-all-closed` → quit.

### Gate check
`50c3e7a` adds:
```js
if (!(nativeTrayOwnedByHost || trayIsAvailable())) return;
```
before `preventDefault` / `hide`.

Traced cases:
- **darwin + host** → hide; Dock kept until quit; `shouldQuitOnLastWindowClosed` stays `false` so Dock/Cmd+Tab can return.
- **win32/linux + tray live** → hide; process stays tray-recoverable.
- **win32/linux + tray missing/destroyed** (incl. Linux StatusNotifier miss) → destroy; `shouldQuitOnLastWindowClosed` returns `true` → quit. Stranding path closed.
- **`isQuitting`** → destroy allowed; Dock retired on `will-quit` / quitting `closed`.

Close gate and quit predicate stay aligned. Hide only when a recoverable owner exists.

### Verification
- `node --test test/control-center-electron.test.mjs` → 69 pass, 1 skipped, 0 fail
- Source regex asserts the gated close handler
- `shouldQuitOnLastWindowClosed` matrix updated for embedded darwin keep-alive

### Notes (non-blocking)
- Branch tip diverges from current `main` (`faa3f3d`) by 2 unrelated commits. Diff vs PR base and vs main tip share the same patch-id for these files.
- PR body still has unchecked manual macOS Dock smoke (close → Dock remains → Cmd+Tab reopen → host quit clears tile). Static+unit evidence covers the prior FAIL; that manual pass is still worth one operator click before merge.

Do not merge from this comment. Independent re-verify only.

PR #604, opened 2026-09-03
fix: allow ChatGPT account switches without disabling the router

duolahypercho:
## Summary
- ChatGPT account selection in Control Center was forcing a native catalog refresh while the router still owned `model_catalog_json`, so live `codex debug models` echoed the merged picker and aborted with “already-merged catalog.”
- Native capture now skips that live probe while the routed catalog is active, discards polluted `models_cache` entries that already contain routed slugs, and reuses a prior native capture when a forced refresh still fails.
- Control Center ChatGPT account add/remove now paints optimistically: new rows appear immediately as “Adding…”, removed rows disappear on confirm, and failures roll the list back when the durable write does not land.

## Test plan
- [x] `node --test test/catalog.test.mjs` (includes new live-probe / capture-path assertions)
- [x] `node --test apps/control-center/test/renderer.test.mjs` (includes delayed add/remove optimistic UI coverage)
- [ ] In Control Center Settings, add a ChatGPT account and confirm the row appears immediately
- [ ] Select / switch accounts while the router is running and confirm no “already-merged catalog” error
- [ ] Remove an account and confirm it disappears immediately; confirm a failed remove restores the row

### DIFF
--- a/apps/control-center/src/pages/SettingsPage.tsx
+++ b/apps/control-center/src/pages/SettingsPage.tsx
@@ -3,7 +3,17 @@ import { AppWindow, Check, Eye, LogIn, Moon, Plus, RefreshCw, Server, ShieldChec
 import { Badge, Button, Dialog, InlineNotice, PageHeader, SectionHeading, Toggle } from "../components";
 import { compactNumber } from "../lib";
 import { LANGUAGE_OPTIONS, type LanguageId, type Translate } from "../i18n";
-import type { ChatGptAccountPool, ChatGptSessionStatus, DoctorSnapshot, PresenceSnapshot, RouterControlApi, RouterHealth, RouterTarget, VisionEngine } from "../types";
+import type {
+  ChatGptAccountPool,
+  ChatGptSessionStatus,
+  ChatGptSubscriptionAccount,
+  DoctorSnapshot,
+  PresenceSnapshot,
+  RouterControlApi,
+  RouterHealth,
+  RouterTarget,
+  VisionEngine,
+} from "../types";
 import { useOptimisticValues, type RunAction } from "../useOptimisticValues";
 
 // Mirrors RETENTION_MIN/MAX/DEFAULT_TTL_DAYS in src/tool-result-retention.mjs.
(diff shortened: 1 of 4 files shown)


duolahypercho:
## Independent shipping verification

**VERDICT:** PASS+NOTES

| | |
|---|---|
| base | `b90c307335c28fe3d1f73a3fee420af2a8a8020f` |
| head | `60d50ddb9df1a2a64d0c29b933fe89a80733a2d5` |
| patch-id | `01965c28cfb37daf39a456d6315fff04d53cb89f` |
| reviewer | independent (did not author this patch) |

### Catalog ownership / `model_catalog_json`

PASS. Account switch still only refreshes catalog *artifacts* (`publishCatalog({ refreshNative: true })` after `restoreAccountCatalog`) and does not retarget or clear the managed `model_catalog_json` pointer. Publication still writes `merged-models.json`; config ownership stays with the existing marked block.

The abort path is fixed at the right layer:
1. Skip live `codex debug models` while `routedCatalogActive()` (managed router transport / catalog echo).
2. Discard `models_cache.json` entries that already contain routed slugs.
3. On forced refresh failure, reuse a prior `native-models.json` instead of failing the switch.

That keeps the router owning the merged catalog path without requiring the operator to disable the router first.

### Regression test coverage

PASS with notes. Locked:
- `liveAccountCatalogProbeAllowed` matrix (idle / routed-active).
- Source-shape assert that `captureNative` wires the gate + polluted-cache discard + reuse message.
- Control Center optimistic add/remove (real delayed round-trip UI test).

Not locked behaviorally:
- No exercised `captureNative` / `nativeCatalog({ refreshNative: true })` run under a routed config that proves the live probe is not spawned and that an “already-merged” throw no longer aborts publication.
- The catch-and-reuse prior native capture path is only string-matched, not executed.
- No switch-path integration covering `chatgpt-profile-switch` → `refreshActiveCatalog` with a polluted or missing account catalog.

### Notes (non-blocking)

- A successful bundled-only recapture after discarding a polluted `models_cache` will overwrite a richer prior `native-models.json`; switch proceeds, but account-specific native rows can shrink until Codex rewrites a clean cache.
- PR manual checks for live account switch / add-remove still unchecked; CI is green on this head.

Safe to land from a catalog-ownership perspective. Stronger behavioral capture/switch coverage would turn this into a hard PASS.

PR #605, opened 2026-09-03
feat: add Vertex AI model discovery, curation, and routing

dinsen:
## Summary

Adds Google Cloud Vertex AI as a router-managed provider, including safe Google authentication, curated model discovery, protocol-specific forwarding, and Control Center support.

- Adds a resolver-backed `vertex` provider that uses Google Application Default Credentials (ADC), never a stored API key.
- Stores only validated Vertex project and location in private router state; bearer tokens are obtained from `gcloud`, cached in memory for five minutes, and never written to disk, shown in diagnostics, or included in support bundles.
- Adds `./bin/control vertex status|set <project-id> <location>|clear` and presents ADC setup guidance instead of an API-key prompt throughout CLI, setup, doctor, and Control Center surfaces.
- Discovers eligible publisher models through Vertex Model Garden with pagination, bounded/untrusted-response validation, credential-safe error reporting, and a cache keyed by publisher configuration.
- Adds an explicit offline curation mode: `./bin/curate-models vertex --static --models <id>`. It only uses reviewed records in `config/vertex/support-catalog.json`; it never silently substitutes static data after a live discovery failure.
- Curates only models with an implemented adapter and copies reviewed metadata into private user-model state.
- Implements both supported Vertex wire contracts:
  - Anthropic Claude models use Vertex `rawPredict` / `streamRawPredict` via the Anthropic Messages surface.
  - Google Gemini models use Vertex's OpenAI-compatible chat-completions endpoint.
- Adds reviewed static entries for Claude Opus 5, Claude Sonnet 5, Gemini 2.5 Pro, and Gemini 2.5 Flash.
- Extends tool-schema relay so routed models are valid `collaboration.spawn_agent` model choices.
- Makes routed-model catalog construction lazy and memoized, fixing the macOS request-deadline regression; normal turns without `spawn_agent` avoid that unnecessary work.
- Adds Vertex branding and icon coverage in the web Control Center and macOS tray app.

## Setup and operator behavior

1. Configure Google credentials locally with `gcloud auth application-default login` (or an existing `gcloud auth login` session).
2. Set the Vertex project and location with `./bin/control vertex set PROJECT_ID LOCATION`.
3. Select/curate models normally, or use explicit `--static` curation when Model Garden discovery is unavailable.

Static curation is intentionally not a permission or entitlement check: requests still require Vertex API enablement, IAM access, and publisher-model access.

## Security and compatibility notes

- Vertex rejects API-key collection and API-key writes at every supported entry point.
- The provider configuration allows only the Google ADC resolver and only for the Vertex protocol.
- Forwarding chooses the endpoint and request shape from an explicit per-model adapter; an arbitrary discovered model cannot become routable without a reviewed adapter.
- Vertex avoids Gemini's trailing-user-turn rewrite because Vertex Anthropic and OpenAI-compatible endpoints have their own request contracts.
- Provider errors are redacted, and discovery retains the existing SSRF/redirect/body-size protections.

## Validation

- Added 328 lines of credential/configuration coverage, including protected state, redaction, token caching, missing-ADC handling, key-write refusal, Control Center behavior, setup, doctor, and support-bundle handling.
- Added 698 lines of Vertex provider coverage for adapters, Model Garden discovery/pagination, cache identity, static curation, endpoint selection, routing, streaming, and error redaction.
- Added namespace-relay coverage for routed subagent model schemas.
- Current fork CI passed on Ubuntu, macOS, and Windows, plus Electron Linux/Windows, unified macOS packaging, Homebrew validation, and audit: https://github.com/dinsen/codex-router/actions/runs/33791629496

## Test plan

- [x] GitHub Actions: `npm run check`, `npm test`, and dependency audit on Ubuntu, macOS, and Windows.
- [x] GitHub Actions: Electron integration, packaging, and platform smoke coverage on Linux and Windows.
- [x] GitHub Actions: unified native macOS build, bundle verification, lifecycle smoke test, and packaging.
- [x] Automated Vertex tests cover ADC/configuration redaction, token caching, API-key refusal, reviewed static curation, Model Garden discovery and pagination, adapter selection, streaming routes, and provider error redaction.
- [x] Automated namespace-relay tests cover the routed `collaboration.spawn_agent` model schema.
- [ ] In a project with Vertex API, IAM, and Model Garden access, configure ADC and a project/location; curate a reviewed model and complete a routed text turn.
- [ ] Manually verify a Claude Vertex stream and a Gemini Vertex chat-completions turn, including tool use and image input where the selected model supports it.
- [ ] Manually run `./bin/curate-models vertex --static --models claude-sonnet-5` without Model Garden access; confirm it only curates reviewed metadata and that a later request still reports any missing entitlement normally.
- [ ] Check `./bin/model-router codex doctor` and a generated support bundle on a configured host for redacted authentication details.

## Reviewer focus

- Confirm the supported Model Garden publisher/model mapping and reviewed capability metadata.
- Confirm that ADC/token handling, state permissions, diagnostics, and support-bundle behavior do not create a secret-persistence path.
- Confirm the two adapter contracts remain appropriately isolated as new Vertex model families are added.
- This PR intentionally does not add automatic project provisioning, arbitrary dynamic model routing, or a live quota-consuming probe.


PracticalSwan:
Draft compatibility note from current-main maintenance: the `src/namespace-relay.mjs` change that augments `collaboration.spawn_agent`'s `model` schema with every `routedSubagentModels` entry now conflicts with the provider-affinity boundary established by #629 and tightened by #632.

On current main, the request-local `SPAWN_AGENT_MODELS` set is used by `sanitizeSpawnAgentModel(...)` to decide whether a model-generated `spawn_agent.model` override is accepted. This PR adds all routed model slugs to that allowed set, so a routed parent can be encouraged/allowed to emit a different routed provider/model for an in-session child. Fresh Windows validation for #632 confirmed that preserving a valid explicit model on `spawn_agent` is a real provider/billing-boundary crossing path.

Before this draft is marked ready, please reconcile the contract explicitly:

- `codex_app.create_thread`: an explicit model remains a user-visible independent task choice.
- `collaboration.spawn_agent`: the model argument is model-generated and should be pinned to the current routed parent, not advertised as a menu of arbitrary routed models.

If Vertex models need to be valid parent models for routed subagents, they can still participate by being the current route; the child should inherit/pin that exact parent slug. The provider-affinity regression tests from #632 (flattened and native Responses forms, including a valid-but-different explicit model) should be carried into the eventual rebase.

This is separate from the Vertex ADC/adapters themselves; I have not completed a full 51-file security review of this draft.

duolahypercho:
Thank you for this — the Vertex work is careful, and I want to be clear up front that I'd like it to land. I checked it out and ran the Vertex surface against your base: 101 tests pass and `npm run check` is clean. The ADC handling in particular is the right shape — token from `gcloud`, memory-cached, never written, and `writeProviderCredential` refusing an API key outright.

It needs a rebase before it can move, and one design question resolved first. The branch is 206 commits behind and conflicts in four files — only one hunk each, but three of them (`src/namespace-relay.mjs`, `src/router.mjs`, `src/chat-tool-surface.mjs`) sit exactly where `main` was rewritten for the Responses-native identity fix and the provider-affinity boundary, so the reconciliation is a judgement call rather than a text merge.

The design question is the one raised in review: augmenting `collaboration.spawn_agent`'s `model` schema with every routed slug widens the `SPAWN_AGENT_MODELS` set that `sanitizeSpawnAgentModel()` gates on, which lets a routed parent name a different routed provider for its child — the billing boundary #629/#632 exist to hold. Vertex doesn't need it: a Vertex model can be the parent route and the child pins that exact parent slug. I'd suggest dropping that part entirely; it's separable from the adapters and discovery, which are the valuable half.

Three smaller things for the rebase:
- `src/vertex-credentials.mjs:28` calls `execFileSync("gcloud", …)` directly. On Windows `gcloud` is `gcloud.cmd`, which Node has refused to spawn without a shell since CVE-2024-27980 — please route it through `src/spawnable-command.mjs` (AGENTS.md has the rule and the `devinCliVersion` precedent). Right now `failureReason()` would map that error to `adc-not-configured`, so a Windows user with working ADC gets told to configure ADC.
- The registry entry needs a `planNote`. Your own summary says a curated model still needs Vertex API enablement, IAM, and publisher-model access — that belongs where someone connects, not as a 403 inside Codex.
- AGENTS.md's provider checklist also asks for `src/provider-account-usage.mjs` wiring (or a graceful "usage unavailable") and an `OAUTH_CLIS` entry so the tray's sign-in button installs `gcloud` and runs `gcloud auth application-default login` in one click.

One practical note: the PR is from your fork's `main` branch, which makes a rebase force-push awkward and blocks CI approval. If you push the rebased work to a topic branch and open it against `main`, that'll be much easier to iterate on. Happy to review again once it's out of draft.

duolahypercho:
Gave this a triage-depth review since it's still a draft. Short version: **the expensive part is right and worth finishing** — the problems are concentrated in how Vertex is declared and in one contract violation, not in the discovery work.

**What's good:** the discovery + curation half follows the house pattern closely — `fetchUntrustedModelCatalog`/`validateModelCatalogPayload`, `model-catalog-cache`, `mergeDiscoveredModels`, curation funnelled through `user-models.mjs`. The support-catalog allowlist is a genuinely good idea: a discovered Model Garden id cannot become routable without a reviewed entry naming an implemented adapter. And `test/vertex-provider.test.mjs:526` spawns a real forwarder against a real upstream and asserts the exact target paths and headers — that test would fail if the adapters were reverted.

**Blocking: the credential reader sits above the `--no-discovery` guard.** The Vertex branch is inserted at `src/provider-credentials.mjs:222`, thirteen lines *above* the `if (discoveryDisabled()) return undefined;` at :235 — whose own comment says it sits "before everything that does [read]". `credentialStatus()` bypasses it a second time at :378, and `doctor.mjs` iterates every provider through `credentialStatus`, so a doctor run on a `--no-provider --no-discovery` install still spawns `gcloud` and reads the user's ADC. AGENTS.md ~1668 is explicit that the guard belongs at the reader, not only at its callers. There is no `CODEX_ROUTER_NO_DISCOVERY=1` test anywhere in the 1,026 new test lines — that test is the one I'd want most.

**Please split out `routedSubagentModels`.** It isn't Vertex, and it contradicts policy that landed after your fork: `ea3dbb59` (#629, "keep routed subagents on parent model") and `b944249d` (#632, "enforce routed subagent provider affinity"). Main now removes an unsupported explicit `model` from `spawn_agent` and fails closed "rather than silently crossing a billing boundary"; this widens the enum with every routed slug, and `test/namespace-relay.test.mjs:122` asserts exactly the cross-provider spawn #632 prevents. It is also what makes the rebase expensive — dropping it turns four conflicts into two mechanical ones (`test/provider-catalogs.test.mjs` is a count change, `src/chat-tool-surface.mjs` is one options object taking both keys). Heads up too that resolving `src/namespace-relay.mjs` in this PR's favour would silently delete the `CUSTOM_TOOL_IDENTITIES` fix from `e29bcdd8`.

**Design note worth deciding before more work:** `protocol: "vertex"` is a hosting platform, not a wire protocol. Every other provider's `protocol` answers "what body shape goes on the wire", and ~35 call sites across 11 files branch on it; this patches four and leaves the rest taking the `else`. `providerModelEndpoint()` already returns `undefined` for Vertex, which is harmless only until a Vertex model sets `supportedEndpoints`. The mechanism for "one provider, different endpoints per model" exists — `authMode: "per-model"` / `perModelEndpoint` / `endpointForModel`. Similarly, `credential: { resolver: ... }` is special-cased at ~11 call sites where `kind: "oauth"` (which ADC effectively is) would carry it, and the registry validator had to grow a bespoke rule forbidding the resolver on non-Vertex protocols — a sign the field is really a one-provider switch.

Top functional bugs:

- **Env-configured project/location resolves a credential but 500s on every request.** `resolveVertexCredential` calls `resolveVertexConfiguration(options)` with `{}`, so `VERTEX_PROJECT_ID`/`VERTEX_LOCATION` are honoured; `vertexForwardBaseUrl` calls it with `{persistent: true}`, which ignores env and throws. Doctor reports configured, the picker lists models, every request fails.
- **`execFileSync("gcloud", …)` runs on the forwarder's event loop, in the request path**, blocking every other provider's in-flight SSE relay once per 5-minute TTL.
- **The catalog-cache identity fingerprints the bearer token**, which rotates roughly hourly — so the cache self-invalidates, and a rotation *during* a fetch throws `credentialChangedError` with no credential change. Fingerprint project/location/publishers/apiVersion and an account identifier instead.
- **`vertex-openai-chat` hardcodes the `google/` prefix** and ignores `vertexPublisher`, which both the registry and the curation path already carry. Latent today; a 404 the moment a non-Google model uses that adapter.
- **Silent fallback from ADC to the operator's personal gcloud login** when `print-access-token` fails for any non-ENOENT reason — billing and audit attribution move to a different principal with only a `source` string to show for it.

Concrete order I'd suggest: drop `routedSubagentModels` into its own PR, move the credential read below `discoveryDisabled()` and add the kill-switch test, make the token resolution async, drop the token from the fingerprint. The discovery design itself I would keep as-is.

duolahypercho:
Thanks for the Vertex discovery/curation work. This draft is superseded by #788, which is already on `main`.

#788 kept the ADC + catalog-only shape (no stored API key; a discovered Model Garden id is not routable until it is curated onto a reviewed adapter) and adapted it onto current main. It deliberately did not take the `collaboration.spawn_agent` widening from this branch, honors `--no-discovery`, spawns `gcloud` through `spawnableCommand`, and refuses a silent `gcloud auth login` token fallback.

Closing this PR. Please do not merge this draft.

PR #612, opened 2026-09-04
fix(opencode): send x-opencode-session on outbound hops

duolahypercho:
## Why

OpenCode Go emailed that requests missing `x-opencode-session` may error starting 2026-09-06. The router never sent that header on chat, discovery, or usage hops. Without it, Go/Zen routes risk hard failures tomorrow.

## Scope

- Add `src/opencode-session.mjs` to resolve a stable opaque id: `thread-id` when present, else a Grok-style hash of the first conversation items, else a named fallback for discovery/usage.
- Attach `x-opencode-session` in `src/api-forwarder.mjs` `upstreamHeaders` for every `ownedBy: "opencode"` provider (Go, Messages, Responses, Zen, Free).
- Attach the same header plus an honest `codex-router/...` User-Agent on OpenCode catalog fetches and the Zen usage probe.
- Cover resolution order and provider gating in `test/opencode-session.test.mjs`; assert the usage probe sends the header.

## Tradeoffs

Body-hash affinity is used when Codex `thread-id` does not survive LiteLLM, because `routedHeaders()` does not forward thread identity today. That keeps one id stable across turns of the same conversation without a larger gateway header-passthrough change.

## Blast Radius

Touches only OpenCode family outbound HTTP. Other providers keep previous headers. No CLI impersonation headers are added.

## Verification

- `node --test test/opencode-session.test.mjs` — 9/9 pass
- `node --test test/provider-account-usage.test.mjs` — pass, including the new usage-header assertion
- `node --test test/provider-api-key-forwarder.test.mjs` — pass

### DIFF
--- a/src/api-forwarder.mjs
+++ b/src/api-forwarder.mjs
@@ -1122,12 +1122,14 @@ function upstreamHeaders(requestHeaders, body, apiKey, provider, extraHeaders = 
   }
   headers["User-Agent"] = `codex-router/${VERSION}`;
   headers["Accept-Encoding"] = "identity";
+  Object.assign(headers, extraHeaders);
+  // OpenCode Go/Zen affinity must win over any session.headers merge above:
+  // starting 2026-09-06 their edge may refuse requests without this header.
   applyOpenCodeSessionHeaders(headers, {
     provider,
     requestHeaders,
     body,
   });
-  Object.assign(headers, extraHeaders);
   // Content-Length is fetch's to compute. An explicit copy is at best
   // redundant, and the HTTP/1.1 dispatcher rejects the request outright
   // (UND_ERR_INVALID_ARG) when a caller-supplied value accompanies a body.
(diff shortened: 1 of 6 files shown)


duolahypercho:
## Independent shipping verification: PASS+NOTES

Verified against head `19aa043` (base `e5ce835`, patch-id `86b208913f99685ce80c8480272455734b2bb630`). Independent of the authoring agent. Did not merge.

- **Outbound hops:** `x-opencode-session` is applied in `api-forwarder` `upstreamHeaders` (after `extraHeaders`), OpenCode catalog fetches in `model-discovery`, and the Zen usage probe in `provider-account-usage`.
- **Tests:** `test/opencode-session.test.mjs` 9/9 locally. Same conversation (extended messages) keeps one id; non-OpenCode providers get no header. Usage probe asserts `x-opencode-session: codex-router-usage` and `User-Agent` `^codex-router/`.
- **No CLI impersonation:** UA is `codex-router/${VERSION}` only. No OpenCode CLI credential or UA forgery in the diff.
- **CI note:** required matrix still pending at verify time (`test` macos/ubuntu/windows, Electron, CodeQL JS/Swift, python-lock). No failures observed. Re-check CI before land.

Issue #568, opened 2026-09-02
[Bug]: Qwen Plan v1 collaboration calls are returned as pre-flattened names and rejected by Codex Desktop


AdrenalineY:
### What happened?

# [Bug] Qwen Plan v1 collaboration calls are returned as pre-flattened names and rejected by Codex Desktop

## Summary

When Alibaba Model Studio Token Plan models are used through `codex-router`, normal responses, streaming, and ordinary tool calls work, but the affected models cannot spawn native Codex Desktop subagents.

The model returns a function call such as:

```json
{
  "type": "function_call",
  "name": "multi_agent_v1__spawn_agent",
  "namespace": null
}
```

Codex Desktop then rejects it:

```text
unsupported call: multi_agent_v1__spawn_agent
```

No native child task, child rollout, or `thread_spawn` relationship is created.

This does not appear to be an authentication, networking, or basic tool-calling failure. The same credentials and route can serve normal responses and execute ordinary tools. The evidence instead points to a tool-identity or namespace-restoration problem affecting legacy v1 collaboration tools on the Chat Completions bridge.

## Environment

- OS: Windows
- Codex Desktop / CLI: `0.151.0-alpha.7.2`
- Runtime reproduction: `codex-router 0.5.1` at commit `4d5ca178ffdf3a0e6def80cdc0957ab5c4da93a1`
- Current upstream source audit: `main` at commit [`44029b8729b8166f8a662f5b62883c192bbeb9df`](https://github.com/duolahypercho/codex-router/commit/44029b8729b8166f8a662f5b62883c192bbeb9df)
- Provider: `qwen-plan`
- Endpoint: `https://token-plan.cn-beijing.maas.aliyuncs.com/compatible-mode/v1`
- Credentials: valid; secrets are intentionally omitted

Affected models observed in the runtime reproduction:

- `qwen-plan/deepseek-v4-flash-0731`
- `qwen-plan/glm-5.2`

## Steps to reproduce

1. Configure `qwen-plan` with valid Alibaba Model Studio Token Plan credentials.
2. Select either affected model in Codex Desktop.
3. Start a new task and ask the model to spawn multiple native Codex Desktop subagents.
4. Inspect the resulting rollout.
5. Observe that the model calls `multi_agent_v1__spawn_agent` and that Codex Desktop rejects it as an unsupported call.

Example prompt:

```text
Spawn five native Codex Desktop subagents in parallel and assign one independent analysis task to each subagent.
```

## Actual behavior

[...]

Issue #581, opened 2026-09-03
Windows: router never comes back after power event / unlogged exit (LogonTrigger task + RestartOnFailure did not relaunch)


ALCHEMISTK:
## Windows: router never comes back after power event / unlogged exit — LogonTrigger task + RestartOnFailure did not relaunch it

**Environment:** Windows 11, router 0.5.1 (commit 37e9174, install manifest), target `codex`, Node 24.19.0. Clients: Codex desktop.

**Setup (router-installed Task Scheduler task "Codex Router"):**
- Trigger: `LogonTrigger` (`AtLogOn`, current user)
- Action: `wscript.exe //B //NoLogo <state>/start-codex-router-hidden.vbs` (vbs waits on `cmd` -> `node src/start.mjs`; `Wait=True`)
- Settings: `MultipleInstancesPolicy=IgnoreNew`, `RestartOnFailure Count=999 Interval=PT1M`, `ExecutionTimeLimit=PT0S`

**Observed:**
1. Router was serving fine, then its process tree exited with no error in `router.log` (log simply stops; last lines are normal timing lines) and no crash event in the Windows Application log (no 1000/1002/1026 for node around the exit window).
2. After the exit, nothing was listening on the router port (~hours). Codex desktop retried: endless "Reconnecting... waiting for network".
3. The scheduled task never relaunched it: `Get-ScheduledTaskInfo` showed `LastRunTime` unchanged since the logon launch, and `LastTaskResult = 4294967295` (0xFFFFFFFF). `RestartOnFailure` count never advanced; the task stayed `Ready` (not running).
4. Manually re-running the same launcher (`wscript …vbs`) restores the router; everything works again.

**Hypothesis:** the tree died during a sleep/shutdown/fast-startup cycle (process suspended then terminated, or a silent exit). The task instance ends when the waiting `wscript` exits with the dead node's status — but `RestartOnFailure` did not relaunch it, and `LogonTrigger` doesn't re-fire on wake-from-sleep or fast startup. Expected: `RestartOnFailure` should have relaunched within 1 minute, per the README statement that on Windows "the service still exits and the OS supervisor rebuilds it".

**Ask:** can the Windows supervisor be strengthened? Options: also register a repeating/minutely wake trigger, or ship a small health-backed watchdog (TCP probe against the router port -> relaunch same hidden launcher). A loopback health endpoint already exists (unauthenticated leaf), so the probe needs nothing new.

[...]

PR #616, opened 2026-09-04
fix(namespace-relay): restore dotted tool delimiters (#611)

### LINKED ISSUE #611: [Bug]: Dot-form tool names (collaboration.spawn_agent, mcp__*.name) arrive in custom-model threads and are rejected as unsupported call
## Summary

In custom-model threads served through `codex-router`, the tool list arrives
with **dot-separated** tool names. The relay table only knows the
double-underscore flattening, so these calls are relayed through
unrecognized and the app answers `unsupported call`. This breaks both
collaboration spawn and MCP tools on the affected routes, while normal
responses and ordinary tools (shell, file edits) work fine on the same
route with the same credentials. This looks like a tool-identity /
namespace-restoration problem, not a model capability problem.

Names as received in the thread:

- `collaboration.spawn_agent` (relay knows `collaboration__spawn_agent`)
- `mcp__agentmemory.memory_sessions` (relay knows
  `mcp__agentmemory__memory_sessions`)

Result of calling them verbatim:

- `unsupported call: collaboration.spawn_agent`
- `unsupported call` on the dotted MCP calls

## Environment

- OS: macOS (arm64)
- codex-cli 0.153.0, Codex Desktop harness
- Router install at commit `b90c307` ("fix: flatten recursive tool schemas
  for Muse Spark opencode routes")
- Route: `opencode-free-responses/muse-spark-1.3-contributor-free`
- Upstream head checked at report time: `e5ce835`; the 6 commits between
  touch image-input flags, grok-oauth, account switching, dock, deps, and
  egress hardening -- none touch the relay, so no fix is pending upstream
  for this as far as I can tell.

## Steps to reproduce

1. Serve `opencode-free-responses/muse-spark-1.3-contributor-free` through
   the router in Codex Desktop.
2. Start a **fresh** thread on that route (this is not a stale toolset:
   fresh threads fail identically).
3. Ask the model to spawn a native subagent.
4. The model calls `collaboration.spawn_agent` (the exact spelling from its
   tool list) and Codex answers `unsupported call`.
5. Same for dotted MCP names such as `mcp__agentmemory.memory_sessions`.

## What I already ruled out

- Stale thread toolset (#82 recipe): fresh threads fail the same way, and
  the tools **are** surfaced here -- just with dots. So this is not #82.
- Route selection: the Subagents toggle is On for this route,
  `multi-agent-settings.json` has mode `selected` with the route enabled,
  and after a manual `node src/catalog.mjs` publish the agent definition
  exists at `~/.codex/agents/router-model-opencode-free-responses-muse-spark-1-3-contributor-free.toml`
  (`routed_agents: 1`). Spawning still fails at call-relay time.
- Model capability: the same route streams, calls ordinary tools, and
  carries multi-step work. Only the dotted collaboration/MCP identities
  are rejected.

## Relation to #568

This provides the request-side tool inventory asked for in #568: here the
client sends collaboration calls dot-separated with `namespace: null`,
and the relay has no reverse mapping because the table is built from exact
flattened `__` names only ("never by splitting names"). I understand from
the #568 analysis that mapping these by name alone is deliberately
declined for v1 collaboration -- recording this case so the dot-form
arrival itself is visible, since it also affects MCP tools, not just
collaboration.

No secrets or credentials are included in this report.

duolahypercho:
## Why

On OpenCode Muse Free Responses routes, collaboration and MCP calls arrive as `namespace.tool` (and `mcp__server.tool`) while the relay table only knows `__` joins. Codex answers `unsupported call`, so spawn and MCP fail even though ordinary tools work.

## Scope

- `buildNamespaceLookups` registers `${namespace}.${name}` as an alias only when that spelling is not already owned.
- `buildNamespaceLookupsFromTools` seeds the same aliases from namespace containers and collaboration flats.
- Test covers the two #611 spellings against the request map.

Does not invent namespace splits from unknown dotted strings (#568).

## Blast Radius

Namespace restore on routed Responses/chat paths. Non-OpenCode routes keep identical `__` restoration; dotted aliases are additive.

## Verification

- `node --test --test-name-pattern='issue #611' test/namespace-relay.test.mjs` — pass
- `node --test test/namespace-relay.test.mjs` — 115 pass
- `node --test test/qwen-plan-namespace-restoration.test.mjs` — run in CI

Fixes #611

### DIFF
--- a/src/namespace-relay.mjs
+++ b/src/namespace-relay.mjs
@@ -41,6 +41,9 @@ import {
 // tool itself. Namespace names themselves may contain the delimiter
 // (`mcp__codex_apps__github`), so restoration always resolves through the map
 // built from the exact tools that were flattened -- never by splitting names.
+// The same map may also index a dotted inventory alias (`namespace.tool`) when
+// a Responses-native model echoes that wire form (#611); that is still an
+// exact inventory hit, not a split.
 
 export const NAMESPACE_DELIMITER = "__";
 const DEFAULT_FUNCTION_NAMESPACE = "functions";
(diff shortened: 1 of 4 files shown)


duolahypercho:
Superseded: this branch mixed #581 + #611 under the wrong name. Reopening as separate PRs.

PR #618, opened 2026-09-04
fix(windows): relaunch router via minute heartbeat trigger (#581)

### LINKED ISSUE #581: Windows: router never comes back after power event / unlogged exit (LogonTrigger task + RestartOnFailure did not relaunch)

duolahypercho:
## Why

Task Scheduler `RestartOnFailure` only covers actions that fail to *start*. After a silent process exit or power event, the LogonTrigger task stays Ready forever and the router never comes back.

## Scope

Register a repeating minute heartbeat trigger with `MultipleInstances IgnoreNew` and `StartWhenAvailable` so a dead instance is relaunched without stacking duplicates while one is still running.

## Blast Radius

Windows service task XML only. macOS/Linux LaunchAgent/systemd unchanged.

## Verification

- `node --test test/service-render.test.mjs` (focused assertions for the heartbeat trigger)

Fixes #581

### DIFF
--- a/src/service-windows.mjs
+++ b/src/service-windows.mjs
@@ -103,11 +103,13 @@ function wrapper() {
 // window style of 0. Without it the wrapper owned a console window that stayed
 // on screen for the router's lifetime and reappeared on every watchdog restart.
 //
-// The `True` wait flag is what keeps Task Scheduler's restart settings alive:
-// Run then blocks until the wrapper exits and returns its exit code, which the
-// script re-raises through WScript.Quit. Quitting with a fixed 0 (or letting the
-// script fall off the end) would report every crash as a clean exit and silently
-// disable RestartCount/RestartInterval.
+// The `True` wait flag keeps the task instance alive for the router's lifetime
+// so Task Scheduler state and `schtasks /End` track the real process tree.
+// Propagating the wrapper exit code still matters for diagnostics
+// (`LastTaskResult`), but it does not drive relaunch: Task Scheduler's
+// RestartOnFailure only covers actions that fail to start, not a non-zero
+// exit after a successful start (issue #581). Relaunch after exit or a power
+// event comes from the minute heartbeat trigger in installTask().
 function launcher() {
   // A Windows path cannot contain a double quote, but escape it anyway so a
   // hand-edited state directory can never break out of the string literal.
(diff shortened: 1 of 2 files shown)


PracticalSwan:
Follow-up from live Windows validation: the heartbeat added here correctly self-heals a dead router, but it also restarts the service after an intentional `service stop` because the scheduled task remains enabled. I reproduced the router returning stopped/health down, then relaunching on the next minute heartbeat. PR #630 fixes that by disabling the scheduled task before explicit stop and re-enabling it on start/restart, while leaving heartbeat recovery unchanged for an intentionally enabled service. Live verification crossed a full heartbeat boundary and confirmed the stopped task remained Disabled with no router process until explicit start.

PR #621, opened 2026-09-05
fix: refresh native catalog without rewriting config when safe

PracticalSwan:
## Summary

Make `refresh-catalog` avoid rewriting `~/.codex/config.toml` when the signed-in Codex account cache is already a valid native-only source.

## Problem

`refresh-catalog` currently always disables the routed transport before recapturing native models. On Windows that first config mutation can fail with `EPERM` if another local process has `config.toml` open without delete sharing, even though `catalog.mjs --refresh-native` can safely refresh from Codex's separate native `models_cache.json`.

I reproduced this after updating Codex Desktop to 26.901.5003.0 / `codex-cli 0.153.3`: the wrapper failed on the atomic `config.toml` rename, while the catalog-only refresh succeeded from a native-only account cache and cleared the stale-catalog warning without changing routing.

## Fix

- Add a fail-closed preflight that permits in-place refresh only when `models_cache.json` is valid and contains no checked-in routed slugs, or when discovery is disabled and capture is bundled-only.
- In ordinary signed-in routed mode, use that safe in-place catalog refresh without touching `config.toml`.
- Preserve the existing disable/restore transaction when the cache is missing, invalid, or contaminated.
- Preserve the existing journaled transport flow for login-free mode even when the preflight would otherwise pass.

## Verification

- `node --test test/catalog.test.mjs test/refresh-catalog.test.mjs` — 75/75 passed
- `npm run check` — passed
- `git diff --check` — passed

No login-free safety behavior or credential handling is weakened.

### DIFF
--- a/src/catalog.mjs
+++ b/src/catalog.mjs
@@ -202,6 +202,21 @@ function readModelsCache() {
   }
 }
 
+// A routed custom catalog never rewrites Codex's account cache. When that cache
+// is valid and contains no routed slugs, it remains a safe native source even
+// while model_catalog_json points at the merged router catalog.
+export function nativeCacheCanRefreshInPlace(cache = readModelsCache()) {
+  const catalog = cache?.catalog;
+  return (
+    Boolean(validNativeCatalog(catalog)) &&
+    !catalog.models.some((model) => MODEL_BY_SLUG.has(String(model.slug)))
+  );
+}
+
+export function nativeCatalogCanRefreshInPlace() {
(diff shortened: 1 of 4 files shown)


PracticalSwan:
Closing this as superseded by merged PR #624, which incorporated the safe in-place native refresh approach and expanded it with automatic native-catalog drift detection. The remaining same-version Desktop runtime replacement edge case is tracked separately in #627.

PR #622, opened 2026-09-05
feat: passively refresh mirrored native models

B1gM8c:
## Problem

An OpenAI Responses-compatible gateway can expose new native-compatible model
ids before codex-router's checked-in registry knows about them. The existing
refresh command updates the native catalog, but operators still have to notice
the provider change and curate another route manually. A periodic polling loop
would solve freshness at the cost of repeated network traffic and could restart
the router while Codex turns are active.

## Solution

- Allow an OpenAI Responses provider to opt in to private model discovery and
  native-model mirroring.
- Mirror only the exact intersection of the signed-in Codex catalog and the
  provider's live `/models` response. Generated entries carry an ownership
  marker; manually curated entries remain authoritative, and a transiently
  incomplete provider response never deletes a working route.
- Refresh passively from local lifecycle events rather than a polling loop.
  Router startup and the end of traffic schedule one check after 30 seconds of
  continuous idle time. A new request cancels that idle grace period.
- Coalesce triggers across processes with an owner-private durable reservation.
  Successful checks cool down for 24 hours; failures back off exponentially
  from 1 hour to a 24-hour cap. The worker is detached, time-bounded, and never
  performs provider discovery on the synchronous request path.
- Skip route publication and service restart when the mirrored intersection is
  unchanged. The explicit `refresh-catalog` command remains the immediate
  override and resets the cooldown.
- Preserve an exact boolean kill switch in the generated macOS, Linux, and
  Windows service definitions.

Codex still reads its picker catalog at application startup, so a newly
published model becomes visible after fully quitting and reopening Codex.

## Safety properties

- Provider opt-in is limited to the OpenAI Responses protocol.
- Scheduling state is bounded, schema-validated, mode `0600`, and written
  atomically without upstream errors, response bodies, URLs, or credentials.
- A reservation is persisted before network I/O, preventing restart storms if
  the worker crashes.
- Existing user routes and unrelated provider configuration are not rewritten.
- No private provider configuration or credentials are part of this branch.

## Verification

- `npm run check` — passed.
- Related regression suite — 174 passed, 1 platform skip, 0 failed.
- Passive-refresh adversarial suite — 12 passed, including 1,000 competing
  in-process triggers, 12 concurrent processes sharing one reservation, traffic
  cancellation, corrupt/symlink state, launcher failures, and secret-retention
  checks.
- Full `npm test` — 3,622 passed, 25 skipped. Two unrelated failures were
  investigated: the overloaded Antigravity timing test passed immediately in
  isolation; the browser-panel assertion expects English while the browser
  renders the host's Chinese locale, and the same assertion fails on a pristine
  `origin/main` checkout.
- Installed from this branch on macOS: service readiness passed and the router
  PID stayed stable. With active Codex traffic present, the idle gate correctly
  withheld the refresh worker.
- The complete PR diff passed targeted private-value and credential-shape
  scans.


B1gM8c:
Closing this superseded PR because a local installer-test fixture mutated the head worktree during pre-push and appended unrelated test commits. No private configuration or credentials were included; a clean replacement PR will follow.

PR #623, opened 2026-09-05
feat: passively refresh mirrored native models

B1gM8c:
## Problem

An OpenAI Responses-compatible gateway can expose new native-compatible model
ids before codex-router's checked-in registry knows about them. The existing
refresh command updates the native catalog, but operators still have to notice
the provider change and curate another route manually. A periodic polling loop
would solve freshness at the cost of repeated network traffic and could restart
the router while Codex turns are active.

A managed mirror also regenerated its full metadata record. If an operator had
live-tested one exact provider/model route and set
`supportsSearchHistory: true`, the next catalog refresh removed that proof.
Continuing a conversation containing a completed `web_search_call` then failed
locally with `model_search_not_supported`, even though that exact upstream
route accepted the history.

The native catalog can also expose the client-only `ultra` reasoning rung for a
model whose exact private route has been live-tested. Managed refreshes
previously removed that route-specific override, forcing the operator to add it
back after every metadata refresh.

## Solution

- Allow an OpenAI Responses provider to opt in to private model discovery and
  native-model mirroring.
- Mirror only the exact intersection of the signed-in Codex catalog and the
  provider's live `/models` response. Generated entries carry an ownership
  marker; manually curated entries remain authoritative, and a transiently
  incomplete provider response never deletes a working route.
- Refresh passively from local lifecycle events rather than a polling loop.
  Router startup and the end of traffic schedule one check after 30 seconds of
  continuous idle time. A new request cancels that idle grace period.
- Coalesce triggers across processes with an owner-private durable reservation.
  Successful checks cool down for 24 hours; failures back off exponentially
  from 1 hour to a 24-hour cap. The worker is detached, time-bounded, and never
  performs provider discovery on the synchronous request path.
- Skip route publication and service restart when the mirrored intersection is
  unchanged. The explicit `refresh-catalog` command remains the immediate
  override and resets the cooldown.
- Preserve an operator-supplied boolean `supportsSearchHistory` only on the
  same managed provider/model route. New models remain conservative: the flag
  is neither inferred from Responses compatibility nor copied from another
  model.
- Preserve an operator-added `ultra` rung only on the same managed route and
  only while the signed-in native catalog continues to advertise `ultra`. New
  mirrors stay conservative, and another route cannot donate this capability.
- Preserve an exact boolean kill switch in the generated macOS, Linux, and
  Windows service definitions.

Codex still reads its picker catalog at application startup, so a newly
published model becomes visible after fully quitting and reopening Codex.

## Safety properties

- Provider opt-in is limited to the OpenAI Responses protocol.
- Scheduling state is bounded, schema-validated, mode `0600`, and written
  atomically without upstream errors, response bodies, URLs, or credentials.
- A reservation is persisted before network I/O, preventing restart storms if
  the worker crashes.
- Existing user routes and unrelated provider configuration are not rewritten.
- Search-history replay and `ultra` remain exact-route, explicitly verified
  capabilities; a stale native `ultra` declaration is removed rather than kept.
- No private provider configuration or credentials are part of this branch.

## Verification

- `npm run check` — passed.
- Related regression suite — 174 passed, 1 platform skip, 0 failed.
- Passive-refresh adversarial suite — 12 passed, including 1,000 competing
  in-process triggers, 12 concurrent processes sharing one reservation, traffic
  cancellation, corrupt/symlink state, launcher failures, and secret-retention
  checks.
- Search-history/mirror focused suites — 15 passed, 0 failed: capabilities on
  a different route are not inherited, exact verified overrides survive managed
  metadata refresh, unsupported routes remain rejected, verified replay is
  forwarded, and `ultra` is withdrawn when native metadata no longer exposes it.
- GitHub Actions is green across the full Ubuntu, macOS, and Windows test
  suites, the unified native macOS lifecycle, both Electron packages,
  deterministic provider discovery, and GitGuardian.
- Installed from this branch on macOS: service readiness passed and the router
  PID stayed stable. With active Codex traffic present, the idle gate correctly
  withheld the refresh worker.
- The complete PR diff passed targeted private-value and credential-shape
  scans.

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,6 +593,29 @@ broader vendor profile just to repair `tool_choice`. The provider's own
 local to your machine and are not vetted by the repository's compatibility
 tests.
 
+A private OpenAI Responses gateway can opt into automatic native-model
+mirroring in its local registry fragment:
+
+```json
+{
+  "protocol": "openai-responses",
+  "allowPrivate": true,
+  "mirrorNativeModels": true
+}
+```
+
+`allowPrivate` is required only for a private or loopback discovery endpoint.
+During `./bin/refresh-catalog`, the router asks the signed-in Codex account for
+its current native catalog and asks each opted-in provider for `/models`. It
+automatically adds or updates only exact model-id matches that are visible to
+the account and advertised by that provider. Provider-only ids, hidden native
+models, and manually curated entries are left untouched. Automatically
+mirrored routes remain conservative: client-only `ultra` reasoning is not advertised
+for an automatically mirrored upstream whose `/models` response cannot certify
+it. A manually curated route remains authoritative. Codex still loads the
+resulting `model_catalog_json` at startup, so fully quit and reopen the app to
+see a newly mirrored model.
+
 The same managed OpenAI base URL also serves `/v1/embeddings`, but only for a
 model whose local or checked-in metadata explicitly names the capability. A
 model that is both conversational and embedding-capable declares its normal
--- a/README.md
+++ b/README.md
@@ -612,7 +612,11 @@ the account and advertised by that provider. Provider-only ids, hidden native
 models, and manually curated entries are left untouched. Automatically
 mirrored routes remain conservative: client-only `ultra` reasoning is not advertised
 for an automatically mirrored upstream whose `/models` response cannot certify
-it. A manually curated route remains authoritative. Codex still loads the
+it, and search-history replay is not inferred from another model or from
+Responses compatibility alone. After live-testing one exact managed route, an
+operator may set its `supportsSearchHistory` boolean in the protected
+`user-models.json`; later mirror refreshes preserve that route-specific proof.
+A manually curated route remains authoritative. Codex still loads the
 resulting `model_catalog_json` at startup, so fully quit and reopen the app to
 see a newly mirrored model.
 
--- a/README.md
+++ b/README.md
@@ -610,12 +610,14 @@ its current native catalog and asks each opted-in provider for `/models`. It
 automatically adds or updates only exact model-id matches that are visible to
 the account and advertised by that provider. Provider-only ids, hidden native
 models, and manually curated entries are left untouched. Automatically
-mirrored routes remain conservative: client-only `ultra` reasoning is not advertised
-for an automatically mirrored upstream whose `/models` response cannot certify
-it, and search-history replay is not inferred from another model or from
-Responses compatibility alone. After live-testing one exact managed route, an
-operator may set its `supportsSearchHistory` boolean in the protected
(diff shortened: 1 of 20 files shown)


PracticalSwan (changes requested):
Current `main` has moved materially in the exact catalog/refresh paths this PR changes, and GitHub correctly reports the branch as conflicting. This needs a semantic rebase before it is merge-ready.

I reproduced the current three-way conflicts locally against `origin/main=ea3dbb5` (merge base `5db9b31`). The important overlaps are not cosmetic:

- `src/catalog.mjs`: merged #624 added `nativeCacheCanRefreshInPlace()` / `nativeCatalogCanRefreshInPlace()` so signed-in refresh can use the separate native cache without rewriting `config.toml`; #627 then added same-version Codex runtime/binary fingerprint handling. The current conflict markers show the PR side would otherwise drop pieces of that newer authority/fingerprint logic.
- `src/refresh-catalog.mjs`: #624 added the signed-in in-place refresh path specifically to avoid unnecessary config transport rewrites / Windows sharing failures. Preserve that behavior when integrating this PR's live-account refresh and mirror logic.
- `src/router.mjs`: current main has #624's 5-minute `nativeCatalogDriftCheckTimer`, while this PR adds the event-driven idle `passiveCatalogRefresh` scheduler. The rebase should intentionally replace/retire the periodic poll when passive mirroring owns the same refresh responsibility; do not leave both mechanisms active and allow duplicate refresh/restart scheduling.
- `src/router.mjs` also contains merged #629 provider-affinity behavior for routed `collaboration.spawn_agent` calls. That must survive the rebase unchanged.
- The corresponding `test/catalog.test.mjs` and `test/refresh-catalog.test.mjs` conflicts need to retain #624/#627 regression coverage while adding the passive-mirror cases.

The core passive-refresh design remains compelling (idle gate, durable singleflight/cooldown, no synchronous provider discovery), but I cannot approve the stale branch because an incorrect conflict resolution can regress native catalog freshness, same-version runtime replacement detection, Windows config safety, or routed subagent billing boundaries.

After rebasing onto current main, please re-run at least the catalog/refresh/passive suites plus the full `npm test`/`npm run check` matrix. I can re-review the resolved integration.

duolahypercho:
Thanks for this — the idle-gated, single-flight refresh design is solid. Since it was opened, the native catalog refresh path has been reworked three times (#624, #635, and the client_version freshness fix for #645), so this branch would need a redesign on top of current main rather than a conflict resolution, and it would need to replace the 5-minute drift poll rather than run beside it.

Closing to keep the queue clean; a fresh PR against current main is very welcome, and the search-history preservation piece would be a good standalone PR on its own.

PR #630, opened 2026-09-05
fix(windows): preserve explicit service stop with heartbeat

PracticalSwan:
## Why

Follow-up to #618. The minute heartbeat correctly recovers a dead Windows router, but it also re-fires after an intentional `service stop` because the scheduled task remains enabled.

On a live Windows install I reproduced this sequence:

1. `service stop` returned `{"state":"stopped"}` and router health went down.
2. The scheduled task remained enabled.
3. The next minute heartbeat relaunched the router automatically.

That makes explicit stop non-persistent.

## Fix

- Disable the scheduled task before ending the active instance on explicit `stop`.
- Re-enable the task before `start` or `restart` runs it.
- Keep stop idempotent when the task is already missing.
- Leave the heartbeat itself unchanged, so crash/power-event recovery still works whenever the service is intentionally enabled.

## Verification

- Added a regression that fails on current `main` before the fix.
- `node --test test/service-render.test.mjs` — pass on Windows (platform-safe scheduler stub cases skip on win32 by repository design).
- Adjacent service/process/task-state/hidden-console tests — 35 total, 19 pass / 16 intentional platform skips, 0 fail.
- `npm run check` — pass.
- `git diff --check` — pass.
- Live Windows validation of the same service semantics:
  - after `stop`: task `Disabled`, no router start process, health down;
  - after waiting 63 seconds across the heartbeat boundary: task still `Disabled`, no router process, health still down;
  - after `start`: task `Running`, two triggers present, `StartWhenAvailable=True`, router health restored;
  - Codex `config.toml` hash remained unchanged.

This preserves #618's self-healing behavior without overriding an explicit user stop.

### DIFF
--- a/src/service-windows.mjs
+++ b/src/service-windows.mjs
@@ -394,6 +394,13 @@ function taskExists() {
   }
 }
 
+function setTaskEnabled(enabled) {
+  schtasks(
+    ["/Change", "/TN", taskName, enabled ? "/ENABLE" : "/DISABLE"],
+    { quiet: true, mutating: true },
+  );
+}
+
 function taskState() {
   const script =
     "try { [Console]::Out.Write((Get-ScheduledTask -TaskName $env:CODEX_ROUTER_TASK).State.ToString()) } catch { exit 1 }";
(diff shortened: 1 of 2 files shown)


PracticalSwan:
Fresh maintenance validation after Codex Desktop updated to 26.901.5280.0 / codex-cli 0.153.4: the installed Windows router service remains healthy, with two scheduled-task triggers and StartWhenAvailable enabled. The historical 0x800710E0 LastTaskResult continues to correspond to an IgnoreNew duplicate-trigger refusal while the task is already Running, not a dead router. I did not alter the live scheduled task. PR #630 remains relevant specifically to preserving an explicit stop across heartbeat triggers.

PracticalSwan:
Additional note: this validation was read-only with respect to the scheduled task itself; no stop/start/reinstall was performed during the check.

PR #631, opened 2026-09-05
test: make Control Center locale deterministic

tanakaduxi:
## Problem

On hosts whose browser locale is not English, renderer tests can render
translated Control Center UI while continuing to locate English labels.
For example, a Japanese locale causes English role and text locators to
time out even though the product is behaving as configured.

## Root cause

The Control Center chooses its initial language from the stored
`codex-router-language` preference, then `navigator.language`, and finally
falls back to English. The renderer test pages did not set either input,
so their language depended on the host.

## Fix

Create renderer test pages through a small helper that stores the English
Control Center preference with `page.addInitScript()` before navigation
and application initialization.

The existing assertions, selectors, timeouts, and product source are
unchanged.

## Validation

- Focused renderer regressions: 3/3 passed
- Locale regression check: passed
- `npm run check`: passed
- `npm run build`: passed
- `npm test`: 96 passed, 0 failed, 1 documented skip

### DIFF
--- a/apps/control-center/test/renderer.test.mjs
+++ b/apps/control-center/test/renderer.test.mjs
@@ -570,6 +570,14 @@ const chromiumPath = [
   process.env.LOCALAPPDATA && path.join(process.env.LOCALAPPDATA, "Google", "Chrome", "Application", "chrome.exe"),
 ].find((candidate) => candidate && existsSync(candidate));
 
+async function newEnglishTestPage(browser, options) {
+  const page = await browser.newPage(options);
+  await page.addInitScript(() => {
+    localStorage.setItem("codex-router-language", "en");
+  });
+  return page;
+}
+
 test("the production renderer exposes model discovery and picker actions", { timeout: 120_000 }, async () => {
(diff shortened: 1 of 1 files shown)


PracticalSwan (approved):
Validated independently on Windows 11 against head `55f5b13`.

The change is scoped to the renderer test harness and correctly installs the `codex-router-language=en` preference with `page.addInitScript()` before product scripts run on navigation.

Local verification in a clean detached worktree using the PR's committed Control Center lockfile:
- `npm --prefix apps/control-center run check` — pass
- `npm --prefix apps/control-center run build` — pass
- `node --test apps/control-center/test/renderer.test.mjs` — 6/6 pass

I then temporarily forced every test page in the disposable worktree to Playwright locale `ja-JP` while retaining this PR's init script. The same renderer suite still passed 6/6, which directly exercises the locale-independence this PR is intended to provide. The temporary locale forcing was reverted and the worktree was clean afterward.

GitHub Actions currently shows `action_required` rather than a test failure, consistent with an external-contributor workflow awaiting approval. I found no code issue in the proposed test-only change.

PR #632, opened 2026-09-05
Fix routed subagents crossing provider billing boundaries

B1gM8c:
## Summary

- pin every in-session `collaboration.spawn_agent` call to the routed parent model
- keep explicit model selection unchanged for independent `codex_app.create_thread` tasks
- document the provider-affinity boundary and add native/flattened response regressions

## Bug

PR #629 fixed the omitted-model path: a routed parent that emitted `spawn_agent` without `model` now passed its routed model to the child.

A routed parent can also emit a valid explicit native model, for example:

```json
{
  "task_name": "review",
  "model": "gpt-5.6-sol"
}
```

Because the prior implementation treated every valid explicit value as authoritative, Codex created the child on the native OpenAI provider even though the parent was running through a private routed provider. This silently crossed the provider and billing boundary and consumed the user's native Codex quota.

The `model` field in an in-session subagent tool call is model-generated output, not a separate user-visible model choice. It therefore should not override the parent provider affinity. Independent new tasks remain different: an explicit model on `codex_app.create_thread` is still preserved.

## Fix

For routed responses, `collaboration.spawn_agent` now always receives the exact routed parent slug. The rewrite covers:

- flattened Chat Completions bridge calls (`collaboration__spawn_agent`)
- native Responses namespace calls (`namespace: collaboration`, `name: spawn_agent`)
- omitted, unsupported, and valid-but-cross-provider model values

Malformed or ambiguous payloads continue to fail open without byte rewriting, and cloud/new-thread/follow-up behavior is unchanged.

## Verification

- `node --test test/subagent-model-inherit.test.mjs test/namespace-relay.test.mjs` — 130 passed
- `node --test --test-name-pattern='pins subagents' test/routing.test.mjs` — 1 passed
- `node --test test/routing.test.mjs` — 116 passed
- `npm run check` — passed
- `git diff --check` — passed

The local full-suite pre-push hook also ran. It hit unrelated host-dependent failures in the browser panel, installer rollback fixture, and process-tree timing; the focused routing suites above passed, and GitHub CI should provide the clean-environment result.

No provider credentials, private endpoints, tunnel identifiers, or local configuration files are included.


PracticalSwan:
Fresh Windows validation against current `main`/the locally adapted #629 behavior confirms the remaining provider-affinity hole this PR fixes.

Environment:
- Codex Desktop `26.901.5280.0`
- bundled `codex-cli 0.153.4`
- Windows 11 x64

Before applying the #632 affinity logic, both forms below preserved a valid explicit native model emitted by the parent:
- flattened `collaboration__spawn_agent`
- native `{ namespace: "collaboration", name: "spawn_agent" }`

With routed parent `opencode-go/deepseek-v4-flash` and tool argument `model: "gpt-5.6-sol"`, `injectSessionModelForSpawnCalls(...)` returned the original object unchanged in both cases. That leaves a real path for an in-session routed subagent to cross back to the native provider/quota.

After applying the pinning logic, both forms rewrite to the routed parent model, while an explicit model on `codex_app__create_thread` remains untouched. Local verification:
- `node --test test/subagent-model-inherit.test.mjs test/namespace-relay.test.mjs` — 129/129 pass
- `node --test --test-name-pattern='pins subagents' test/routing.test.mjs` — 1/1 pass
- `npm.cmd run check` — pass
- `git diff --check` — pass

I also activated the same focused behavior in the live local router after confirming `health.activity.state=idle` / `activeCount=0`; the supervised restart came back healthy on a new PID and the Codex config was unchanged by that restart.

This looks like a genuine provider/billing-boundary fix rather than only a model-selection preference. The PR's own CI matrix is also fully green across Windows, Linux, macOS, Electron, Unified native macOS, and provider discovery.

PR #634, opened 2026-09-05
feat: add Muse Spark OpenCode free model

kimparrot:
OpenCode's anonymous Zen catalog exposes Muse Spark 1.3 Contributor Free, but codex-router does not currently surface it in the native model picker. This adds the exact Responses route as a listed, credential-free model with xhigh as its highest/default picker effort, the documented context/compaction limits, and no native apply_patch claim.

The curation guard keeps Muse's model-specific `auto-tool-choice` repair from being inherited by unrelated models in OpenCode's mixed free catalog.

Validation:
- `node --test test/gemini-3.8-muse-1.3-fable-5.1.test.mjs test/registry.test.mjs test/curate-models.test.mjs` (111/111)
- `npm run check`

### DIFF
--- a/config/opencode-free/responses/muse-spark-1.3-contributor-free.json
+++ b/config/opencode-free/responses/muse-spark-1.3-contributor-free.json
@@ -1,0 +1,49 @@
+{
+  "version": 1,
+  "models": [
+    {
+      "slug": "opencode-free-responses/muse-spark-1.3-contributor-free",
+      "gatewayModel": "opencode-free-responses-muse-spark-1-3-contributor-free",
+      "upstreamModel": "muse-spark-1.3-contributor-free",
+      "provider": "opencode-free-responses",
+      "listed": true,
+      "displayName": "Muse Spark 1.3 Contributor (OpenCode Free)",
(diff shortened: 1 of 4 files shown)


PracticalSwan (approved):
Independent current-main validation on Windows looks good. I trial-merged this head onto current `origin/main` (`a2f58d7`) with no conflicts. OpenCode's live anonymous `/zen/v1/models` catalog currently exposes both `muse-spark-1.3` and the exact `muse-spark-1.3-contributor-free` id, and the current Zen docs list the Contributor Free route on `/v1/responses` as free for a limited time. Current main already allowlists that exact Responses id under `opencode-free-responses` (949ae39), so the route passes the anonymous-provider boundary.

Local verification on the merged tree:
- `node --test test/registry.test.mjs test/gemini-3.8-muse-1.3-fable-5.1.test.mjs` — 60/60 pass
- `node --test test/curate-models.test.mjs test/model-discovery.test.mjs test/provider-catalogs.test.mjs` — 70/70 pass
- `npm run check` — pass
- `git diff --check --cached` — pass

The curation change also preserves explicit Muse model profiles while preventing that model-specific profile from being projected onto unrelated Zen Free models. I found no blocking issue in this diff.

PR #638, opened 2026-09-06
fix: repair routed agents on startup

PracticalSwan:
## Problem

Codex Router publishes managed `router-model-*.toml` definitions alongside the routed model catalog, and doctor checks them as one coupled state. After a Windows reboot I reproduced a state where the router service, routed catalog, gateway routes, provider selection, and Codex config were all healthy, but the expected `zai-coding/glm-5.3` managed agent definition was missing (`0 of 1 current definitions`).

A normal `catalog.mjs` publication immediately restored the definition without changing `config.toml`, and the repaired file stayed stable during runtime. The startup reconciliation currently republishes only when native model metadata drifts, so a current native catalog plus drifted/missing managed-agent files is not self-healed.

## Fix

- Extend the existing post-health startup reconciliation to detect routed-agent catalog drift as well as native-model drift.
- Reuse the same effective model/settings calculation as doctor/catalog:
  - selected configured listed models
  - current picker visibility
  - multi-agent settings and proven capabilities
  - `subagentEligibleModels()`
  - `routedCodexAgentStatus()`
- Republish the installed picker when either native catalog drift or routed-agent drift exists.
- Treat inactive/foreign router transport and unreadable configuration as non-actionable: startup remains write-free rather than guessing.
- Keep the current behavior unchanged when both native and routed-agent state are current.

This does not add another writer; repair goes through the existing target picker publication and catalog lock.

## Verification

- Focused native-drift + managed-agent tests: **12/12 pass**.
- Adjacent lifecycle suite (`native-catalog-drift`, `codex-agent-catalog`, `refresh-catalog`, `doctor-routing-mode`): **22/22 pass**.
- `npm run check`: pass.
- `git diff --check`: pass.
- Real local recovery: a normal catalog publish restored the missing definition to **1/1 current** without modifying `config.toml`, and a 120-second watcher observed no subsequent deletion or config mutation.

No provider inference request was used for this verification.


PracticalSwan:
Refreshed this branch onto current upstream `main` (`d1805f635de365761a7ff515fff94865567649b4`) using a normal merge commit, with no rebase/force-push. New head: `d8f7a3296aaafb67cc2cc56fcc1f9fa41ae654a5`.

Post-refresh local verification:
- focused startup native/routed-agent drift regressions: 3/3 pass
- 9/9 agent-catalog tests completed successfully before the broader multi-file harness stalled in teardown; no test assertion failed
- `git diff --check origin/main...HEAD`: clean
- PR diff remains isolated to the original 4 files

Provider model discovery is green on the refreshed head; the main CI workflow is queued at the time of this comment.

PracticalSwan:
Final verification found and corrected one issue in the startup-repair implementation before merge: the first cut checked only `# BEGIN codex-router`, while the real config manager writes managed markers such as `# BEGIN codex-router-managed` / `# BEGIN codex-router-provider-managed` (and the Kimi proxy variant). That could make installed-state detection silently return false.

Current head `ee68915ed70b02f6e9e012c8dee3406b2346424b` now uses the same managed-marker compatibility pattern as the integration surface and adds a regression covering the real marker shapes.

Post-correction verification:
- focused `native-catalog-drift.test.mjs`: **4/4 pass**
- real current `config.toml` marker probe: **true**
- `npm run check`: pass
- `git diff --check`: pass
- current upstream `main` (`5f8a7cd`) merged normally; no rebase/force-push
- PR remains exactly **4 changed files**
- refreshed CI: **success**
- refreshed Provider model discovery: **success**

PracticalSwan:
Fresh live-state recurrence while this PR is pending: the installed router stayed healthy (`/health ok=true`, no degraded services, routed catalog/gateway still 4/4) but the managed routed-agent definition drifted to **0/1 current** again. I could not attribute the writer confidently, so I am not claiming a new test defect from timing alone.

A normal catalog publication restored **1/1 current** immediately (`routed_agents: 1`, `removed_agents: 0`) with `config.toml` SHA-256 unchanged before/after (`DAD0821170F1D2CC8A2859EE9D59D60FB182A6BA169BDC08DFCB8A4E2A72EB6F`), and doctor returned to exit 0 without restarting Codex or the router.

This is another direct reproduction of the exact state #638 is intended to self-heal: healthy router/catalog plus independently drifted managed agent files.

PR #643, opened 2026-09-06
fix(commandcode): bound tool names and repair recursive schemas

### LINKED ISSUE #626: Command Code rejects routed turns with >64-character tool names or recursive JSON schemas
## Summary

A Codex turn routed to `commandcode/meta/muse-spark-1.3-contributor` is rejected before generation when the attached tool surface contains a function name longer than 64 characters.

Observed response:

```text
HTTP 400: `name` must be at most 64 characters, got 80
```

The offending provider-facing tool in this request was:

```text
mcp__openai_api_key_local_confirmation__confirm_openai_api_key_local_destination
```

Its length is exactly 80 characters.

## Root cause

`buildRoutedRequest()` sends chat-completions providers through `chatProviderToolSurface()`. That function flattens namespaces, but currently calls `flattenNamespaceTools(merged.tools)` without a name-length bound for Command Code.

The router already has deterministic bounded aliases and reverse restoration in `flattenNamespaceTools(..., { maxNameLength: 64 })`; Command Code is simply not opting into that existing path.

This reproduces on current `origin/main` at `0f8441f3c9f5d66afda2e92da2023fbab7d30f98`.

## Suggested fix

For provider IDs `commandcode` and `commandcode-messages`, call:

```js
flattenNamespaceTools(merged.tools, { maxNameLength: 64 })
```

Keep other non-Groq providers byte-compatible with the existing unbounded behavior.

A regression test should assert that:

- the exact 80-character tool is exposed to Command Code under a 64-character deterministic alias;
- the alias differs from the client name;
- `rewriteNamespaceResponsePayload()` restores a returned function call to the original client identity.

## Validation

A local two-file patch using the existing alias machinery passes:

```text
node --test test/chat-tool-surface.test.mjs test/commandcode-generate.test.mjs test/commandcode-relay.test.mjs
32 tests passed, 0 failed
```

The broad suite also exercises this regression successfully. Its unrelated existing Search Sidecar cancellation still reproduces when run alone.

duolahypercho:
Closes #626.

## Problem

A Codex turn routed to a Command Code model is rejected before generation when the attached tool surface contains a function name longer than 64 characters:

```text
HTTP 400: `name` must be at most 64 characters, got 80
```

The reported tool was `mcp__openai_api_key_local_confirmation__confirm_openai_api_key_local_destination` — exactly 80 characters.

`buildRoutedRequest()` sends chat-completions providers through `chatProviderToolSurface()`, which flattened namespaces with no name-length bound for Command Code. The router already has deterministic, reversible bounded aliases behind `flattenNamespaceTools(..., { maxNameLength: 64 })`; Command Code simply was not opting into that existing path.

Behind that error sits a second one from the same validator: `Recursive JSON schemas are not currently supported`. Both had to be fixed for the turn to complete.

## Fix

- `chatProviderToolSurface()` sends `commandcode` and `commandcode-messages` through the existing 64-character bounded alias route. Every other non-Groq provider keeps its tool surface byte for byte.
- Both variants join the non-recursive schema repair. The gate is provider-wide rather than per model because the validator is the endpoint's, not the model's.
- `needsNonRecursiveOpenCodeToolCompatibility` is renamed to `needsNonRecursiveToolSchemaCompatibility`: it no longer names one vendor now that a second, unrelated validator shares it.

The aliases stay deterministic and reversible, so `rewriteNamespaceResponsePayload()` restores a call the model makes under the bounded spelling to the client's own tool identity.

## Tests

Five regression tests in `test/chat-tool-surface.test.mjs`, covering both provider variants:

- the exact 80-character tool from the report reaches the provider under a 64-character alias;
- the alias differs from the client name and round-trips back to it through `rewriteNamespaceResponsePayload()`;
- the alias is stable across identical surfaces;
- a non-Command Code chat provider still receives the unbounded 80-character name.

Full suite: **3668 tests, 0 failures** (26 skipped, all platform-gated).

## Credit

The diagnosis, the root cause, and the shape of this patch are [@erc-dev-pm](https://github.com/erc-dev-pm)'s, recorded in #626 after they were unable to push their local commits. They verified the fix end to end against the live route (`codex exec -m commandcode/meta/muse-spark-1.3-contributor` completing normally). This PR implements that patch with tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+- **Command Code no longer rejects a routed turn over a long tool name or a
+  recursive schema.** A Codex turn carrying a client tool such as
+  `mcp__openai_api_key_local_confirmation__confirm_openai_api_key_local_destination`
+  (80 characters) was refused before generation with ``HTTP 400: `name` must be
+  at most 64 characters, got 80``, and the turn behind it then hit
+  `Recursive JSON schemas are not currently supported` (issue #626).
+  `chatProviderToolSurface()` now sends both `commandcode` and
+  `commandcode-messages` through the router's existing bounded alias route at
+  64 characters, and both variants join the non-recursive schema repair. The
+  aliases stay deterministic and reversible, so a call the model makes under
+  the bounded spelling is restored to the client's own tool identity. Every
+  other non-Groq provider keeps its tool surface byte for byte.
+
 - **The macOS tray no longer spawns a Node process every second to read
   health.** `refreshActivity()` polls health once a second and ran
   `bin/control health --json` each time, which boots Node and control.mjs's
--- a/src/chat-tool-surface.mjs
+++ b/src/chat-tool-surface.mjs
@@ -212,13 +212,31 @@ function withRequiredAppTools(tools, required) {
 // definition is required; those definitions are added back before the request
 // is admitted. A client surface (or client + required definitions) over the cap
 // is refused locally rather than truncated.
+// Command Code validates the provider-facing tool `name` at 64 characters and
+// refuses the whole request over a longer one, so a client tool such as
+// `mcp__openai_api_key_local_confirmation__confirm_openai_api_key_local_destination`
+// (80 characters) fails the turn before generation (issue #626). Both provider
+// variants front the same validator. They opt into the router's existing
+// bounded alias route, which is deterministic and reversible, so
+// `rewriteNamespaceResponsePayload()` still restores the client's own identity.
+// Every other non-Groq provider keeps the unbounded surface byte for byte.
+const BOUNDED_TOOL_NAME_PROVIDERS = new Set(["commandcode", "commandcode-messages"]);
+const BOUNDED_TOOL_NAME_LENGTH = 64;
+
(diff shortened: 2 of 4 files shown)


PracticalSwan (changes requested):
Blocking request-shape mismatch remains for forced Command Code tool choices.

The new bounded-name path correctly aliases the reported 80-character tool definition to a deterministic 64-character provider name, but `buildRoutedRequest()` only calls `flattenToolChoice()` inside the `provider?.id === "groq"` branch (plus the separate Console-Go Responses compatibility path). `commandcode` / `commandcode-messages` therefore send the aliased definition while a forced `tool_choice` can still carry the original long name.

I reproduced this against head `9f0e84e0f6ae43acace74b29bfcb892be8cd59e6`: `chatProviderToolSurface(..., "commandcode")` produced `mcp__openai_api_key_local_confirmation__confirm_ope_f06b659abc33` (64 chars), while the original forced choice remained the 80-character client name; explicitly applying `flattenToolChoice()` produces the matching 64-character alias.

Please apply the same request-local alias mapping to Command Code `tool_choice` (including `allowed_tools`) and add a regression that exercises the actual routed request path, not only definition aliasing/response restoration.

PR #644, opened 2026-09-06
fix(dashboard): total 24h traffic from every provider, not a capped sample

duolahypercho:
## Problem

The Dashboard reported two different totals for the same 24-hour window, on the same screen:

| Element | Showed | Source |
|---|---|---|
| "Tokens, last 24h" tile | **991,777,637** | sums all 42 provider rows ✓ |
| 24H traffic chart caption | **68,702,024** | newest 1,000 events ✗ |
| actual traffic in that window | 992,005,114 across 9,033 events | |

`buildHourlyTrafficBuckets` was built entirely from the probe's `usageEvents`, which `recentUsageEvents()` caps at 1,000 rows. An ordinary busy day is many times that — 9,000 events in 24 hours is routine — so the newest 1,000 covered **under two hours**. Twenty-two of the twenty-four bars were drawn empty, and the caption printed that truncated sum with the words "in the last 24 hours" next to a tile that had added up every provider row.

The 7D/30D views never had this bug: `buildDailyTrafficBuckets` aggregates `retained.providers[].dailyUsageBuckets`. Only the hourly view was missing an equivalent aggregation.

## Fix

`hourlyUsageRollup()` aggregates the **uncapped** window into 24 fixed buckets, published on the probe as `usageEventHours`, and the chart is built from it.

- **Cheaper as well as correct.** 24 buckets serialize to 5 KB; the capped event sample is 233 KB.
- **Bars carry the rollup's own hour stamps** rather than a grid anchored on the renderer's clock, so the newest measured hour stays on the chart when the hour turns over between the snapshot and the render.
- **One ledger read per probe.** `emitProbe()` reads the window once and derives both views — the tray polls it constantly and the ledger is large, so the rollup must not cost a second pass.
- **The rollup is also the middle rung of the summary tile's fallback**, between provider rows and the capped sample, so a router without provider counters no longer understates a busy day by an order of magnitude.
- **The capped sample remains the fallback** for a router that predates the rollup, and the caption says which one is in use.

The rollup's token math mirrors the renderer's `tokenCountFromEvent` / `trafficPartsFromEvent` exactly — billed counts preferred over raw, cached share clamped to the input it belongs to, an unreported field kept distinct from a measured zero — and a test pins that.

## Verification

Against the affected install, the same window now reads **1,001,119,088 tokens across 9,101 requests over 21 non-empty hours**, in agreement with the provider-row tile, where the chart previously said 68.7M across 2 hours.

- Full suite: 3,672 tests, **3,646 pass, 0 fail**, 26 skipped
- `tsc --noEmit` clean
- New tests: rollup aggregates past the 1,000-row cap and requests the window uncapped; token/cache-share math matches the renderer's; an empty ledger still returns a full, honest set of hours; contract tests pin the renderer preference order and the single-read probe

### DIFF
--- a/apps/control-center/src/pages/DashboardPage.tsx
+++ b/apps/control-center/src/pages/DashboardPage.tsx
@@ -33,6 +33,7 @@ import type {
   RouterHealth,
   RouterTarget,
   UsageEvent,
+  UsageEventHour,
   UsageBucket,
   UsageMetric,
   ViewId,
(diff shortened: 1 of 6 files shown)


PracticalSwan (changes requested):
The new hourly rollup still does not match the existing true rolling-24h provider totals.

`hourlyUsageRollup()` anchors 24 buckets at the start of the current hour: `first = currentHourStart - 23h`. That covers only 23h + the elapsed portion of the current hour. At 17:30, the oldest included timestamp is 18:00 the previous day, whereas `provider-usage.mjs` correctly uses `now - 24h` (17:30 the previous day).

Direct reproduction on head `26a4c1660ec608dd024634e8e94993a8601e0466`: with `now = 2026-09-06T17:30:00Z`, an event at `2026-09-05T17:45:00Z` is only 23h45m old and belongs in the rolling 24h window, but the rollup's first bucket is `2026-09-05T18:00:00Z` and the result is 0 requests / 0 tokens.

Please preserve a true `now - 24h` accounting boundary. One workable shape is to retain hourly display buckets while including the oldest partial-hour slice in the first bucket (and clipping future/current-hour edges to `now`). Add a boundary regression for an event between `now - 24h` and the first whole-hour boundary.

PR #650, opened 2026-09-07
fix(commandcode): align forced tool choice with bounded aliases

PracticalSwan:
## Problem

Follow-up to merged #643. Command Code now bounds provider-facing tool names to 64 characters, but `buildRoutedRequest()` only rewrote `tool_choice` aliases for Groq (and the separate Console Go path). A forced Command Code choice could therefore keep the client's original 80-character name while the advertised tool used a hashed <=64-character alias, producing two identities in one request.

## Fix

- When namespace/tool names are flattened for `commandcode` or `commandcode-messages`, pass object `tool_choice` through the same reversible alias map as the tool list and stored history.
- Keep all other provider behavior unchanged.
- Add an end-to-end router regression for both Command Code provider variants using the exact 80-character tool from #626/#643.

## Verification

After normally merging current upstream `main` (`c2b2bd1`, including #658, #659, and #660) into this branch without rebasing or force-pushing:

- `test/namespace-relay-routing.test.mjs`: **23/23 pass**, including the forced Command Code alias regression.
- `node scripts-check.mjs`: pass (`v2-agent applications valid (6)`, syntax checks passed).
- `git diff --check` against current `main`: pass.
- Resulting PR diff remains isolated to the original three files.
- Current branch head: `cb9991d66120880a4ab0203547037d9923838d3b`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+- **Command Code forced tool choices now use the same bounded alias as the tool definition.**
+  The 64-character compatibility added in #643 shortened provider-facing tool names but
+  left an object `tool_choice` at the client's original spelling, so a forced long tool
+  could still be rejected as unknown. Forced choices for both Command Code variants now
+  pass through the same reversible namespace alias map as the advertised tools.
 - **Command Code no longer rejects a routed turn over a long tool name or a
   recursive schema.** A Codex turn carrying a client tool such as
   `mcp__openai_api_key_local_confirmation__confirm_openai_api_key_local_destination`
--- a/src/router.mjs
+++ b/src/router.mjs
@@ -3171,11 +3171,15 @@ async function buildRoutedRequest({ request, payload, route, agedInput }) {
       tools = repairToolSchemaRoots(tools, { nonRecursive: true });
     }
   }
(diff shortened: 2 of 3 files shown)


PracticalSwan:
CI note after rebasing-by-merge onto current `main` (`c6e7f5d`): the refreshed head is still exactly the intended 3-file Command Code change, and the root Windows suite (`test (windows-latest)`) completed successfully. The only red job is `Electron control center (Windows)`, which #650 does not modify.

The failing assertion was the existing `trusted install provenance overrides contradictory package-manager environment` test, timing out at ~30.47s. On #651/current-main, the same test on the same Windows Server 2025 / `windows-2025-vs2026` image and Node 24.19.0 passed in ~12.99s, and that Electron job completed 97/97 non-skipped tests successfully. I also ran that exact test six times on a clean detached current-main worktree on Windows; all 6 passed, with the repeated runs taking ~1.37-1.42s each. This therefore appears to be a Windows timing flake rather than a regression from #650.

Additional refreshed-head verification: `test/namespace-relay-routing.test.mjs` 23/23 passed locally, including the new Command Code forced-choice regression; direct `node scripts-check.mjs` passed. I attempted to rerun only the failed Actions job, but the linked GitHub integration lacks Actions write permission (403), so no rerun was started.

PR #652, opened 2026-09-07
fix(grok-oauth): preserve worker progress and reject unsuccessful turns

Leshabeats:
After a tool result, a Grok repair could contain a complete-looking tool call and then fail, become incomplete, or end without a successful response terminal. The forwarder could still release the held action or private final answer. LiteLLM can also translate a forwarder stream error into an untyped error envelope followed by empty message closes, obscuring the failure at the Responses boundary.

Require a successful response terminal before releasing held repair output, and normalize recognized Grok gateway error envelopes to one safe terminal error. Reasoning stays live; unsuccessful first attempts are not repaired or replayed, and failed repairs do not release held client actions. The Grok normalizer closes repaired reasoning as incomplete, drops subsequent frames, and runs before usage observation so activity records the failure.

Add capability-protected `GET /v1/activity` and `bin/control activity [thread-id]` for diagnosing quiet routed workers. Live requests remain visible until handler cleanup, independently of tray retention; the snapshot exposes dispatch counts, byte/event timestamps, observed phase, cancellation cause, and bounded recent outcomes. Shipped agent guidance distinguishes a polling timeout from an execution deadline and requires checking the existing worker before replacement.

### Validation

Current head: `f00a8382b91c0315e47167b09e1e65a9f2a24818`, refreshed onto upstream `7e0a2b72ebb85ed497d9df8529ac251262261d51`. Conflict resolution was checked with range-diff. `npm run check`, `git diff --check` and 143 focused local tests pass. Full Linux, macOS and Windows CI passes on this head. The full-suite evidence for the refreshed revision is hosted CI; earlier pre-rebase local results are not substituted for it.

The focused tests cover successful/failed terminal gating, normalized gateway errors, activity lifecycle, cancellation, no replay after client output, and capability-protected metadata without content.

### Scope and limits

- Activity describes Router HTTP requests, not native Codex worker lifecycle or retries inside LiteLLM/xAI. An empty/unknown snapshot is not evidence that a worker finished. Uninstrumented subpaths remain `unobserved`.
- No prompts, response text, tool arguments, or credentials are retained in activity. `/health` retains its existing contract; activity requires the caller capability.
- The existing bounded-parser fallback for oversized or invalid UTF-8 frames before stream mutation is preserved. Other providers retain their existing stream transforms.
- Live pilot timings belong to downstream stacked variants, not an isolated evaluation of this PR. No generation-speed improvement is claimed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+- **Routed workers expose read-only request progress.** `bin/control activity
+  [thread-id]` reports active requests and bounded recent outcomes behind the
+  caller capability, including observed stream events and cancellation causes.
+  Quiet polling never ends a request, and the shipped agent guidance separates
+  polling intervals from worker deadlines.
+- **Grok repairs require a successful response terminal.** Failed, incomplete,
+  and truncated streams return a terminal error without releasing withheld
+  client actions or private final answers. Reasoning remains live.
 - **The ChatGPT Web provider is removed: using it risked an OpenAI account
   ban.** `chatgpt-web` routed Codex turns into an unofficial browser automation
   of chatgpt.com, driven through a separately installed launcher on loopback
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 - **Grok repairs require a successful response terminal.** Failed, incomplete,
   and truncated streams return a terminal error without releasing withheld
   client actions or private final answers. Reasoning remains live.
+- **Grok gateway stream errors reach Codex as terminal failures.** Untyped
+  gateway errors are normalized without exposing upstream diagnostics or
+  appending empty message closes, and request activity records the failure.
 - **The ChatGPT Web provider is removed: using it risked an OpenAI account
   ban.** `chatgpt-web` routed Codex turns into an unofficial browser automation
   of chatgpt.com, driven through a separately installed launcher on loopback
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -393,6 +393,24 @@ is correct." is not talked into a call the client would then run. Raise
 `CODEX_ROUTER_GROK_PROGRESS_ONLY_MAX_TEXT` to fire less often on that
 user-message path; those settings do not weaken the post-tool invariant.
 
+For a quiet worker, run `bin/control activity <thread-id>` from the installed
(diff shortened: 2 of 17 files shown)


duolahypercho:
Thanks @Leshabeats. The terminal gating is a real safety fix: a failed or truncated repair should never release a held client tool, and the terminal matrix in the tests is thorough. Some changes are needed before this can merge. I checked each item against head `f00a8382`:

**Blocker**

1. **Stop accepting events after `response.completed` too** (`src/grok-oauth-turn.mjs`, `applyResponsesEvent`). The guard only ignores later events after `failed`/`incomplete`. A `response.output_item.done` that arrives after `response.completed` is still added to `toolCalls`, and the turn still reads as `completed`, so `classifyAfterToolRepair` releases it. That breaks the PR's own rule that nothing is released without a successful terminal event. Please stop accepting events after any terminal status, and add the `completed` case to `an upstream failure cannot be revived by later completed or tool events`.

**Activity diagnostics** (still open from the earlier Codex review)

2. `src/request-progress.mjs` `event()` classifies terminals by `type` only. A `response.completed` event with `response.status: "failed" | "incomplete"` is recorded as a successful request, while the Grok turn code in this same PR does read the inner status.
3. The single routed transport retry (the `routed transport retry 1/1` branch from #659) calls the global `fetch`, not `fetchObservedUpstream`, so `upstreamAttempts` and `lastHeadersAt` are wrong on that path.
4. Only `handleResponses` calls `progress.cancel("client_disconnected")`. Native, search, image and embeddings requests also go through `beginRequestActivity`, but a cancelled one stays `running` and can later be put down to `execution_deadline`.
5. `skills/codex-router/SKILL.md` tells every agent to run `~/.local/share/codex-router/bin/control activity`. On Windows the checkout is `%LOCALAPPDATA%\codex-router`, and `codex-router.ps1` only accepts a fixed list of commands that doesn't include `activity`, so the documented diagnostic can't be run there.

**Scope**

Could you split this into two PRs? The Grok terminal gating and gateway error normalization (`grok-oauth-*`, `grok-reasoning-summary-compat`, their tests) can land as soon as item 1 is fixed. `/v1/activity`, `control activity`, and the new `SKILL.md` guidance add a new diagnostics API and change the behaviour of every user's agents ("do not spawn a replacement…"). That deserves its own review once items 2–5 are addressed.

The only merge conflict with current `main` is `CHANGELOG.md`.

duolahypercho:

PR #662, opened 2026-09-08
fix(control-center): keep loading when Codex usage probe exits

slavakurilyak:
## Summary

- Control Center's ChatGPT usage read spawned `codex app-server` and treated process exit as a hard failure.
- If the app-server died after one of the two account replies, Electron painted the Node stack as "Some router data could not load" over an otherwise healthy snapshot.
- Keep a partial usage slice when any account reply landed, and return an empty unavailable payload from `control account` instead of crashing.

## Root cause

`readCodexAccountUsage` waits for both `account/rateLimits/read` and `account/usage/read`. The Codex app-server can answer one, then exit 1 before the other. That rejection escaped `printAccountUsage`, Node dumped the exception to stderr, and the Control Center IPC layer forwarded the stack into the global warning banner.

JSON-RPC `error` replies were already treated as optional. Process exit was not. Same class of bug as the earlier rate-limits hard-fail that painted the Models page.

Reproduced on macOS with Codex Router Control Center against `codex-cli 0.153.4`, ChatGPT Plus, and an isolated profile home. Direct `node src/control.mjs account` succeeded later; the banner is the transient/exit path, not a missing login.

## Validation

- `node --test test/codex-account-usage.test.mjs test/chatgpt-account-control.test.mjs test/discovery-mode.test.mjs` (19 passed)
- `node src/control.mjs account` still returns Plus plan usage on this machine

## Manual smoke test

1. Open Codex Router Control Center on a signed-in ChatGPT account.
2. If the usage probe is interrupted, the dashboard should keep loading. Quota tiles may show Unavailable; there should be no Node stack under "Some router data could not load".
3. Refresh. When the probe completes, ChatGPT usage should populate as before.


slavakurilyak:
macOS `test` is red on a tray watchdog test this PR does not touch: `disarming after the process is gone is inert` (`ProviderCatalogTests.swift`). It expected `printf ok` to finish in 0.05s and got SIGTERM (status 15). `npm test` passed on all three OSes, and the first CI run on this branch was green. Happy to retrigger if that’s useful.

PracticalSwan:
Current `main` has moved past this branch with merged #648 (`account-usage-partial-answers`), so the conflict is real but the two fixes are complementary rather than competing.

I reconciled the behavior against current-main semantics locally:

- keep #648's timeout behavior: if one account read is silent, return the reply that arrived and synthesize only the missing optional half;
- add this PR's process-exit behavior: if the temporary app-server exits after at least one account reply, normalize/return the partial replies instead of rejecting;
- if it exits before any account reply, still reject;
- keep the defensive stdin write and `control account` unavailable payload from this PR.

Focused `test/codex-account-usage.test.mjs` with both #648 regressions plus the two exit regressions passes **14/14**; `scripts-check.mjs` passes; a real `node src/control.mjs account` invocation exits 0. No provider call or private transcript was involved.

So I would refresh this PR onto current `main` by preserving #648's `emptyResponse`/partial-timeout path and routing both the timeout and post-one-reply exit paths through the same normalized partial-reply helper.

duolahypercho:
Thanks @slavakurilyak, and @PracticalSwan for the reconciliation notes. I pushed a merge of current `main` onto this branch (a merge, not a rebase, so your commit is untouched). What changed:

**`src/codex-account-usage.mjs`**
- Combined with #648 as suggested: the timeout, both-answered, and exit-after-one-reply paths all go through #648's `emptyResponse`. An app-server that exits after one reply now returns the same shape as one that stays silent.
- **Settles on `close` instead of `exit`.** Node can emit `exit` while the last reply is still in the stdout pipe, and settling there closes the reader over an answer the app-server had already written. The new test `replies still in the pipe when the app-server exits are not discarded` fails on d014ac5e with `exited before replying (1)` and passes here. The fake app-server now emits `close` after stdout drains, as a real child process does.
- Your guarded stdin write and both exit regressions are kept.

**`src/control.mjs`: the catch is dropped.** The macOS tray (`refreshAccountUsage`) and the panel (`refreshPanel`, `refreshIslandUsage`) both keep the last good usage when `control account` fails, and replace it when it succeeds. Turning a failure into a blank success would wipe real numbers on one bad 30-second poll and clear the tray's error. Nothing reads `unavailable` / `unavailableReason` today.

The case in your repro, where the app-server exits after one reply, now resolves at the source, so no banner. The remaining case, exiting before *any* reply, is a real failure and still reaches the Control Center. If we want it off the global banner, that belongs in `App.tsx` (the account-usage `settleRead` currently feeds `readErrors`), probably shown on the Usage page. That would be a separate follow-up.

Verified locally: `codex-account-usage`, `chatgpt-account-control`, `chatgpt-account-pool`, `discovery-mode` 41/41 pass, and `node scripts-check.mjs` passes.

PR #637, opened 2026-09-06
fix: preserve native subagent relay auth and quota failures

PracticalSwan:
## Problem

Routed subagent turns can contain native Codex `encrypted_content`. The router correctly opens that payload through the signed-in native backend before sending plaintext to the external provider, but native relay failures were losing important HTTP semantics:

- HTTP 429 was rewritten to local 502, causing retry storms against an exhausted native relay.
- HTTP 401 was also rewritten to 502, preventing Codex's built-in ChatGPT authentication recovery from seeing a recoverable unauthorized response and retrying after refresh.

A real completed Codex Desktop session reproduced the 429 path at scale: **63** native collaboration relay quota failures were surfaced locally as 502 before this fix.

## Fix

- Preserve native collaboration relay `429` as HTTP 429 with `ERR_NATIVE_AGENT_RELAY_RATE_LIMITED`.
- Remember only the exact account-and-ciphertext 429 failure key for a short 60-second anti-storm window.
- Keep that cache bounded to 128 entries and retain only the already-hashed key plus expiry timestamp; no upstream body, credential, or plaintext is stored.
- Keep failures isolated by native account and delegated payload.
- Retry native normally after the short 429 backoff expires.
- Preserve native relay `401` as HTTP 401 with `ERR_NATIVE_AGENT_RELAY_UNAUTHORIZED`, while replacing the upstream response body with a sanitized local error.
- Do **not** cache 401. A retry after Codex refreshes credentials reaches native again.
- Leave all other native relay failures fail-closed as 502, so unreadable ciphertext is still never sent to the external provider.
- Allow the 429 anti-storm window to be tuned with `CODEX_ROUTER_AGENT_RELAY_FAILURE_BACKOFF_MS` / `MODEL_ROUTER_AGENT_RELAY_FAILURE_BACKOFF_MS`.

The 401 behavior is intentionally aligned with current Codex client behavior: its OpenAI Responses client recognizes recoverable unauthorized transport errors, runs authentication recovery, and retries the request.

## Verification

After normally merging current upstream `main` (`0eaed3e`, including merged #650, #658, #659, and #660) into this branch without rebasing or force-pushing:

- Focused 401/429 relay regressions remain included.
- Full `test/routing.test.mjs`: **118/118 pass**.
- `node scripts-check.mjs`: pass (`v2-agent applications valid (6)`, syntax checks passed).
- staged `git diff --check`: pass.
- Resulting PR diff against current `main` remains isolated to the original four files.
[...]

PR #663, opened 2026-09-08
Fix native collaboration relay quota errors and retry storms

B1gM8c:
## Problem

A routed private-model turn can contain a native Codex collaboration payload encrypted by the signed-in ChatGPT backend. The router must relay that opaque payload to the native `/responses` endpoint before the private provider can continue.

When the signed-in Codex account is out of usage or rate limited, that relay returns HTTP 429. The router currently rewrites every non-success relay response to a generic 502. Codex therefore reports `The local router could not complete the request`, incorrectly suggesting a crashed local service or failed private provider.

Because successful payloads are cached but relay failures are not, rapid retries and multiple tasks can repeatedly hit the already-limited native account while the private provider itself remains healthy.

## Root cause

`relayEncryptedAgentPayloadOnce()` discarded the native response status and always threw status 502. Relay coalescing covered one identical ciphertext only; there was no short-lived account-scoped backoff after a native 429.

## Fix

- Preserve native relay 429 as HTTP 429 with stable code `native_collaboration_relay_rate_limited`.
- Return an actionable explanation instead of blaming the local router.
- Propagate a safe `Retry-After` value.
- Add bounded in-memory account-scoped backoff after a native relay 429: honor the upstream window with a five-minute local cap, default to 60 seconds without a usable window, isolate accounts, cap entries at 64, and expire automatically.
- Expose only aggregate backoff counters in `/health`; retain no account identity, authorization value, ciphertext, or plaintext there.
- Leave successful plaintext caching and ordinary private-provider routing unchanged.

## Verification

- `npm run check` passes.
- Focused native-relay tests: 4/4 pass.
- Complete `test/router-resilience.test.mjs`: 14/14 pass.
- Adversarial coverage proves the 429 is not mislabeled as 502, `Retry-After` reaches the caller, a different ciphertext for the same limited account does not reach native upstream, another account remains unaffected, and a headerless backoff expires before a later successful relay.

A local full-suite run reached 3,659 passing and 25 skipped tests. Four unrelated environment/baseline failures remained: one browser assertion expected English while the machine rendered Chinese; two doctor fixture failures reproduce unchanged on `origin/main`; and an installer test was intercepted by a machine-wide pre-commit hook and passes when that external hook is suppressed.

## Privacy and scope

The commit contains only `src/router.mjs` and `test/router-resilience.test.mjs`. It contains no local provider configuration, model overlay, API key, account identifier, usage log, cache, or generated artifact.

### DIFF
--- a/src/router.mjs
+++ b/src/router.mjs
@@ -359,8 +359,20 @@ const AGENT_PAYLOAD_CACHE_TTL_MS =
     : 15 * 60 * 1_000;
 const AGENT_PAYLOAD_CACHE_MAX_BYTES = 8 * 1024 * 1024;
 const AGENT_PAYLOAD_CACHE_MAX_ENTRIES = 256;
+const configuredAgentRelayBackoffMs = Number(
+  process.env.MODEL_ROUTER_AGENT_RELAY_BACKOFF_MS ||
+    process.env.CODEX_ROUTER_AGENT_RELAY_BACKOFF_MS ||
+    60_000,
+);
+const AGENT_RELAY_BACKOFF_MS =
+  Number.isFinite(configuredAgentRelayBackoffMs) && configuredAgentRelayBackoffMs > 0
+    ? Math.floor(configuredAgentRelayBackoffMs)
+    : 60_000;
+const AGENT_RELAY_BACKOFF_MAX_MS = 5 * 60_000;
+const AGENT_RELAY_BACKOFF_MAX_ACCOUNTS = 64;
 const agentPayloadCache = new Map();
 const agentPayloadCacheInFlight = new Map();
+const agentRelayBackoffs = new Map();
 let agentPayloadCacheBytes = 0;
 const agentPayloadCacheMetrics = {
   hits: 0,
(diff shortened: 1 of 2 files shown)


duolahypercho:
Thanks for the careful write-up and the adversarial tests, @B1gM8c. Closing this as superseded, not rejected. #637 landed on `main` about ten hours after this was opened and fixes the same bug:

- A native collaboration relay 429 is kept as HTTP 429 (`ERR_NATIVE_AGENT_RELAY_RATE_LIMITED`) instead of being rewritten to 502.
- Native attempts are suppressed for a 60-second window after a 429, bounded and isolated per account (`router preserves relay 429 and suppresses repeated native attempts per account` in `test/routing.test.mjs`).
- A relay 401 is also kept as 401 so Codex can refresh its session.

Rebasing this branch would add a second, parallel backoff next to #637's: a separate map, env var (`AGENT_RELAY_BACKOFF_MS` vs `AGENT_RELAY_FAILURE_BACKOFF_MS`) and error code. It also conflicts in `src/router.mjs`.

A few things here are still worth having. A small follow-up built on #637's `agentPayloadRelayFailures` / `nativeAgentRelayRateLimitError` would be welcome:

1. **Follow `Retry-After`** (capped). `main` always waits a fixed 60 seconds and doesn't pass the header back to the caller.
2. **Back off per account, not per payload.** #637 keys the window on account + ciphertext, so a different payload from the same limited account still reaches the native backend.
3. **A clear message.** The response body still says "The local router could not complete the request."

(The red `Unified native macOS app` job was a Control Center lifecycle timeout, unrelated to this diff.)

PR #664, opened 2026-09-08
Fix OpenAI continuation after external tool calls

webhype:
## What failed

A Codex task could use Kimi K3, then stop working when the user switched back to
an OpenAI model. In the reported case, the user switched because K3 credits had
run out. Forking the task did not help because the fork inherited the same
saved tool call.

The saved call had an optional `id` starting with `tool_`. OpenAI rejected it:

```text
Invalid 'input[78].id': 'tool_example_foreign'. Expected an ID that begins with 'fc'.
```

The identifier above is synthetic. This is an item-ID problem, not a tool-name
prefix problem. K3 becoming unavailable prompted the switch; it did not make
the ID incompatible. The offending call was already in the saved history.

## Why compaction was involved

Compaction is the step where Codex asks a model to shorten the conversation so
work can continue. In this incident, the model switch triggered that step.
Codex tried the previous Kimi model, then retried with the newly selected OpenAI
model. The logs recorded `CompHashChanged` and
`previous-model compaction failed; retried with current model`.

The shortening request still included the old tool call. OpenAI rejected its
`id` before it could return a shorter conversation, so the task never reached
the new message. Compaction exposed the bad ID; it did not create it. An
ordinary continuation can include the same history, which is why the repair
covers both.

## What this changes

In the existing `normalizeNativeInput()` function, omit a full
`function_call` item's optional `id` when it is a string that does not start
with `fc`. Apply this only when preparing history for OpenAI.

`id` identifies the conversation item. The separate `call_id` connects the
tool call to its result and must remain unchanged. OpenAI's input type makes
`id` optional and `call_id` required; the observed backend error supplies the
prefix requirement. See
[OpenAI's function-call input type](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_function_tool_call_param.py).

The patch preserves `call_id`, the result, tool names, arguments, other fields,
and IDs that already start with `fc`. It does not modify saved conversations
or apply this rule to external-provider requests. It needs no new dependency
or migration and can repair already-saved full calls on the next request.

## How to reproduce

1. Use an unpatched router at `0eaed3e17ceb2bbaa2c57a8808bb64eea74cf86a`.
2. Select `kimi-oauth/k3` in Codex and ask it to read a harmless repository file
   and summarize it. It must actually call a tool; text-only chat is not enough.
3. Retain a saved `function_call` whose string `id` starts with `tool_` or
   otherwise does not start with `fc`.
4. Switch that task to an OpenAI model and ask it to continue. In the reported
   incident this happened after K3 credits ran out; there is no need to
   deliberately exhaust credits to test the ID bug.
5. Check the failed OpenAI request for the ID-prefix error above. If the switch
   causes Codex to shorten the conversation first, that request can fail before
   the new message is answered. A generic high-demand warning alone does not
   establish this cause.

For a deterministic regression check on this PR branch, without provider calls
or credits:

```bash
node --test --test-name-pattern='native function-call IDs' test/routing.test.mjs
```

The test uses local substitute servers and fake credentials. Before the source
fix, all six OpenAI request cases failed with the expected 400; the external
case passed. After the fix, all 8 reported tests pass, including the parent.

To reproduce the pre-fix failure, keep the new test and run it with
`src/router.mjs` from the baseline commit in a disposable checkout. The named
test does not exist at that baseline by itself. Do not change a live
installation to recreate the failure.

## Verification

- Covers ordinary continuation, `/responses/compact`, and `/responses` with a
  final `compaction_trigger`, each with caller-supplied and router-supplied
  sessions.
- Compares complete forwarded histories, checks repeated history, and verifies
  that the external Kimi route keeps its original item ID.
- Fresh focused run: **8 passed, 0 failed**.
- Full Node suite on the same source and test patch: **3,671 passed, 0 failed,
  26 skipped** out of 3,697 tests. Skips include platform-specific checks, an
  optional YAML parser check, and the opt-in real LiteLLM adapter test.
- `node scripts-check.mjs` and `git diff --check` passed.
- The affected user updated the router, manually restarted it, and confirmed
  that the previously broken task worked again.

## Recovery screenshot and repair notes

The screenshot supplied by the user shows an earlier generic high-demand
warning followed by a successful answer. It illustrates the reported recovery;
the assistant's self-description is not proof of the exact selected model.

Earlier warning followed by a successful answer

The [repair plan and verification record](https://github.com/webhype/codex-router/blob/66cef219b76337fb496f22d122c4779adad7830b/docs/OPENAI-TOOL-CALL-ID-REPAIR-PLAN.md) explain the field rule,
test coverage, alternatives, and limits. The implementation and these optional
explanatory materials are separate commits so the fix can be taken on its own.

This fixes the rejected optional tool-call ID. It does not fix Kimi quota
limits, provider outages, authentication errors, or other damaged history.


PracticalSwan:
Fresh prerequisite evidence from a current Windows Codex session, without a provider probe:

- Codex Desktop `26.903.8094.0` / CLI `0.153.4`
- codex-router `0.5.1`
- clean router reference `7e0a2b72ebb85ed497d9df8529ac251262261d51`
- one completed routed subagent used `zai-coding/glm-5.3`

A read-only scan of that child rollout found **20 saved `function_call` items, all 20 with optional item IDs using the `call_...` prefix rather than `fc...`**. The parent rollout has no foreign function-call item IDs.

The child completed normally and has not been replayed to native OpenAI, so I do **not** have a fresh OpenAI 400 to report from this session. This establishes the concrete saved-history prerequisite only: if that child thread is later resumed/switched to native OpenAI, its stored item IDs are exactly the shape this PR is intended to normalize.

Current `main`'s `normalizeNativeInput()` still handles reasoning/item-reference/compaction normalization but does not omit non-`fc` function-call item IDs. No live patch was deployed in this maintenance pass because the affected child is complete and no actual OpenAI continuation failure has occurred.

This may be useful justification for refreshing the PR onto current `main` rather than relying only on the older incident. No exact IDs, prompt/tool content, thread IDs, local paths, account data, or credentials are included.

duolahypercho:
The fix is right. The code applies cleanly to current `main` and the 8 new tests pass there. Three things before merge:

1. **Cover `custom_tool_call` too.** Bridged custom calls (e.g. `apply_patch`) keep the provider's `id` via `...rest` in `src/namespace-relay.mjs` (~L2152), so they will likely hit the same 400 on OpenAI replay. Please strip non-native IDs there as well, or at least add a test.
2. **Drop the docs commit.** The 1 MB PNG and the plan doc shouldn't live in the repo; the PR description already covers them.
3. **Rebase onto `main`.** Only `CHANGELOG.md` conflicts.

duolahypercho:
Picked this up in #697, which keeps your commit as-is and extends the fix to custom tool calls (`ctc`) and messages (`msg`). The docs/PNG commit is left out. Thanks for the detailed report and tests!

PR #670, opened 2026-09-09
fix: keep active Grok OAuth streams alive across long reasoning pauses

Leshabeats:
Grok OAuth can remain active while pausing longer than the 30-second empty-completion prelude. This gives an already-started Grok stream a separate, bounded ten-minute idle window (`CODEX_ROUTER_GROK_STREAM_STALL_MS`). Headers-only attempts and other providers retain their existing limits; cancellation and the prohibition on replaying a stream after bytes reach the client remain intact.

Stacked on #652. The timeout-only change is commit `6a1b5f231588a03c3476d4999365b208e7e05c10`; the preceding three commits refresh #652 onto current main and should land through that PR.

Validation: the new Grok idle regression fails on the parent without this fix; 179 adjacent tests and all 36 empty-completion router tests pass, including an actual 35-second idle, a 30-second non-Grok control, cancellation, terminal errors and no replay. `npm run check` passes. The full local suite reported 3720 passed, 26 skipped and two doctor tests timing out against the host's real service; both pass with the repository's existing service-platform isolation setting. Current-head Linux/macOS/Windows CI passes. No live model calls are part of CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@
 - **Grok gateway stream errors reach Codex as terminal failures.** Untyped
   gateway errors are normalized without exposing upstream diagnostics or
   appending empty message closes, and request activity records the failure.
+- **Grok OAuth streams use a ten-minute idle bound after the prologue is released.**
+  A reasoning pause longer than the 30-second empty-completion prelude is not an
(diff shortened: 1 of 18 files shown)


duolahypercho:

PR #671, opened 2026-09-09
feat: correlate Grok requests and export private run metrics

Leshabeats:
Usage records can now be joined to `/activity` by the same instance-scoped `requestId`, including success, failure and cancellation. Grok OAuth 4.6 records separate UTF-8 JSON byte sizes for ingress instructions, tools and input history. Existing usage records remain readable; provider reasoning counters preserve the distinction between missing and explicit zero.

The local report exports request/token/timing totals, observed tool and patch failures, the first test-command attempt and terminal outcome. It projects metrics only, never prompts, tool arguments/results, credentials or reasoning text. It retains multiple charged attempts per request, reports counter coverage and does not treat byte sizes or text TTFT as decoding throughput. Filename reads, quoted examples and comments do not count as test attempts. CLI shell failures remain visible even when the tool reports completed; repeated updates are counted once.

Stacked on #652: its first three commits supply the existing `/activity` observer. This changes observation, not routing, billing, retries or tool availability. Live comparative runs remain outside ordinary CI.

Validation: request correlation fails on the parent without the wiring. Focused tests cover exact correlation, old records, absent/zero reasoning counters, terminal paths, other-model controls, UTF-8 sizes and content-leakage canaries. All 12 report tests and `npm run check` pass at `cf7e9d6420833778aeec681594f6441d5e35044d`. The two report corrections each fail their regression before the fix. Full local and current-head hosted CI results are recorded separately; the local full suite had two host-service doctor timeouts, which passed with the existing service-platform isolation setting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+- **Routed workers expose read-only request progress.** `bin/control activity
+  [thread-id]` reports active requests and bounded recent outcomes behind the
+  caller capability, including observed stream events and cancellation causes.
+  Quiet polling never ends a request, and the shipped agent guidance separates
+  polling intervals from worker deadlines.
+- **Grok repairs require a successful response terminal.** Failed, incomplete,
+  and truncated streams return a terminal error without releasing withheld
+  client actions or private final answers. Reasoning remains live.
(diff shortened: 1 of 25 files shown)


duolahypercho:

PR #672, opened 2026-09-09
feat: add scoped Grok apply_patch examples with full protocol verification

Leshabeats:
Only `grok-oauth/grok-4.6` receives two short create/update examples in the native custom `apply_patch` description before LiteLLM translation. The original description and grammar format remain intact. Other routes, same-named ordinary functions, tool calls/results and patch application rules retain their existing behavior. No new write tools, automatic patch repairs or additional model requests are introduced.

LiteLLM already preserves the grammar. This is an experimental description change, not a repair for lost grammar. The live pilot did not qualify the package for expansion: malformed delimiters still occurred in an uncontaminated run, and some other runs lacked read isolation. No speedup or quality improvement is claimed; this PR remains a draft.

Deterministic validation covers route/collision controls, immutable grammar, streamed Unicode/quotes/newlines, malformed content, and call/result identity. A standalone verifier exercises Router → hash-locked LiteLLM proxy → Grok forwarder → local mock xAI → restored Codex response. Ordinary Node tests need no personal Python installation; live model calls are excluded from CI. The stock Codex patch tool passes create/update/delete and rejects invalid format/unmatched context without mutation.

`npm run check`, focused Node suites and the real proxy roundtrip pass. Full Linux/macOS/Windows tests and the three-OS protocol matrix pass on the same head `75914f1a05960ae4c9e2010fec693ae724aec73c` in the [independent validation run](https://github.com/Leshabeats/codex-router/pull/2/checks). The original upstream Windows job is still in progress; that run has not been discarded or represented as passed.

### DIFF
--- a/.github/workflows/grok-apply-patch-guidance.yml
+++ b/.github/workflows/grok-apply-patch-guidance.yml
@@ -1,0 +1,116 @@
+# Ordinary `npm test` in ci.yml is Node-only. LiteLLM custom-tool translation
+# for grok-oauth/grok-4.6 is proven here against the hash-locked Python tree
+# and a mock xAI Responses server — never against a machine-local install and
+# never against live credentials.
+name: Grok apply_patch protocol
+
+"on":
+  push:
+    branches: [main]
+    paths:
+      - "requirements/**"
+      - "bin/install"
+      - "install.ps1"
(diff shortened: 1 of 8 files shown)


duolahypercho:

PR #673, opened 2026-09-09
fix: preserve native models under signed routing

B1gM8c:
## Summary

Fix signed ChatGPT coexistence after Codex tightened custom-provider model validation, while preventing native subscription models from being captured by external aliases or native redirect rules.

## Problem

Recent Codex builds reject an explicit routed model such as `private/gpt-6-astra` when signed routing leaves the root provider as `openai`:

> The 'private/gpt-6-astra' model is not supported when using Codex with a ChatGPT account.

Switching the root provider to the router satisfies custom-provider validation, but exposed three compatibility bugs:

1. Catalog construction treated every `model_provider = "codex-router"` configuration as login-free, even when the provider still required the operator's ChatGPT authentication.
2. Native aliases and `native-redirect` could capture an unprefixed official GPT slug and route it to an external provider (for example Grok), independently of model failover.
3. A thread switched from a routed model back to an official ChatGPT model could replay an unstored private-provider `rs_` reasoning reference and receive a persistence-related 404.

## Fix

- Add a versioned `provider-switch` signed-routing state that:
  - selects the router as the root provider required by current Codex validation;
  - keeps `requires_openai_auth = true` so ChatGPT account authentication remains active;
  - snapshots and exactly restores the previous root provider and provider table;
  - fails closed if either root-provider or table ownership drifts.
- Distinguish signed routing from login-free routing when building the catalog.
- Resolve exact routed slugs first, then force unprefixed official GPT slugs through native ChatGPT before consulting aliases or native redirect.
- Return both OpenAI-compatible `data` and Codex-native `models` arrays from `/models` for custom-provider discovery.
- Treat official passthrough as stateless for mixed-thread reasoning provenance, dropping only unstored `rs_` references outside the existing V1 compact contract.

Explicit external selections such as `private/gpt-6-astra` and `grok-oauth/grok-4.6` continue to route normally. Model failover behavior is unchanged.

## Validation

- `npm run check` — passed.
- Focused catalog/config/alias tests — 130/130 passed.
- Focused routing regressions — 5/5 passed, covering dual-shape discovery, official passthrough before aliases/redirect, explicit external routes, and mixed-thread reasoning cleanup.
- `npm test` — 3,670 passed, 25 skipped, 4 failed. The exact same four failures reproduce on an unmodified `origin/main` worktree in this environment:
  - one locale-dependent panel text assertion;
  - two doctor JSON-output truncation assertions;
  - one install rollback test affected by the user's global ECC pre-commit hook in the temporary fixture repository.
- Live smoke observation with failover disabled showed official `gpt-5.6-sol` requests using `provider=openai`, private models using `provider=private`, and router health reporting no degraded routes.

## Scope and security

This PR contains only generic source and regression-test changes. It does not include local provider configuration, API keys, tunnel identifiers, private endpoints, generated catalogs, or runtime state.

### DIFF
--- a/src/catalog.mjs
+++ b/src/catalog.mjs
@@ -22,6 +22,7 @@ import {
   NATIVE_ALIAS_PATH,
   NATIVE_CATALOG_PATH,
   PORTS,
+  SIGNED_PROVIDER_MODE_PATH,
 } from "./paths.mjs";
 import {
   codexAuthStatus,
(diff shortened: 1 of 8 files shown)


duolahypercho:
Thanks @B1gM8c. The tests are real: the three new routing tests fail on current `main` and pass here, and CI is green. Before this can merge, a few behaviour changes need a decision or a fix. I checked each one at head `610831f4`:

1. **This changes the rule in AGENTS.md, but AGENTS.md isn't updated.** "Codex safety boundaries" says the config manager may change the root `model_provider` only when the user turns on the tray's login-free mode. The new version-4 signed state writes `model_provider = "codex-router"` whenever signed routing is on (`src/config-manager.mjs`, `managedSignedProviderSwitchContents`). If that's the intended new rule, the PR needs to update AGENTS.md and explain why.

2. **Existing signed users get converted without doing anything, and a rollback then breaks.**
   - `signed-enable` now upgrades any current `mode: "root-openai"` state to version 4.
   - `src/refresh-catalog.mjs` runs `config-manager enable` and then `signed-enable`, so an ordinary update or catalog refresh flips the user's root provider.
   - Current `main`'s `readSignedProviderModeState` accepts only versions 1–3. After a rollback it throws `Invalid signed router provider state`, so `enable`/`disable`/`signed-disable` keep failing until someone edits the state file by hand.
   - Please either convert only when the user toggles signed routing, or keep the new state readable by the previous release. Also add a test for the upgrade path.

3. **Native-redirect is silently disabled for GPT slugs under any signed state.**
   - `officialModelsMustPassthrough()` only checks that `signed-provider-mode.json` exists, so it applies to version 1–3 states too, not just the new mode.
   - `resolveRegisteredRoute` then skips both aliases and `native-redirect` for official native slugs.
   - A signed-in user with exhausted OpenAI quota who ran `control native-redirect set <slug>` would have Codex's background memory/review turns sent back to ChatGPT, where they fail. `native-redirect set` still reports success.
   - If that's intended, it needs a warning where the redirect is set. Otherwise, limit it to the new mode.

4. **This contradicts `docs/SUBAGENT-CERTIFICATION.md`.** That doc says signed routing is not a way around "model is not supported when using Codex with a ChatGPT account", because its provider block has `requires_openai_auth = true` and so uses the same account. This PR's premise is that switching the root provider does get past that error. If your live test shows the doc is wrong, please update it and describe the evidence (Codex version, how the account was signed in, and exactly which request got through).

Two smaller points:
- Control Center labels a session's model with its `model_provider` (`apps/control-center/electron/ipc.mjs`), so native GPT sessions under the new mode will show `codex-router`.
- On signed installs, `isOfficialNativeSlug` reads and parses `native-models.json` synchronously on every native `/responses` turn.

The `rs_` cleanup, `loginFreeFromSignals`, and the `/models` shape look right.

B1gM8c:
Superseded by #699, rebuilt from current `main` to address this review.

The new PR:

- documents and constrains the explicit root-provider switch;
- reuses rollback-readable v1 state and includes a cross-version downgrade check;
- prevents ordinary update/catalog refresh from silently changing signed-routing mode;
- preserves native redirect as an independent control and adds operator warnings;
- corrects the subagent certification claims with exact, scoped evidence;
- fixes the Control Center provider label and avoids new hot-path state/catalog reads; and
- keeps the accepted `/models` and request-local `rs_` cleanup fixes.

No local configuration, credentials, endpoints, generated catalogs, or runtime state are included.

duolahypercho:
Closing as superseded by #699. Thanks @B1gM8c.

Issue #565, opened 2026-09-02
[Bug]: Windows Control Center tray opens multiple PowerShell windows during status refreshes


abdullahsamirashour:
## What happened

On Windows, the current unified Electron Control Center/tray opens a burst of visible **Windows PowerShell** terminal windows during routine status refreshes. In my setup this is especially noticeable every time I send a message in Codex: several PowerShell windows appear at once, then close one by one.

The router itself remains healthy; this appears isolated to the Control Center/tray child-process path.

Tested on a clean upstream checkout at:

`44029b8729b8166f8a662f5b62883c192bbeb9df`

Working tree was clean and `origin` was `https://github.com/duolahypercho/codex-router.git`.

## Process trace

A `Win32_ProcessStartTrace` capture showed the visible PowerShell processes are direct children of Control Center command invocations launched by the tray process.

Representative chain:

```text
Codex Router.exe --tray-only
  -> Codex Router.exe <checkout>\src\control.mjs --json
    -> powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -Command "& { & '<checkout>\bin\codex-router.ps1' service status --json }"
      -> node.exe <checkout>\src\service.mjs status --json
```

The same burst included commands for:

```text
codex-router.ps1 service status --json
codex-router.ps1 account status --json
codex-router.ps1 providers status --json
codex-router.ps1 presence status --json
codex-router.ps1 provider-usage --json
```

Eight `powershell.exe` processes were captured in one short burst. The parent/grandparent chain consistently pointed back to:

```text
Codex Router.exe --tray-only
```

No credentials were present in the captured command lines.

## Reproduction

1. Windows 10/11.
2. Use a clean current checkout with the Windows Control Center/tray installed and running.
3. Keep the router backend running normally.
4. Use Codex normally; in my case, sending a message reliably triggers the Control Center status/activity refresh.
5. Multiple visible Windows PowerShell / Windows Terminal windows open and then close.

## Isolation / workaround

I disabled and stopped **only** the `Codex Router Tray` scheduled task:

```text
schtasks.exe /Change /TN "\Codex Router Tray" /DISABLE
schtasks.exe /End /TN "\Codex Router Tray"
```

After that:

- no `Codex Router.exe --tray-only` process remained;
- the main router stayed healthy on `127.0.0.1:4202` (`TcpTestSucceeded: True`);
- Codex/Sol continued working normally;
- sending messages in Codex produced **no PowerShell popups**;
[...]

PR #674, opened 2026-09-09
fix: hide Windows Terminal windows on Control Center refresh

### LINKED ISSUE #581: Windows: router never comes back after power event / unlogged exit (LogonTrigger task + RestartOnFailure did not relaunch)

jaakjuu1:
## Problem

#565 / #571 added `windowsHide` on background PowerShell helpers. That is not enough when **Windows Terminal is the default terminal**: `powershell.exe`, `cmd.exe`, and `node.exe` are still console-subsystem binaries, so a GUI parent (Control Center / tray / `wscript`) still gets a visible window.

On a Windows 11 install with WT as default this showed up as:

- an empty PowerShell/Terminal window on every Control Center **Refresh**
- flashes from Task Scheduler (including the one-minute heartbeat while the task was already Running, LastResult `0x800710E0`)
- a persistent console if the scheduled task reached `node.exe` / LiteLLM directly (`DETACHED_PROCESS` made that worse)

`CREATE_NO_WINDOW` via WMI `Win32_ProcessStartup` is rejected (Create status 21). `SW_HIDE` on `WScript.Shell.Run(cmd.exe, 0, True)` still lets WT attach.

## Approach

Background work has to start from a **GUI-subsystem** host. Console children are not hidden reliably once WT owns the default terminal.

| Path | Change |
| --- | --- |
| Control Center Refresh | Spawn `Codex Router.exe` (`process.execPath`) with `ELECTRON_RUN_AS_NODE=1` instead of wrapping Node in `powershell.exe`. Abort still uses `taskkill /T`. |
| `control.mjs` | Skip the nested Job Object wrap when already running as that Electron Node host (`stdio: "inherit"` was forcing `windowsHide: false` and a visible powershell Job owner). |
| Presence | `targetIsActive` reads `SERVICE_PROCESS_STATE_PATH` and `process.kill(pid, 0)` instead of `service.mjs status` → `Get-ScheduledTask` PowerShell. |
| `file-security.mjs` | ACL PowerShell is launched through `pythonw.exe` + `src/windows-job-host.pyw` (`CREATE_NO_WINDOW`). Under `ELECTRON_RUN_AS_NODE` the ACL pass is skipped so Refresh cannot open a console. |
| `process-tree.mjs` | Job invocation prefers `pythonw` when `.venv\Scripts\pythonw.exe` is present; `effectiveWindowsHide` stays true on win32 when stdout is not a TTY. |
| Scheduled task | Generated `winexe` (`csc /target:winexe`, pythonw fallback) starts `node src/start.mjs` with `CreateNoWindow`. VBS no longer runs `cmd.exe`. Heartbeat TimeTrigger removed; logon + `RestartOnFailure` remain. |
| `start.mjs` | `windowsHide: true` on console-subsystem children (node / LiteLLM). |

`src/windows-job-host.pyw` copies child stdout/stderr onto fd 1/2 so a GUI parent can still read JSON. Wrapping Refresh in pythonw alone dropped that pipe and produced `router-control:getPresence: Router returned invalid JSON`.

## Tradeoffs (please read)

1. **Issue #581:** the minute heartbeat was the supervisor for unlogged exits / power events because `RestartOnFailure` does not relaunch after a *started* action exits. Those IgnoreNew ticks still flashed an empty Terminal window. This PR drops the heartbeat so the desktop stays quiet. A started action that later exits will not come back until the next logon or an explicit start — same gap #581 described.
2. **ACL skip in Control Center:** private-file hardening is skipped when `ELECTRON_RUN_AS_NODE=1`. CLI/service paths still run the PowerShell ACL helper (via pythonw when available). Refresh therefore does not re-ACL files it writes.
3. **Presence:** `installed || loaded` is no longer inferred from Task Scheduler. A task that exists but whose pid file is missing/dead reports inactive until the process is up.

## Test

- `node --test test/windows-hidden-consoles.test.mjs test/service-render.test.mjs` — 12 pass, 16 skip (Windows live scheduler cases)
- named Job Object tests in `test/process-tree.test.mjs` and `test/control-center-electron.test.mjs` — 2 pass
- `node scripts-check.mjs` — syntax checks passed
- Full `test/process-tree.test.mjs` / `test/control-center-electron.test.mjs` suites were **not** run here; they hang in this Job Object environment
- Manual: Control Center Refresh on Windows 11 + Windows Terminal default no longer opens a terminal window; presence JSON still returns

Follow-up to #571 / #565. Does not close #581 — it documents accepting that gap to keep the desktop quiet.

### DIFF
--- a/apps/control-center/electron/command-runner.mjs
+++ b/apps/control-center/electron/command-runner.mjs
@@ -638,6 +638,21 @@ export function windowsJobProcessInvocation(
     windowsHide: Boolean(windowsHide),
     windowsVerbatimArguments: Boolean(windowsVerbatimArguments),
   }), "utf8").toString("base64");
+  const pythonw = path.join(sourceRoot, ".venv", "Scripts", "pythonw.exe");
+  // powershell.exe is a console binary. Spawning it from Electron opens a
+  // visible Windows Terminal window on Refresh even with windowsHide.
+  // pythonw.exe is GUI-subsystem and starts the Job Object runner hidden.
+  if (existsSync(pythonw)) {
+    return {
+      command: pythonw,
+      args: [
+        path.join(sourceRoot, "src", "windows-job-host.pyw"),
+        windowsPowerShell(environment),
+        path.join(sourceRoot, "src", "windows-process-tree.ps1"),
+        payload,
+      ],
+    };
+  }
   return {
     command: windowsPowerShell(environment),
     args: [
@@ -897,12 +912,12 @@ function runEntrypoint(entry, args = [], {
     // separately installed runtime. In CLI/tests process.execPath is already
     // Node; in the desktop host ELECTRON_RUN_AS_NODE switches that same signed
     // executable into its Node mode.
-    const invocation = process.platform === "win32"
-      ? windowsJobProcessInvocation(process.execPath, [entry, ...args], {
-          sourceRoot,
-          environment: childEnvironment,
-        })
-      : { command: process.execPath, args: [entry, ...args] };
+    // Windows Control Center is already a GUI-subsystem executable. Wrapping
+    // it in powershell.exe (or pythonw -> powershell) opened a Terminal
+    // window on Refresh and, with pythonw, dropped stdout so JSON reads
+    // failed. Spawn this same GUI binary in Node mode; abort still uses
+    // taskkill /T. The Job Object helper remains for other console hosts.
+    const invocation = { command: process.execPath, args: [entry, ...args] };
     let child;
     try {
       child = spawn(invocation.command, invocation.args, {
--- a/src/control.mjs
+++ b/src/control.mjs
@@ -25,6 +25,7 @@ import {
   PROVIDER_API_KEY_POOL_PATH,
   PROVIDER_CREDENTIAL_STORE_PATH,
   PROVIDER_SELECTION_PATH,
+  SERVICE_PROCESS_STATE_PATH,
 } from "./paths.mjs";
 // Same reasoning: presence is a property of the shared plane, not of a target,
 // so the overview can resolve it statically without perturbing those probes.
@@ -109,9 +110,13 @@ const maximumControlOperationMs = boundedAntigravityOperation
   : restartBearingOverlayOperation
     ? 1_310_000
     : 850_000;
-if (!selfReplacingControl && !boundedOperationChild(process.env, {
+if (
+  !selfReplacingControl
+  && process.env.ELECTRON_RUN_AS_NODE !== "1"
+  && !boundedOperationChild(process.env, {
   maximumMs: maximumControlOperationMs,
-})) {
+})
+) {
   // Every ordinary control invocation enters one separately terminable tree.
   // Its owner gets ten seconds beyond the cooperative child budget so it can
   // escalate a full process-group termination. Catalog desktop watchdogs keep
@@ -144,13 +149,16 @@ function targetIsActive(target) {
   if (target === "cursor") return existsSync(CURSOR_PUBLISHED);
   if (target === "claude") return existsSync(CLAUDE_PUBLISHED);
   if (target === "openclaw") return existsSync(OPENCLAW_PUBLISHED);
-  const result = spawnSync(process.execPath, [path.join(REPO_ROOT, "src", "service.mjs"), "status"], {
-    env: { ...process.env, MODEL_ROUTER_TARGET: target },
-    encoding: "utf8",
-  });
+  // Do not spawn service.mjs here. That path shells out to powershell.exe
+  // Get-ScheduledTask, and Windows Terminal shows that console on every
+  // Control Center Refresh. The service pid file is enough to know the
+  // shared router plane is up.
   try {
-    const status = JSON.parse(result.stdout);
-    return Boolean(status.installed || status.loaded);
+    const parsed = JSON.parse(readFileSync(SERVICE_PROCESS_STATE_PATH, "utf8"));
+    const pid = Number(parsed?.pid);
+    if (!Number.isSafeInteger(pid) || pid <= 0) return false;
+    process.kill(pid, 0);
+    return true;
   } catch {
     return false;
   }
--- a/src/file-security.mjs
+++ b/src/file-security.mjs
@@ -15,6 +15,26 @@ import path from "node:path";
 const WINDOWS_PRIVATE_ASYNC_TIMEOUT_MS = 30_000;
 const WINDOWS_PRIVATE_ASYNC_OUTPUT_LIMIT = 64 * 1024;
 
+function windowsPythonwHost() {
+  const sourceRoot = process.env.MODEL_ROUTER_SOURCE_ROOT;
+  if (!sourceRoot || process.platform !== "win32") return undefined;
+  const pythonw = path.join(sourceRoot, ".venv", "Scripts", "pythonw.exe");
+  const host = path.join(sourceRoot, "src", "windows-job-host.pyw");
+  if (!existsSync(pythonw) || !existsSync(host)) return undefined;
+  return { pythonw, host };
+}
+
+function execFileSyncHidden(command, args, options) {
+  if (process.env.ELECTRON_RUN_AS_NODE === "1") {
+    return Buffer.from("");
+  }
+  const hidden = windowsPythonwHost();
+  if (hidden) {
+    return execFileSync(hidden.pythonw, [hidden.host, command, ...args], options);
+  }
+  return execFileSync(command, args, options);
+}
+
 // Windows private-file hardening is one bounded PowerShell operation per
 // atomic write. Keeping the async path one-shot is deliberate: Windows
 // PowerShell does not provide a stable line-oriented stdin protocol when it is
@@ -167,19 +187,13 @@ function terminateWindowsChild(child) {
 function protectPrivateFilesWin32(paths) {
   const list = [...paths];
   try {
-    execFileSync(
+    execFileSyncHidden(
       "powershell.exe",
       powershellPrivateCommandArgs(),
       {
         env: windowsPowerShellEnvironment(list),
         stdio: ["ignore", "ignore", "pipe"],
         timeout: 15_000,
-        // Every private write reaches this helper, including the ones a
-        // Control Center status refresh performs. A console child of a GUI
-        // parent gets its own window unless this is set, which is how a
-        // routine refresh produced a burst of visible PowerShell windows
-        // (issue #565). The script is non-interactive and its stdio is
-        // already redirected, so nothing is hidden from the operator.
         windowsHide: true,
       },
     );
@@ -200,19 +214,29 @@ function protectPrivateFilesWin32(paths) {
 // credentials out of the helper environment and cap diagnostic output.
 function protectPrivateFilesWin32Async(paths) {
   const list = [...paths];
+  if (process.env.ELECTRON_RUN_AS_NODE === "1") {
+    return Promise.resolve(list);
+  }
   return new Promise((resolve, reject) => {
     let settled = false;
     let timer;
     let stderr = "";
-    const child = spawn(
-      "powershell.exe",
-      powershellPrivateArgs(),
-      {
-        env: windowsPowerShellEnvironment(list),
-        stdio: ["ignore", "ignore", "pipe"],
-        windowsHide: true,
-      },
-    );
+    const hidden = windowsPythonwHost();
+    const child = hidden
+      ? spawn(hidden.pythonw, [hidden.host, "powershell.exe", ...powershellPrivateArgs()], {
+          env: windowsPowerShellEnvironment(list),
+          stdio: ["ignore", "ignore", "pipe"],
+          windowsHide: true,
+        })
+      : spawn(
+          "powershell.exe",
+          powershellPrivateArgs(),
+          {
+            env: windowsPowerShellEnvironment(list),
+            stdio: ["ignore", "ignore", "pipe"],
+            windowsHide: true,
+          },
+        );
     const settle = (error) => {
       if (settled) return;
       settled = true;
@@ -361,7 +385,7 @@ export function ensureCheckoutReadable(checkoutPath) {
   ].join("\n");
   const encoded = Buffer.from(script, "utf16le").toString("base64");
   try {
-    execFileSync(
+    execFileSyncHidden(
       "powershell.exe",
       ["-NoLogo", "-NoProfile", "-NonInteractive", "-EncodedCommand", encoded],
       {
@@ -400,7 +424,7 @@ export function privateFileIsProtected(target) {
     "[Console]::Out.Write(($acl.AreAccessRulesProtected -and -not $hasInheritedRule -and $hasFullControl -and -not $hasForeignAllow).ToString())",
   ].join("; ");
   try {
-    return execFileSync(
+    return execFileSyncHidden(
       "powershell.exe",
       ["-NoLogo", "-NoProfile", "-NonInteractive", "-Command", script],
       {
@@ -408,8 +432,6 @@ export function privateFileIsProtected(target) {
         env: { ...process.env, CODEX_ROUTER_PRIVATE_FILE: target },
         stdio: ["ignore", "pipe", "ignore"],
         timeout: 15_000,
-        // Verification runs from the same GUI-parented paths as the write
-        // above; see issue #565.
         windowsHide: true,
       },
     ).trim().toLowerCase() === "true";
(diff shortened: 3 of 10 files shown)


duolahypercho:
Thanks for digging into this, the symptoms are real and the write-up made them reproducible in reasoning. I traced the mechanism against the libuv and conhost sources and it points to a smaller fix than the one here. Reference implementation on top of current `main`, which you can cherry-pick or use as a diff: `fix/windows-674-consolidation` (commit bb5f6554).

### Mechanism

- Windows Terminal as default terminal ignores `SW_HIDE` (microsoft/terminal#12570, #15311). conhost leaves a console windowless only when the process was created with `CREATE_NO_WINDOW` (`ConsoleConnectionDeservesVisibleWindow` in `src/host/srvinit.cpp`), and descendants that inherit that console stay windowless.
- libuv turns Node's `windowsHide` into `CREATE_NO_WINDOW` only when **no stdio slot is inherited** (`src/win/process.c`). With `stdio: "inherit"` it sets `SW_HIDE` alone, which is why `windowsHide: true` on inherited-stdio spawns changes nothing under Terminal.
- Under the Control Center, `control.mjs` runs inside a GUI-subsystem executable that has no console. Every console child it spawns without `CREATE_NO_WINDOW` allocates a fresh, visible Terminal window. The refresh window comes from `schtasks.exe` queries without `windowsHide` and from the Job Object self-wrap, which inherits stdio. The persistent window is `cmd.exe` started with window style 0 by the VBS.

### What I'd keep from this PR

- `windowsHide: true` on `schtasks` in `service-windows.mjs`.
- The `pythonw.exe` + generated `.pyw` launcher, but with the `.pyw` running `cmd.exe /D /C start-codex-router.cmd` with `CREATE_NO_WINDOW` (so the wrapper stays the single definition of environment, log, and exit code, and the wscript exit-code test keeps passing), and the direct `cmd.exe` form as the fallback when the venv is absent.

### What I'd drop, and why

- Heartbeat trigger removal: regresses #581 (#618, #630). A tick dropped by `IgnoreNew` starts no process, and a tick that relaunches goes through the windowless launcher, so it draws nothing once the flags are right.
- ACL skip under `ELECTRON_RUN_AS_NODE`: the helper already spawns with `windowsHide` and non-inherited stdio, so it is already windowless.
- Spawning `Codex Router.exe` directly and skipping the `control.mjs` self-wrap: loses kill-on-close containment. Instead, in `runProcessTree`, when `stdio === "inherit"` on win32 and no standard stream is a TTY, spawn the PowerShell owner with `["ignore", "pipe", "pipe"]`, write the child's output to `process.stdout`/`process.stderr`, and pass `windowsHide` through to the contained command.
- `targetIsActive` pid-file probe: changes presence from installed-or-loaded to running.
- The csc-compiled winexe and `windows-job-host.pyw`: an unsigned exe compiled at install time and a second process host, no longer needed with the above.
- `windowsHide` on `start.mjs` children with inherited stdio: a no-op per the libuv rule; they inherit the windowless console from the launcher.

The reference branch also extends `test/windows-hidden-consoles.test.mjs` to `schtasks.exe`, pins the relay shape in `test/process-tree.test.mjs`, restores the heartbeat tests, and adds a `render-python` check that parses the generated file. All suites pass on macOS; the live Windows 11 + Terminal check is the part only you can do. If you rebase onto `main` and push, I'll approve the CI runs.

duolahypercho (changes requested):
Requesting changes. The detailed reasoning is in my earlier comment; this is the checklist to get #674 mergeable. The reference branch `fix/windows-674-consolidation` (commit bb5f6554, PR #693, closed) already contains each item and passes the suites on macOS, so cherry-picking is fine.

**Must change**

1. `src/service-windows.mjs`: restore the minute heartbeat trigger and the stop/start enable logic (#581, #618, #630). Dropped `IgnoreNew` ticks start no process, so the heartbeat cannot draw a window once the launcher is windowless.
2. `src/service-windows.mjs`: drop the csc-compiled winexe. Keep `pythonw.exe` + generated `.pyw`, but make the `.pyw` run `cmd.exe /D /C start-codex-router.cmd` with `CREATE_NO_WINDOW` instead of re-implementing the environment and log redirect. The wrapper stays the single definition and `test/service-render.test.mjs`'s wscript exit-code case keeps passing. Keep the direct `cmd.exe` form as the VBS fallback when the venv is absent.
3. `src/file-security.mjs`: remove the `ELECTRON_RUN_AS_NODE` skip and the pythonw host. The existing calls already use `windowsHide: true` with non-inherited stdio, which is `CREATE_NO_WINDOW`; nothing there is visible.
4. `apps/control-center/electron/command-runner.mjs` and `src/control.mjs`: keep the Job Object owner and the self-wrap. Fix the visible owner in `src/process-tree.mjs` instead: when `stdio === "inherit"` on win32 and no standard stream is a TTY, spawn with `["ignore", "pipe", "pipe"]`, write the child's output to `process.stdout`/`process.stderr`, and pass `windowsHide` through to the payload. Node only sets `CREATE_NO_WINDOW` when no stdio is inherited; `!process.stdout.isTTY ? true : ...` alone leaves the PowerShell owner visible.
5. `src/control.mjs`: restore `targetIsActive` via `service.mjs status`. The pid-file probe changes presence from installed-or-loaded to running.
6. `src/start.mjs`: drop the `windowsHide` on inherited-stdio children; it is a no-op and the comment would mislead. Children inherit the windowless console from the launcher.
7. Remove `src/windows-job-host.pyw` and the pythonw branches in both `windowsJobProcessInvocation` copies.

**Keep**

- `windowsHide: true` on `schtasks` in `service-windows.mjs`.
- The VBS `FileExists` selection of pythonw and the `render-python` verb.

**Tests and docs**

- Restore the heartbeat tests; extend `test/windows-hidden-consoles.test.mjs` to `schtasks`; pin the relay shape in `test/process-tree.test.mjs`; add a CHANGELOG entry and a short AGENTS.md rule (the reference branch has wording for both).
- Rebase onto `main` (58 commits behind). I've approved the pending CI runs; each new push from the fork needs approval again, ping me.

The live Windows 11 + Windows Terminal check after reinstalling the service is the part only you can run; please note the result in the PR.

duolahypercho:
CI result for b395d72 (now approved and run): ubuntu, macOS, and both Electron jobs pass; the Windows job fails on this PR's own test, `the generated Windows launcher returns the wrapper's exit code to its script host` (test/service-render.test.mjs:462): expected exit code 42, actual 1, stdout/stderr empty. The runner has no `.venv`, so the VBS's `pythonw.exe` branch fails to start and there is no `cmd.exe` fallback. Item 2 of the review covers it: keep the direct `cmd.exe /D /C wrapper` form as the Else branch (the reference branch's VBS checks `FileExists` on both pythonw and the .pyw before choosing it), and have the .pyw run the wrapper rather than node so the exit code still reaches Task Scheduler.

duolahypercho:
Thanks for staying on this — I re-verified against current `main` and the two root causes you identified are both still there: `src/process-tree.mjs:752` leaves the PowerShell Job owner on inherited stdio with `windowsHide:false`, and `schtasks()` in `src/service-windows.mjs:137` still has no `windowsHide`. Your diagnosis is right; it's the delivery that needs another pass.

The branch merges cleanly onto current `main` (no conflicts), so a rebase is cheap. The blocker is CI-visible: `test (windows-latest)` fails on this PR's own `test/service-render.test.mjs:462` (expected exit 42, got 1). I reproduced the cause locally by rendering the launcher — the generated VBS picks `pythonw.exe` in the `Else` branch without a `FileExists` check on it and with no `cmd.exe` third branch, so on a host with no `.venv` the scheduled task exits 1 and the router never starts. That's worse than the visible window.

Two more things beyond the earlier review:

- `src/control.mjs:116` gates on `ELECTRON_RUN_AS_NODE` with no platform check, and `command-runner.mjs:863` sets that variable on macOS and Linux too — so the bounded-operation tree is dropped on all three platforms, not just Windows.
- `src/file-security.mjs:28` makes `privateFileIsProtected()` return `false` under the desktop app (it calls `.trim()` on a `Buffer`, which throws into the `catch`), and skips ACL hardening for files that `writeAtomic` itself notes may carry proxy credentials. Both ACL call sites already use non-inherited stdio with `windowsHide:true`, so they were already `CREATE_NO_WINDOW` — the skip costs hardening and buys nothing.

Also worth flagging: dropping the heartbeat trigger from `installTask()` regresses #581/#618/#630, and the suite doesn't catch it — the heartbeat cases in `test/windows-task-state.test.mjs` exercise `setTaskEnabled`, which this branch leaves intact, so they pass while the trigger is gone.

`fix/windows-674-consolidation` (bb5f6554) is still on origin if you want to cherry-pick; it's 57 commits behind now. Its `process-tree.mjs` change is the piece I'd most want carried over — it relays the child's output through pipes when the caller has no console, which is what actually gets `CREATE_NO_WINDOW` applied. Ping me and I'll approve the CI runs; the live Windows 11 + Terminal check after a service reinstall is the part only you can do.

duolahypercho:
Thanks for staying on this — the diagnosis behind it is right, and both root causes are still live on `main`. I read the whole diff. Flagging that the head here (`b395d725`, 9 Sept) predates the 10 Sept review, so the checklist items are still open and the branch is ~120 commits behind.

Three things I'd want addressed before anything else, because they are not about the console-window behaviour:

**1. Provider API keys end up in a file that never gets hardened.** `csharpHiddenLauncher()` embeds `serviceEnvironment()` as C# string literals and `compileHiddenExe()` runs `csc /target:winexe` to produce `start-codex-router-hidden.exe`. `providerApiKeyServiceEnvironment` returns live key *values*. The `.cs` goes through `writeAtomic`, which hardens it — its own comment says "Proxy URLs may contain credentials" — but the `.exe` is created by csc directly, so no `chmodSync(0o600)`, no `protectPrivateFile`, and it keeps the state directory's inherited ACLs. `strings` on it returns the keys. Compiling at install time is also a new AV/SmartScreen and supply-chain surface.

**2. `targetIsActive` returns false forever on macOS and Linux.** The new probe reads `SERVICE_PROCESS_STATE_PATH`, which `src/start.mjs:510` writes only under `process.platform === "win32"` — the comment there is explicit that other platforms don't need the marker. ENOENT → catch → `false`. Downstream, `applyProviderSelectionForTargets` pushes codex onto `skipped` and silently applies no provider selection at all. It also bypasses `readServiceProcessState`/`serviceProcessOwns`, which exist to validate `processIdentity`, `commandLine` and the `sourceRoot`/`stateDir` anchors — the comment there is "a hand-edited record for another checkout must never become a kill switch".

**3. ACL hardening is skipped under the desktop app.** `execFileSyncHidden` and `protectPrivateFilesWin32Async` short-circuit on `ELECTRON_RUN_AS_NODE === "1"`. `writePrivateFile`'s only error signal is that exec throwing, so it reports success and renames a file whose DACL was never replaced. Separately `privateFileIsProtected` calls `.trim()` on the returned `Buffer` → `TypeError` → `false`, which makes ChatGPT-login lease verification and caller-key rotation recovery throw unconditionally on Windows. `ELECTRON_RUN_AS_NODE` is also set with no platform guard, so the `control.mjs` self-wrap is dropped on macOS and Linux too.

**On the mechanism itself:** the `pythonw` shim does not apply `CREATE_NO_WINDOW` where it was missing. `effectiveWindowsHide` becomes `true` while `stdio` stays `"inherit"`, and per libuv's `src/win/process.c` that sets `SW_HIDE` alone — which is exactly what Windows Terminal ignores. The flag change is a no-op and the hiding is delegated entirely to running `pythonw.exe`, which then sets from Python the flag that should have been set from Node. The shim also `communicate()`s, so it buffers all output until exit and never forwards stdin — and `windowsJobProcessInvocation` selects it whenever `.venv/Scripts/pythonw.exe` exists, which is the normal installed layout, with no check on `stdio`. Your own comment in `command-runner.mjs` records that failure ("with pythonw, dropped stdout so JSON reads failed") and works around it only for the Electron entrypoint.

**The red Windows check is genuine, not one of this repo's known flakes.** Those are timeout-shaped; this is a deterministic exit-code comparison at `test/service-render.test.mjs:462` (expected 42, got 1) on a launcher this PR generates. The VBS `Else` branch runs `pythonw` with no `FileExists` guard and no `cmd.exe` fallback, so on a runner with no `.venv` `shell.Run` raises and the script quits 1. The new `assert.doesNotMatch(script, /cmd\.exe/)` forbids the fallback that would fix it. The same hits any user without a `.venv`: the scheduled task exits 1 at logon and the router never starts — worse than a visible window.

Also: two of the three modified tests were rewritten into `if (pythonw exists) … else <the old assertion>`. No CI runner has that path, so both would pass with the change reverted.

**The small version of this fix**, which I'd happily take: rebase, keep the one-line `schtasks` `windowsHide` and the `render-python` verb, and change `runProcessTree` to `stdio: ["ignore","pipe","pipe"]` + `windowsHide: true` so Node actually sets `CREATE_NO_WINDOW`. The winexe and pythonw hosts should come out entirely — that removes findings 1 and 3 along with them, and leaves the heartbeat trigger and `targetIsActive` untouched.

duolahypercho:
@jaakjuu1 heads-up that #744 from @brianbell-x has landed in review covering the
same Windows console root cause. It changes `runProcessTree` so an inherited
stdio that carries no TTY no longer forces `windowsHide` off, with a
`GetConsoleWindow()` probe asserting the child is windowless — so it touches
`src/process-tree.mjs` and `test/process-tree.test.mjs`, which this PR also
changes. This PR covers ground #744 does not: `src/control.mjs`,
`src/service-windows.mjs`, `src/start.mjs` and `src/windows-job-host.pyw`.

I would rather not land overlapping fixes to the same function. Could you and
@brianbell-x agree on the split — this rebasing onto that TTY check and keeping
the surfaces it does not reach, or #744 folding into this one? I will merge in
whichever order you settle on.

Note this PR's workflow runs have not been approved yet, so its CI is unproven;
say the word once the split is decided and I will approve them.

duolahypercho:
Thanks for the persistence on this one, and sorry it went stale.

The core problem — Windows Terminal ignoring `SW_HIDE`, leaving console windows visible — is now fixed on `main` by a narrower mechanism: `src/process-tree.mjs` relays through non-inherited pipes when `stdio` is `inherit` on win32 with no TTY, so `CREATE_NO_WINDOW` actually applies. That landed in `4cc5a088` and `2d589d72`, after this PR was opened, and it is the same fix asked for in item 4 of the review here.

The rest of this branch (the `pythonw` host and `.pyw` script, the csc-compiled winexe, dropping the minute heartbeat, and replacing `service.mjs status` with a raw pid check) goes further than that, and two of those changes would undo behaviour that is already on `main`: the heartbeat from `daf8fdc0`/`4f7c56be`, and `targetIsActive`'s installed-versus-running distinction. The ACL skip under Electron also drops private-file hardening on refresh.

Closing as superseded rather than asking for another rebase on a 58-commit gap. If you still see a visible console on current `main`, please open an issue with the exact repro (Windows build, terminal app, and which command spawns it) — that would be the most useful next step.

PR #678, opened 2026-09-10
feat(grok): consolidate runtime reliability, patch compatibility and native Fast

Leshabeats:
Grok OAuth turns can be cut off during long reasoning pauses, lose useful progress, or expose an unsuccessful repair as a completed turn. This consolidates the runtime fixes with private request diagnostics, scoped patch compatibility, and native Codex Fast support onto current upstream main.

## Behavior

- Give the Router stream guard a bounded ten-minute post-prologue stall allowance; preserve cancellation and the existing no-replay boundary after client output. Require successful response terminals before releasing held repairs and normalize gateway failures. Give streamed WebSocket errors a failure status understood by native Codex.
- Add authenticated activity snapshots and optional usage request IDs, provider-reported reasoning counters, bounded context byte sizes, and local metrics-only run reports. Reports exclude prompts, tool content, credentials, and reasoning text.
- Append native apply_patch examples only for grok-oauth/grok-4.6. Include explicitly opt-in structured-patch and negotiated PreToolUse-hook experiments with strict serialization, identity preservation, native errors, and cancellation tests. Codex retains execution, sandbox, permissions, and context matching. The hook is not installed or trusted automatically; a protected executable/module graph and independent client preflight are required.
- Advertise optional priority/Fast in the Grok 4.6 catalog and preserve it across the locked LiteLLM bridge for streaming HTTP, JSON HTTP, and WebSocket. Record requested versus provider-reported tier separately; missing data stays missing and aggregated retry billing is not assigned one attempt's tier. Default service tier and other model catalogs are unchanged.

## Integration and validation

Base: upstream b90aa60e257bbcc33855aad7d43954c1a09b1311. The merge preserves upstream chat-wire search-history replay and host-service isolation in doctor fixtures.

On consolidated head 9c3797a617163e91c7288dd7a256d08b42b2bfef:
- npm run check and git diff --check pass.
- Full locked-LiteLLM protocol checks pass locally for ordinary guidance, structured patches, native-hook header and endpoint modes, and all 21 Fast transport scenarios. These use loopback mocks without provider credentials.
- Full local verification: 3,829 passed, 26 skipped, 0 failed (3,855 total) with node --test --test-concurrency=4 test/*.test.mjs. The initial unrestricted run had one timing-sensitive failure in the unchanged Antigravity test; its focused suite passes 15/15. Both results are retained.
- Fresh upstream CI passes Linux/macOS suites, all three OS protocol jobs, native macOS, both Electron jobs, and security checks. The full Windows test job is still pending completion; no all-platform-green claim is made.
- A read-only Grok 4.6 reviewer ran with requested and provider-confirmed priority tier but exceeded its assignment budget without producing a final review. That run is not counted as an approval.

The earlier fork head 14be2df passed Linux/macOS/Windows CI and was installed locally; the user confirmed the native Desktop Fast control appears and works. That is previous-head/runtime evidence, not a substitute for checks on this consolidated head. No comparative throughput or Grok Build parity claim is made.

## Related work

Consolidates the changes proposed in #652, #670, #671, and #672, together with the downstream Desktop hook and Fast work. Those PRs remain open for maintainer choice; this PR does not merge or close them.

Live benchmarks, automatic hook installation, ACP integration, and changes to Codex sources are outside this PR. The hook experiment remains opt-in; see docs/GROK-STRUCTURED-PATCH-EXPERIMENT.md for its trust and acceptance limits.

## Known limitation retained for a separate follow-up

A live request on the previously installed head failed with UND_ERR_BODY_TIMEOUT after approximately 305 seconds. The Grok forwarder still uses the lower HTTP client body timeout (five minutes), which can expire before the Router's ten-minute stream guard. This consolidated PR preserves that lower transport behavior; it does not claim to eliminate all long-pause failures. Aligning the transport timeout and adding a full-path regression will be a separate change.


duolahypercho:
## Maintainer review: fixes pushed before merge

Thanks @Leshabeats for consolidating #652, #670, #671 and #672. Before merging, I split this PR into four areas: stream runtime, diagnostics, patch compatibility, and Fast. Each area had its own independent read-only review. I also checked every automated review finding left on the four source PRs against this branch. Every issue that still applied is fixed here in `14dd8211` and `3fb5c4a3`, with regression tests. I also merged current main twice: #681 needed a real conflict resolution, #685 merged cleanly.

### Stream runtime
- **The ten-minute stall guard now actually binds.** Before, the router→gateway and forwarder→xAI connection pools used Undici's 300 s body idle default, which is the `UND_ERR_BODY_TIMEOUT` at ~305 s reported earlier. The gateway's timeout also tied the guard at 600 s. All three hops now derive their limit from the stall bound (`src/grok-stream-timeouts.mjs`), for Grok OAuth only. The stall setting also rejects values too large for a Node timer.
- **Silent Grok streams get a heartbeat.** Codex abandons a stream after 300 s without a parsed SSE data event and then re-sends (re-bills) the whole turn; comments and pings don't reset that timer. While a Grok stream is silent, the router now relays an identity-only `response.in_progress` event (`src/responses-heartbeat.mjs`). It is sent only after `response.created`, only between complete events, and never after a terminal event, as the last pipeline stage. Grok OAuth only.
- **The first terminal event seals an attempt** (the P1 from #652). The forwarder stops reading there, so an upstream that keeps its socket open no longer delays the answer.
- **A failure stated before any content is released at once.** Previously it was held until the prelude limit and then reported a second time. The WebSocket drain after a terminal failure is now bounded.
- **An optional progress-only retry that does not complete keeps the first answer**, for both streaming and non-streaming requests. Strict post-tool repairs still fail closed.

### Diagnostics
- The routed transport retry and failover candidates now count as observed attempts, and a candidate waiting for headers is attributed to itself.
- When failover finds no taker, activity reports the original route.
- `response.completed` with a failed or incomplete status now settles as `failed` and records `terminalStatus`.
- Native and embeddings disconnects now settle as canceled. A canceled native stream used to be recorded as 200.
- Grok-only diagnostics are reset on failover, the run report totals billed counters, and `codex-router.ps1 activity` is added.

### Patch compatibility
- Structured patch paths are refused when a native header trim would change them, including U+0085.
- After #681 made custom tools namespace-aware, codecs attach only to a plain `apply_patch` declared in the request. A raw call under an undeclared namespace is treated as a codec bypass.
- CI now fails a hung job instead of holding the runner: the test job has `timeout-minutes`, and `npm test` has `--test-timeout`. This was the six-hour Windows job on #672.

### Fast
- A routed body keeps `service_tier` only when the route serving it advertises that tier. A Grok 4.6 priority request therefore never reaches a failover candidate or a compaction for another route, while curated models that advertise tiers keep them.
- A usage row covering two attempts no longer carries one attempt's tier.

### Verification
- **Locally**, on the final code before the last merge:
  - The unit, relay, routing and usage suites, including #681's DeepSeek and custom-tool tests, all pass. The forwarder file passes 71/71.
  - The LiteLLM protocol verifiers pass against the hash-locked LiteLLM 1.96.0 gateway: apply_patch guidance, the native hook transport (all three checks), and all 21 service-tier scenarios.
  - Local socket-based runs needed retries because this machine's VPN client was exhausting loopback ports, not because of the code.
- **CI on `daeb616d`:** `npm test` passes on Ubuntu, macOS and Windows. The Python protocol jobs pass on all three platforms, as do the native macOS app, the Homebrew formula, discovery tests, and GitGuardian.
  - **Electron control center (Windows):** failed on `trusted install provenance overrides contradictory package-manager environment` ("Router command timed out"). This test fails identically on main at `3e1fc299`, so it is a known Windows flake, not caused by this PR. It passed on rerun.
  - **Electron control center (Linux):** failed once on a renderer focus-timing wait (`the production renderer exposes model discovery and picker actions`, a mocked UI test that does not touch router code), then passed on rerun.
  - Both Electron jobs passed on rerun; all 12 checks are green on `daeb616d`.

### Left as known limitations
- Tokens from a rejected Grok terminal are logged by the forwarder but not metered by the router, because the error path through the gateway carries no usage.
- Hook mode's `allow` has not been verified against Codex approval under `untrusted` or `on-request`. The experiment docs say so, and hook mode stays opt-in and uninstalled.
- The gateway's Grok `stream_timeout` follows the environment of whichever process renders the LiteLLM config. Restart the service after changing the stall bound.

PR #679, opened 2026-09-10
feat(opencode): add DeepSeek V4.1 Flash support

mirucon:
OpenCode Go now documents DeepSeek V4.1 Flash as `deepseek-flash`, but the router does not list it and discovery blocks it as an unverified protocol. Add the separate `opencode-go/deepseek-flash` route using the documented Chat Completions endpoint, preserving V4 Flash.

The entry publishes text/image input and low/high/max effort. Its 1,000,000-token context compacts at 600,000 to reserve the documented 384,000-token output allowance. Live testing confirmed that forced tool choice returns HTTP 400 while `auto` calls the requested tool, so this model uses the existing model-scoped `auto-tool-choice` profile.

Validation:
- `npm run check` passed.
- Registry, curation, and model/picker suites: 89 passed; focused router-to-gateway forced-tool integration passed.
- Final discovery and desktop suites: 30 passed, 1 browser test skipped because Chromium is unavailable.
- Live OpenCode Go: low-effort tool call and image reading passed; high/max streaming markers passed.
- Full `npm test` completed: 3,561 passed, 108 skipped, 15 failed. The missing discovery snapshot was corrected and passed on rerun. Three dependency-related failures (Antigravity onboarding, desktop panel, exact-slug search transport) passed after installing the worktree dependencies. The remaining 11 failures are Windows symlink-permission errors (9) and legacy-migration path handling (2); the symlink restriction and both migration failures were reproduced on unchanged upstream. The full suite is therefore not green on this host.

Capability sources and probe evidence are recorded in `docs/research/opencode-deepseek-v4-1-flash.md`, including the official Go endpoint table and models.dev record. Native Codex collaboration certification is not claimed, and no v2 promotion is added. The installed router service was not changed or restarted.

### DIFF
--- a/config/opencode/go/deepseek-flash.json
+++ b/config/opencode/go/deepseek-flash.json
@@ -1,0 +1,38 @@
+{
+  "version": 1,
+  "models": [
+    {
+      "slug": "opencode-go/deepseek-flash",
+      "gatewayModel": "opencode-go-deepseek-flash",
+      "upstreamModel": "deepseek-flash",
+      "provider": "opencode-go",
+      "listed": true,
+      "displayName": "DeepSeek V4.1 Flash (opencode Go)",
+      "description": "DeepSeek V4.1 Flash through the opencode Go subscription, with image input and low/high/max reasoning.",
+      "priority": 37,
+      "defaultEffort": "high",
+      "reasoningLevels": [
+        {
+          "effort": "low",
+          "description": "Faster reasoning"
+        },
+        {
+          "effort": "high",
+          "description": "Deep reasoning"
+        },
+        {
+          "effort": "max",
+          "description": "Maximum reasoning depth"
+        }
+      ],
+      "contextWindow": 1000000,
+      "autoCompact": 600000,
+      "inputModalities": [
+        "text",
+        "image"
+      ],
+      "requestProfile": "auto-tool-choice",
+      "compHash": "opencode-go-deepseek-flash-v1"
+    }
+  ]
+}
--- a/docs/research/opencode-deepseek-v4-1-flash.md
+++ b/docs/research/opencode-deepseek-v4-1-flash.md
@@ -1,0 +1,13 @@
+# DeepSeek V4.1 Flash on OpenCode Go
+
+Verified 2026-09-10:
+
+- OpenCode's [Go endpoint table](https://opencode.ai/docs/go/#endpoints) names `deepseek-flash` as DeepSeek V4.1 Flash and routes it to `/zen/go/v1/chat/completions`. This is a separate ID from `deepseek-v4-flash`.
+- Read-only `node src/model-discovery.mjs opencode-go` returned the new ID.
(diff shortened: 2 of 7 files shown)


yishu-ziyu:
Tested this route end to end on macOS (Codex CLI 0.154.0, router 0.5.1 at `b90aa60` plus this PR). It fails one of the five live compatibility checks:

```
$ ./bin/test-model 'opencode-go/deepseek-flash' --live --yes
PASS basic response
PASS streaming
PASS tool calling
FAIL stateless tool result: opencode rejected the request for DeepSeek V4.1 Flash (opencode Go).
     (HTTP 400: ... [invalid_request_error] The `reasoning_content` in the thinking mode must be passed back to the API.)
PASS compaction
```

The same command against `opencode-go/deepseek-v4-flash` passes all five on this host, so it is specific to the new ID rather than to the provider or to the chat wire.

**Cause.** The check replays an assistant `function_call` whose history carries no `reasoning` item. V4.1 Flash is an interleaved-reasoning model: `opencode-go.models.deepseek-flash` in models.dev declares `"interleaved": { "field": "reasoning_content" }`, which the `deepseek-v4-flash` record does not. The upstream therefore refuses an assistant tool-call turn that omits it.

Bypassing the router, both wires reject the same request:

- `POST https://opencode.ai/zen/go/v1/chat/completions` -> 400 `The reasoning_content in the thinking mode must be passed back to the API.`
- `POST https://opencode.ai/zen/go/v1/responses` -> 400 `The reasoning_text in the thinking mode must be passed back to the API.`

Both need the `x-opencode-session` header, otherwise the edge answers `MissingSessionID` first. So this is the provider's contract, not a defect in the Responses-to-chat translation.

**Why it might still be worth handling.** `carryReasoningThroughInput()` in `src/router.mjs` moves reasoning items from history onto the following assistant turn, so an ordinary Codex turn -- which does store them -- is unaffected. What still fails is a tool-call turn whose history has no reasoning item for that call, which is exactly the shape this compatibility check constructs. #298 added the equivalent handling for V4 Flash and V4 Pro (an empty `reasoning_content` when the history has none); if V4.1 Flash is meant to survive that shape too, the same treatment would need to reach this route.

I did not re-test image input or the effort ladder here. The research doc in this PR covers both, and I am not claiming anything about ordinary multi-turn Codex behaviour -- I have not been able to confirm a real task yet.

duolahypercho:
Thanks @mirucon for this, and for the capability research: the 1M window, image input, low/high/max ladder, and the forced-`tool_choice` finding all carried over.

Shortly after this PR opened, OpenCode renamed the model to `deepseek-v4.1-flash`, and models.dev now marks `deepseek-flash` as deprecated (anomalyco/models.dev@2858118c, anomalyco/opencode#48363). #682 routes the renamed id as `opencode-go/deepseek-v4.1-flash` instead. That PR is now merged, so I'm closing this one as superseded.

The stateless tool-result failure @yishu-ziyu reported is DeepSeek's documented thinking-mode rule: when tools are present, every prior assistant turn must pass `reasoning_content` back. It is recorded as a known limitation in `docs/research/deepseek-v4-1-flash-2026-09-11.md`.

PR #680, opened 2026-09-10
feat(deepseek): add DeepSeek V4.1 Flash direct API support

MohamedMohsenX2:
## Summary

- add the live `deepseek/deepseek-flash` route through the official DeepSeek API
- preserve the existing `deepseek/deepseek-v4-flash` route and login-free fallback
- keep the new route conservative: text-only, non-thinking, and no native-v2 claim
- document the live API evidence and add registry/catalog/discovery regression coverage

This PR intentionally covers the official DeepSeek API only. OpenCode Go support
is already being handled separately in #679.

## Evidence

The configured DeepSeek API account's live `/models` response exposed
`deepseek-flash` on 2026-09-10. DeepSeek's public documentation currently
describes the same V4 Flash family under `deepseek-v4-flash`, so this adds the
live ID without renaming or replacing the existing documented route.

Official sources:

- https://api-docs.deepseek.com/api/list-models/
- https://api-docs.deepseek.com/
- https://api-docs.deepseek.com/quick_start/pricing/
- https://api-docs.deepseek.com/quick_start/agent_integrations/codex/
- https://api-docs.deepseek.com/updates/

The exact route was exercised through the router with the stored provider
credential and failover disabled. Basic response, streaming, forced tool call,
stateless tool-result replay, and compaction returned HTTP 200. Thinking mode's
stateless tool-result replay returned HTTP 400 because the upstream required
`reasoning_content` to be echoed, so the route uses the existing
`deepseek-nonthinking` profile and exposes only `minimal` effort.

## Validation

- `npm run check` passed.
- Focused registry, catalog, control, provider-selection, and discovery suite:
  **173 passed, 0 failed**.
- `git diff --check` passed.
- The full `npm test` wrapper was attempted but produced no output for five
  minutes on this host and was stopped; no full-suite pass is claimed.

No credentials, Codex settings, installed router service, or Codex process were
changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,7 @@ Linux installations support the Codex CLI.
 | Kimi K3 (China API) | `kimi-api-cn/kimi-k3` | Separately billed Moonshot **China** platform key |
 | DeepSeek V4 Flash (API) | `deepseek/deepseek-v4-flash` | DeepSeek API key |
 | DeepSeek V4 Pro (API) | `deepseek/deepseek-v4-pro` | DeepSeek API key |
+| DeepSeek V4.1 Flash (API) | `deepseek/deepseek-flash` | DeepSeek API key |
 | Grok 4.5 (OAuth) | `grok-oauth/grok-4.5` | Official Grok CLI OAuth session |
 | Grok 4.5 (API) | `grok-api/grok-4.5` | Separately billed xAI API key |
 | Claude Opus 4.8 (API) | `anthropic-api/claude-opus-4.8` | Separately billed Anthropic API key |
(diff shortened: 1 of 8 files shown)


duolahypercho:
Thanks @MohamedMohsenX2 for jumping on this so quickly.

DeepSeek has since published its V4.1 Flash docs. They confirm `deepseek-flash`, with a 1M context window, image input, and thinking on by default (low/high/max) (https://api-docs.deepseek.com/quick_start/pricing/, https://api-docs.deepseek.com/guides/thinking_mode). #682 uses the same `deepseek/deepseek-flash` slug with those documented values and the `deepseek-thinking` profile. That PR is now merged, so I'm closing this one as superseded.

The non-thinking choice here traces back to DeepSeek's rule that thinking-mode tool turns must pass `reasoning_content` back. That is recorded as a known limitation of the stateless compatibility check in `docs/research/deepseek-v4-1-flash-2026-09-11.md`.

PR #681, opened 2026-09-10
fix(deepseek): preserve native Responses streaming and reasoning replay

asifnawaz:
Direct `deepseek-flash` sessions currently pass through Responses-to-Chat translation, which can lose image references and mishandle reasoning and streaming. Base64 screenshots can also inflate the fallback token estimate and trigger premature compaction.

This PR contains two commits:

1. Bound actual image inputs when replacing explicit zero usage, using DeepSeek's documented 1,024-token maximum. Preserve positive usage, provider telemetry and original image payloads.
2. Route current direct `deepseek-flash` turns and compaction through the existing native `/responses` forwarder. Preserve typed reasoning, images and exact namespace/custom-tool identities; convert recovered Codex task messages into supported user messages. Legacy/reseller Chat routes retain their protocol and consume carried reasoning so it cannot also become a user message, covering the history-shape issue relevant to #653.

The change is model-scoped and reuses shared credentials, transport and tool adapters. Command Code's `/alpha/generate` fallback and model publication in #680 remain separate.

## Validation

- `npm run check` and the full test suite: **3,699 passed, 27 skipped, zero failures** on Linux / Node 24.19.0. Run with `MODEL_ROUTER_TEST_LITELLM_PYTHON=/path/to/pinned-python node --test --test-concurrency=4 test/*.test.mjs` to include the optional offline integration.
- Real pinned LiteLLM 1.96.0: GLM, legacy DeepSeek and Command Code pass positive and duplicate-history negative controls. All endpoints and credentials in this test are synthetic; an invalid explicit Python path fails before setup.
- Two bounded native API requests passed image-bearing tool-result and actual-response replay checks. Provider/client text, reasoning, stream identities and usage matched; 114 output tokens total.
- After local installation, a fresh Codex DeepSeek child received its task and executed a read-only repository command (exit 0). The recorded invocation and output were independently inspected.
- Regression tests retain task/tool/compaction delivery, malformed reasoning, exact tool identity, fragmented SSE, cancellation, no replay after partial output and zero-usage image accounting. Whitespace checks pass; dependency audit reported zero vulnerabilities.

The initial API probe missed Codex's internal `agent_message` type; the real worker failure exposed it, and the added conversion is now covered locally and by the successful delegated command. Uploaded-file operation, long-session generation quality and desktop stability remain outside these bounded checks. The original repetition incident had no raw upstream stream, so this does not claim every repetition loop has the same cause.

[...]

PR #683, opened 2026-09-10
fix: restore client-declared native tool identities

asifnawaz:
Codex can send native tools with already-flattened names such as `image_gen__imagegen`. MCP-only recovery leaves other registered tools with a provider name that Codex cannot dispatch. Restore the client-declared namespace and name before the existing tool adapters run.

Recovery requires live, directly exposed function or custom declarations backed by canonical turn metadata. It reuses existing schema, alias, history and response handling; ambiguous identities and ordinary-name collisions remain unchanged.

Restoration runs only on routes that flatten tools: Chat Completions, direct DeepSeek Responses and Console Go's compatibility path. Other Responses-native routes retain the client's declaration shape. A router-level Meta regression checks flat definitions, history, existing tool-choice policy and returned call identities in streaming and non-streaming modes; it fails on the previously unguarded restore.

Follow-up to merged #681, based on `main` at `da395a1`. One commit across seven files.

Validation:
- All 166 tests in the four affected namespace/custom-tool/Responses test files pass.
- `npm run check`, whitespace checks and final diff review pass.
- Full Linux run: 3,882 passed, 27 skipped; nine desktop tests initially failed because the local checkout control script was group-writable. After correcting that local permission, both affected test files pass (86 passed, one skipped). No source change was needed.
- Tests use synthetic loopback requests. Live provider execution and ImageGen generation were not exercised; cross-platform CI is tracked separately.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1944,6 +1944,14 @@ retry rules on the shared path.
   is not delivered. Flatten namespaces/deferred tools through the existing
   adapters; retain native top-level `apply_patch` and bridge other custom tools.
   Restore exact namespace/name identities, including plain-name collisions.
+- Codex can flatten native and MCP function/custom declarations before sending
+  them to a routed provider. Restore namespaces only for live, directly exposed
+  tools identified by the request's canonical turn metadata, and only on routes
+  that flatten tools. Responses-native routes that skip flattening retain the
+  client's declaration shape. Restore before app expansion and normal flattening
+  so client schemas, custom formats and collaboration model constraints use the
+  existing adapters. Never infer tools from name prefixes or create declarations
+  from metadata alone; ambiguous and ordinary-name collisions remain unchanged.
 - GLM thinking, legacy DeepSeek thinking and Command Code's DeepSeek Flash Chat
   route carry reasoning through LiteLLM as assistant `thinking` parts, restored
   by the forwarder to `reasoning_content`. Remove only successfully carried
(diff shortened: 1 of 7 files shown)


duolahypercho:
Looks good. Merged onto current `main` locally and the three changed test files pass (133/133). CI is green everywhere.

One request before merge: `restorePreflattenedToolNamespaces` now runs on every route, including Responses-native providers that skip flattening (Copilot, Meta, OpenCode, generic Responses). Those previously received Codex's flat tools unchanged; now they get a synthesized `type: "namespace"` tool. Probably fine, but nothing tests it. Please add a test for a non-flattening Responses route with pre-flattened input, or limit the restore to the flattening branches.

asifnawaz:
> Looks good. Merged onto current `main` locally and the three changed test files pass (133/133). CI is green everywhere.
> 
> One request before merge: `restorePreflattenedToolNamespaces` now runs on every route, including Responses-native providers that skip flattening (Copilot, Meta, OpenCode, generic Responses). Those previously received Codex's flat tools unchanged; now they get a synthesized `type: "namespace"` tool. Probably fine, but nothing tests it. Please add a test for a non-flattening Responses route with pre-flattened input, or limit the restore to the flattening branches.

Thanks for reviewing this! I have limited restoration to the routes that flatten tools and added a Meta Responses regression covering pre-flattened definitions, history and returned call identities in both streaming and non-streaming modes :-)

PR #691, opened 2026-09-11
docs: document ChatGPT account blocking for GPT-6-like custom model slugs

duolahypercho:
## Summary

Documents a known Codex-side limitation where recent Codex builds reject custom provider model slugs that resemble native OpenAI model names when using a ChatGPT account.

**This PR does not fix the block** — it provides documentation until the actual code fix (PR #673's provider-switch mechanism) lands.

Related to #689.

## Problem

Users reported that custom OpenAI-compatible provider models with GPT-6-like slugs (such as `unorouter/gpt-6-astra`) are blocked with:

```
The '<provider>/<model>' model is not supported when using Codex with a ChatGPT account.
```

Other models from the same provider (like `unorouter/grok-4.6`) work correctly. The reporter also confirmed that renaming the slug to `unorouter/agi-1` still triggered the same block, indicating Codex's validation examines more than just the public slug.

## Root Cause

Investigation revealed that:
- The error message originates from **Codex/OpenAI server-side validation**, not from this router
- Recent Codex builds validate custom-provider model slugs to prevent impersonation of native OpenAI models
- The validation triggers for slug patterns like `gpt-6-*` and other names resembling official OpenAI models
- This happens **before** the request reaches the router, so the router cannot bypass it

Evidence:
- The error string is only *matched* in this codebase (in `subagent-certify.mjs`) to detect when Codex rejects a model
- It is never *generated* by the router
- PR #673 explains this is a Codex server-side validation requiring a provider-switch code fix
- Issue #645 reported similar problems with native Astra access

## What This PR Does

Since the actual fix requires code changes in PR #673, this PR provides documentation:

1. **Adds troubleshooting documentation** in `docs/TROUBLESHOOTING.md`:
   - Explains the root cause clearly (Codex server-side validation)
   - Notes that renaming slugs does not reliably work (confirmed by reporter)
   - Provides current workaround: sign out of ChatGPT to use login-free mode
   - Points to PR #673 as the planned code fix
   - Explains what the router cannot do (bypass occurs before router is reached)

2. **Updates CHANGELOG.md**:
   - Clearly states "This router release does not fix the block"
   - Points to PR #673's provider-switch mechanism as the actual fix
   - Uses "Related #689" instead of "Fixes #689"

3. **Adds regression tests** verifying that `accountRefusal()` correctly detects this error pattern:
   - Test for original slug pattern (`unorouter/gpt-6-astra`)
   - Test for renamed variants that still trigger the block (`unorouter/agi-1`)
[...]

PR #692, opened 2026-09-11
fix: preserve native models under signed routing (fixes #689)

### LINKED ISSUE #689: [Bug]:  unorouter/gpt-6-astra model blocked server-side for ChatGPT-account sessions, but unorouter/grok-*  works
## What happened?

Provider: UnoRouter ([link](https://unorouter.com/en)) — OpenAI-compatible, base URL https://api.unorouter.com/v1, API-key auth.

Models configured (~/.local/share/codex-router/config/providers.json):
- unorouter/gpt-5.6-luna (upstreamModel: gpt-5.6-luna) — works
- unorouter/gpt-5.6-terra (upstreamModel: gpt-5.6-terra) — works
- unorouter/gpt-5.6-sol (upstreamModel: gpt-5.6-sol) — works
- unorouter/grok-4.5 (upstreamModel: grok-4.5) — works
- unorouter/grok-4.6 (upstreamModel: grok-4.6) — works
- unorouter/gpt-6-astra (upstreamModel: gpt-6-astra) — BLOCKED

## Steps to reproduce:
    1. Install codex-router with unorouter, API key configured.
    2. Select unorouter/gpt-6-astra in the Codex picker.
    3. Send any prompt.

### Actual behavior:
{"detail":"The 'unorouter/gpt-6-astra' model is not supported when using Codex with a ChatGPT 
account."}

## Tried workarounds:
- Renamed slug to unorouter/agi-1, gateway to unorouter-agi-1, display to AGI-1, kept upstreamModel 
gpt-6-astra. Same error, now showing agi-1. 
- unorouter/gpt-5.6-luna and unorouter/grok-4.6 work in the same session with the same ChatGPT account.

## Question:
How were unorouter/gpt-5.6-luna and unorouter/grok-4.6 allowed? Is this an OpenAI-side restriction outside the router's control?

## Evidence (Images):

Image

----

Image

----

Image

### Which app do you route through?

Codex

### Platform

macOS

_No response_

duolahypercho:
Rebased and refreshed PR #673 to fix issue #689.

## Changes

This PR introduces versioned `provider-switch` signed routing (v4) to preserve native ChatGPT models while correctly routing custom provider models through the gateway.

### Core Fix

- **Issue #689**: Custom provider models like `unorouter/gpt-6-astra` were being rejected under ChatGPT-account Codex with "The '...' model is not supported when using Codex with a ChatGPT account."
- **Solution**: Provider-switch signed routing ensures official GPT slugs stay on the native backend while custom provider models route through the gateway as intended.

### Key Implementation

1. **Signed Routing v4 (`provider-switch` mode)**:
   - Introduced in `src/config-manager.mjs` with state management for `previousProviderPresent` and `previousModelProvider`
   - Preserves ChatGPT authentication while routing explicit custom providers

2. **Route Resolution** (`src/router.mjs`):
   - New `resolveRegisteredRoute()` function distinguishes between:
     - Official native models (must passthrough to ChatGPT)
     - Aliased models
     - Custom provider models (route to gateway)
   - Updates `/models` endpoint to return both `data` (OpenAI-compatible) and `models` (Codex custom-provider discovery)

3. **Native Model Protection** (`src/native-alias.mjs`):
   - Added `isOfficialNativeSlug()` and `officialModelsMustPassthrough()`
   - Ensures native GPT models are never captured by external routes

### Testing

Added comprehensive test coverage:
- ✅ `signed routing keeps official slugs off aliases and native redirect`
- ✅ `signed routing routes custom provider models (issue #689)` - **now uses the actual failing case**

**Test fixture now uses the real-world failing slug:**
- Registers generic OpenAI-compatible provider `unorouter` (via `generic-providers.json`)
- Model slug: `unorouter/gpt-6-astra` (exactly as reported in #689, not the generic `custom/` namespace)
- Proves that named provider models like UnoRouter route to gateway while official GPT models stay native

The test fixture matches how users actually register custom providers:
- Generic provider with `adapter: "openai-chat"`
- Provider credential reference
- User model under the named provider namespace

### Documentation

Updated `CHANGELOG.md` with:
- Description of the fix for issue #689
- Explanation of provider-switch signed routing behavior

## Related

- Fixes #689
- Supersedes #673 (rebased with conflict resolution)
- Related to docs-only PR #691 (but provides the actual fix)

<div><a href="https://cursor.com/agents/bc-e27812f6-5d13-4567-9f9e-dfc603752c15?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e27812f6-5d13-4567-9f9e-dfc603752c15&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@
   unfinished initial event within the existing 10 MiB bound and match the
   namespace relay's limit, so later MCP calls retain their client identities.
   Prelude timeouts, empty-completion checks and retry safety remain intact.
+- **Signed routing preserves native models under provider-switch mode.** Issue
+  #689: Custom provider models like `unorouter/gpt-6-astra` are now routed
+  correctly under ChatGPT-account Codex without rejection. Provider-switch
+  signed routing (v4) ensures official GPT slugs stay on the native backend
+  while custom provider models route through the gateway as intended. Fixes #689.
 - **DeepSeek empty-completion guard allows large reasoning after liveness
   release.** Issue #684: Direct DeepSeek V4.1 Flash MCP turns with large
   reasoning deltas no longer hit the empty-completion byte limit prematurely.
--- a/src/catalog.mjs
+++ b/src/catalog.mjs
@@ -22,6 +22,7 @@ import {
   NATIVE_ALIAS_PATH,
   NATIVE_CATALOG_PATH,
   PORTS,
+  SIGNED_PROVIDER_MODE_PATH,
 } from "./paths.mjs";
 import {
   codexAuthStatus,
@@ -504,16 +505,42 @@ function selectedModel() {
 // GPT slugs are unusable there even when a ChatGPT credential file exists.
 // Mode toggles pass the desired state via MODEL_ROUTER_LOGIN_FREE because they
 // rebuild the catalog before rewriting the Codex config.
+//
+// Signed routing also sets `model_provider = "codex-router"`. That is not
+// login-free: official GPT slugs must stay on the ChatGPT account.
+export function loginFreeFromSignals({
+  envLoginFree,
+  modelProvider,
+  signedRoutingStatePresent,
+  loginFreeStatePresent,
+}) {
+  if (envLoginFree === "1") return true;
+  if (envLoginFree === "0") return false;
+  if (signedRoutingStatePresent && !loginFreeStatePresent) return false;
+  if (modelProvider === "codex-router") return true;
+  return undefined;
+}
+
 function loginFreeConfigured() {
   const override = process.env.MODEL_ROUTER_LOGIN_FREE;
-  if (override === "1") return true;
-  if (override === "0") return false;
-  if (!existsSync(CONFIG_PATH)) return false;
+  if (!existsSync(CONFIG_PATH)) {
+    const fromSignals = loginFreeFromSignals({
+      envLoginFree: override,
+      modelProvider: undefined,
+      signedRoutingStatePresent: existsSync(SIGNED_PROVIDER_MODE_PATH),
+      loginFreeStatePresent: existsSync(CODEX_PROVIDER_MODE_PATH),
+    });
+    return fromSignals === true;
+  }
   try {
     const document = scanTomlDocument(readFileSync(CONFIG_PATH, "utf8"));
-    if (tomlStringValue(document, [], "model_provider") === "codex-router") {
-      return true;
-    }
+    const fromSignals = loginFreeFromSignals({
+      envLoginFree: override,
+      modelProvider: tomlStringValue(document, [], "model_provider"),
+      signedRoutingStatePresent: existsSync(SIGNED_PROVIDER_MODE_PATH),
+      loginFreeStatePresent: existsSync(CODEX_PROVIDER_MODE_PATH),
+    });
+    if (fromSignals !== undefined) return fromSignals;
     if (!existsSync(CODEX_PROVIDER_MODE_PATH)) return false;
     // Identity-preserving login-free mode deliberately leaves model_provider
     // unchanged, so the root assignment alone can no longer identify it.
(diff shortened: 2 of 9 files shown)


duolahypercho:
Requesting changes. CI on the current head is green except the Windows job, which failed on `direct DeepSeek Responses preserves images, reasoning, tools and stream boundaries` with a 502 (unrelated to this diff; the previous head's Windows run was green, so a rerun should clear it).

**Blocking**

1. `src/router.mjs` `resolveRegisteredRoute`: native-redirect is skipped for every official GPT slug whenever `signed-provider-mode.json` exists. The redirect exists for exhausted-quota users, who are signed in (see the header of `src/native-redirect.mjs`), so this removes it for exactly its audience, and the new test pins the regression. The PR body's "model failover behavior is unchanged" does not hold for this path. Let an explicit redirect opt-in still win, or gate the passthrough on v4 state only.
2. The `issue #689` routing test passes unchanged on `origin/main` (verified by copying only the test file onto a main checkout). The router already routes a properly registered `unorouter/gpt-6-astra`; nothing in `router.mjs` here changes that. The PR does not explain why that model was rejected while `unorouter/gpt-5.6-luna` and `grok-4.6` worked in the same config. The reporter's error body is `{"detail": ...}`, the ChatGPT backend's shape, which means the router forwarded the turn natively because no route matched. If the real trigger is a mis-registered or unlisted user model, the root-provider switch will not close #689. A deeper fix that helps regardless: refuse a provider-prefixed slug with no registered route with a 400 that names the slug and the provider state, instead of forwarding it natively.
3. `officialModelsMustPassthrough` is a bare `existsSync`: v2/v3 states and stale or invalid files (ones config-manager itself rejects) switch on the new routing and sanitization rules. Validate the state and require version 4.

**Should fix**

4. Stateless reasoning sanitization now applies to every official turn under signed routing, not just mixed threads: reasoning items without `encrypted_content` and all bare `rs_` references are dropped even in pure-native threads. Trigger on payload evidence (an item with `encrypted_content: null`) rather than on the mode file.
5. `/models` (unauthenticated, polled by readiness probes and the Control Center) now returns the full merged catalog, ~150 entries with ~50 fields each including `base_instructions`. Serve the native shape only to Codex custom-provider discovery, keep the public leaf ids-only, and cite where Codex 0.153.4 reads `models` from this endpoint.
6. `isOfficialNativeSlug` reads and parses `native-models.json` on every call before trying `/^gpt-/`; test the regex first and cache the parsed set on size+mtime.

**Minor**

7. `test/native-alias.test.mjs`: the new test writes the signed state file into the shared stateDir and never removes it.
8. `loginFreeConfigured` evaluates the same signals in both branches; compute once.

duolahypercho:
Closing. #707 (merged as f6e2b4d7) fixes #689 at its actual cause. A provider-prefixed slug the running router has no route for is now refused locally with `unrouted_model` instead of being forwarded to ChatGPT, and ChatGPT's refusal is the error in the report.

The root-provider switch isn't needed. On Codex 0.153.2 with a ChatGPT login, a prefixed slug reaches the router with provider `openai`, so Codex does not reject it client-side. The #689 routing test here also passes on main without this diff. The review blockers were still present at the last push: native redirect is skipped for official slugs, and the state gate is a bare `existsSync`.

The part worth keeping, dropping unstored `rs_` reasoning ids before a native request, is covered by the request-local version in #699.

PR #693, opened 2026-09-11
fix(windows): create background consoles with CREATE_NO_WINDOW (supersedes #674)

duolahypercho:
## Summary

Consolidation of #674 by @jaakjuu1 on top of current `main`. It keeps the parts of that PR that fix the reported windows and drops the workarounds that regressed other guarantees. The contributor's commit is preserved in the branch history.

### Mechanism

- Windows 11 with Windows Terminal as the default terminal ignores the hidden window style (`SW_HIDE`) that conhost honored ([microsoft/terminal#12570](https://github.com/microsoft/terminal/issues/12570), [#15311](https://github.com/microsoft/terminal/issues/15311)).
- conhost only leaves a console windowless when the process was created with `CREATE_NO_WINDOW` (`ConsoleConnectionDeservesVisibleWindow` in `src/host/srvinit.cpp`), and every descendant inheriting that console stays windowless.
- libuv turns Node's `windowsHide` into `CREATE_NO_WINDOW` **only when no stdio slot is inherited**; with `stdio: "inherit"` it sets `SW_HIDE` alone.
- Under the Control Center, `control.mjs` runs inside a GUI-subsystem executable with no console, so every console child spawned without `CREATE_NO_WINDOW` allocates a fresh, visible Terminal window. That is the refresh window (`schtasks.exe` queries had no `windowsHide`; the Job Object self-wrap inherits stdio). The scheduled task's `cmd.exe` started with window style 0 is the persistent one.

### Changes

| Path | Change |
| --- | --- |
| `src/service-windows.mjs` | `schtasks` passes `windowsHide`. The VBS launcher prefers `.venv\Scripts\pythonw.exe` + generated `start-codex-router.pyw`, which runs `cmd.exe /D /C start-codex-router.cmd` with `CREATE_NO_WINDOW`; the direct `cmd.exe` form stays as the no-venv fallback. New `render-python` verb; uninstall removes the `.pyw`. The minute heartbeat trigger (#581, #618, #630) is kept. |
| `src/process-tree.mjs` | `runProcessTree` with `stdio: "inherit"` relays through pipes when no standard stream is a TTY on win32, so both the PowerShell owner and the contained command start windowless. Real consoles keep the child attached as before. |
| `test/windows-hidden-consoles.test.mjs` | Scan now covers `schtasks.exe`. |
| `test/process-tree.test.mjs`, `test/service-render.test.mjs` | Pin the relay shape; render both launchers, assert the cmd fallback, parse the generated Python when an interpreter is available; heartbeat tests restored. |
| `AGENTS.md`, `CHANGELOG.md`, `docs/TROUBLESHOOTING.md` | Document the rule and the reinstall step. |

### Not carried over from #674, and why

- Heartbeat trigger removal: regresses #581 relaunch after power events. A tick dropped by `IgnoreNew` starts no process, and one that relaunches goes through the windowless launcher.
- ACL hardening skipped under `ELECTRON_RUN_AS_NODE`: the existing helper already spawns with `windowsHide` and non-inherited stdio, so it is windowless as is.
- Direct `Codex Router.exe` spawn instead of the Job Object owner, and the `control.mjs` self-wrap skip: removes kill-on-close containment; the relay above fixes the visible window instead.
- `targetIsActive` pid-file probe: changes presence semantics from installed-or-loaded to running.
- csc-compiled winexe and the `pythonw` job host: unsigned exe compiled at install time and a second process host; the venv's `pythonw.exe` covers the one place `CREATE_NO_WINDOW` cannot be requested from Node.
- `windowsHide` on `start.mjs` children with inherited stdio: a no-op per the libuv rule above; children now inherit the windowless console from the launcher.

### Verification

- `node scripts-check.mjs`
- `node --test` on `windows-hidden-consoles`, `service-render`, `windows-task-state`, `tray-service-windows`, `gateway-restart`, `startup-cleanup`, `process-tree`, `control-center-electron`, `electron-companion`, `model-overlay-publication`, `service-write-guard`, `windows-operations`, `windows-parity`, `windows-launch-diagnosis`, `service-lifecycle`, `spawnable-command`: all pass on macOS (Windows-only cases skipped).
- Not verified here: the live Windows 11 + Windows Terminal behavior. @jaakjuu1, if you can, reinstall the service from this branch (`./codex-router.ps1 install`), then check Control Center refresh and the scheduled task; the `.pyw` and the `test/service-render.test.mjs` wscript case run on the Windows CI job.

Supersedes #674.


duolahypercho:
Closing in favor of letting #674 land the fix; this branch stays as a reference implementation to cherry-pick from.

PR #699, opened 2026-09-11
fix: make signed ChatGPT routing rollback-safe

B1gM8c:
## Summary

Supersedes #673 and reworks it from the current `main` branch in response to the maintainer review at https://github.com/duolahypercho/codex-router/pull/673#issuecomment-5629821547.

Current Codex validates a prefixed external model against the selected provider before sending the request. Under a ChatGPT login, the root `openai` provider therefore rejects a routed slug before it can reach the local router. This change makes the explicit **Use Router with ChatGPT** toggle select a dedicated ChatGPT-authenticated `codex-router-signed` provider, while keeping the change opt-in, rollback-safe, ownership-checked, and independent from native redirect.

## What changed after #673

1. **Documented and constrained the root-provider exception**
   - Updated `AGENTS.md`, `SECURITY.md`, the README, and the architecture guide.
   - Only an explicit signed-routing toggle from root OpenAI may select `codex-router-signed`.
   - Install, update, repair, and catalog refresh do not silently opt users into or migrate this switch.

2. **Kept the state readable by the previous release**
   - Reused the existing version-1 provider-switch state instead of introducing version 4.
   - Ordinary update preserves active v1 state byte-for-byte.
   - Catalog refresh remembers whether it parked `root-openai` or `provider-switch` mode and restores that same mode.
   - The managed provider block is validated exactly before update or disable; drift fails closed.
   - A cross-version check confirmed that current `main` can disable state written by this branch, restore `model_provider = "openai"`, and clear the state file.

3. **Preserved native redirect as a separate operator control**
   - Removed the #673 behavior that implicitly bypassed aliases and native redirect for signed sessions.
   - Native redirect remains active until explicitly cleared, regardless of signed routing or failover.
   - Both controls now warn about that independence where the operator changes them.

4. **Corrected the subagent documentation without overclaiming**
   - Recorded the exact ordinary-routing evidence: Codex CLI 0.153.4, `codex login status` reporting ChatGPT login, and the request reaching `/v1/responses` through the dedicated provider.
   - Explicitly states that this does **not** certify subagent checks 3-5 and that no quota-consuming subagent certification run was performed.

5. **Addressed the smaller review points**
   - Control Center no longer labels an unprefixed native GPT session as router traffic solely because its transport provider is `codex-router-signed`.
   - The signed-routing path adds no synchronous state or catalog reads to `/responses`; reasoning cleanup relies on request-local evidence.

6. **Retained the accepted fixes from #673**
   - `/models` serves both the OpenAI-compatible `data` array and the Codex account-catalog `models` array.
   - Mixed routed/native turns drop only an explicitly unstored reasoning item and its paired `rs_` reference, using request-local evidence rather than reading router state on the hot path.

## Verification

- `npm run check`
- Focused config, refresh, catalog, caller-key, routing, native-alias, and Control Center harness tests
- Control Center: `npm run check`, all 99 tests (98 passed, one Windows-only skip), and production build
- Full root test suite: all failures were limited to two environment-specific baseline failures reproduced unchanged on unmodified `main`:
  - the local browser test rendered the Chinese `路由在线` label while the test expected English `Router online`
  - the user's global pre-commit hook rejected the repository's literal environment-variable example in an installer rollback fixture
- Cross-version downgrade: enable with this branch, disable with unmodified current `main`, verify provider restoration/state removal, and confirm Codex parses the restored config
- `git diff --check`
- Secret scan of the complete PR diff

## Security and scope

This PR contains no local router configuration, API keys, tunnel identifiers, private endpoints, credentials, generated catalogs, or runtime state. It changes only the general routing bug, rollback behavior, documentation, and tests.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1633,13 +1633,17 @@ about it.
   `model_catalog_json` block plus its marked `model_providers.codex-router`
   table and, when the user has no concurrency preference, its marked
   `[agents].max_concurrent_threads_per_session` default. It may change the root
-  `model_provider` only when the user explicitly
-  enables the tray's login-free mode. In that mode it may also select an
-  enabled external `model`; snapshot both previous values in protected router
-  state and restore them exactly when the mode is disabled.
+  `model_provider` only when the user explicitly enables either login-free mode
+  or signed routing from a root-OpenAI configuration. Signed routing selects
+  the dedicated, ChatGPT-authenticated `codex-router-signed` provider; ordinary
+  install, update, repair, and catalog refresh must never create or migrate that
+  switch implicitly. Keep its state readable by the previous release and
+  restore the prior provider exactly when it is disabled. Login-free mode may
+  also select an enabled external `model`; snapshot both previous values in
+  protected router state and restore them exactly when the mode is disabled.
 - Preserve reasoning settings, profiles, projects, trust, MCP configuration,
   features, and ChatGPT authentication. Preserve `model` and `model_provider`
-  outside the explicitly enabled login-free mode.
+  outside those explicitly enabled routing modes.
 - A user-initiated macOS tray login-mode change may gracefully restart only the
   registered Codex desktop app. This does not authorize an installation task to
   quit Codex, and the tray must never force-terminate it.
--- a/README.md
+++ b/README.md
@@ -1355,6 +1355,22 @@ When the Codex runtime is executing inside WSL, a Windows-style path such as
 If setup appears successful but the Desktop model picker does not change, check
 which Codex home was modified before rerunning setup.
 
+### Use external models while signed in to ChatGPT
+
+The Control Center's **Use Router with ChatGPT** switch keeps ChatGPT
+authentication available while external provider models remain selectable. On
+current Codex builds, an explicit switch from the built-in OpenAI provider
+selects the managed `codex-router-signed` transport so Codex validates prefixed
+model ids against the router before sending them. The prior provider is stored
+in protected state and restored when the switch is turned off. Normal updates
+and catalog refreshes do not silently opt an existing installation into this
+provider switch.
+
+The optional native redirect is independent of this switch and of model
+failover. If native redirect is set, every unmatched native GPT turn that
+reaches the router continues to use its configured external route until
+`./bin/control native-redirect clear` is run.
+
 ### Use Codex without an OpenAI login
 
 The tray's **Use without OpenAI login** switch selects the managed custom
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -121,8 +121,10 @@ credentials.
 The config manager:
 
 - Writes only a marked `openai_base_url` and `model_catalog_json` block.
-- Preserves `model`, `model_provider`, reasoning settings, profiles, and ChatGPT
-  authentication.
+- Preserves `model`, reasoning settings, profiles, and ChatGPT authentication.
+  It changes `model_provider` only for an explicit login-free or signed-routing
+  toggle, records the previous value in protected state, and restores it after
+  an ownership check.
 - Refuses to replace an unmarked user-owned base URL or catalog.
 - Creates `~/.codex/config.toml.pre-codex-router` before its first change.
 - Atomically rewrites the config and restricts it to the current user.
--- a/apps/control-center/electron/ipc.mjs
+++ b/apps/control-center/electron/ipc.mjs
@@ -817,6 +817,15 @@ function readFileEdges(filePath) {
   }
 }
 
+export function codexSessionProvider(model, sessionProvider) {
+  if (model?.includes("/")) return model.split("/", 1)[0];
+  // Signed routing uses a dedicated transport provider so current Codex can
+  // validate external slugs while keeping ChatGPT authentication. Native GPT
+  // turns still go to OpenAI unless the operator separately enabled native
+  // redirect, so do not label them as router traffic in the session list.
+  return sessionProvider === "codex-router-signed" ? "openai" : sessionProvider;
+}
+
 function codexRolloutMetadata(filePath) {
   let sessionMeta;
   let turnContext;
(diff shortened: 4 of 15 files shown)


duolahypercho:
Thanks @B1gM8c. This handles the #673 review carefully: the state stays readable by the previous release, update and refresh don't silently switch anyone's mode, native redirect stays independent, and the ownership checks fail closed.

I'm not taking the root-provider switch without evidence that it's needed. #689 turned out to be an unregistered model being forwarded natively, and #707 (merged) now refuses those locally. With Codex 0.153.2 and a ChatGPT login, `codex exec -m unorouter/zz-probe-689` reached the router as provider `openai`, and the router forwarded it natively, so Codex did not validate the slug client-side. If you have a capture on current main where Codex rejects a prefixed slug that *does* have a route before the request reaches the router, please attach it. Until then `docs/SUBAGENT-CERTIFICATION.md` should stay as it is.

The request-local `rs_` cleanup in `normalizeNativeInput` fixes a real gap. #484 only covers substituted callers. A caller using their own ChatGPT auth who switches from a routed model to GPT still forwards unstored reasoning items and their `rs_` references. #692's test for this fails on main and passes on this branch. Could you open that change as its own PR, with a test that doesn't write a signed-routing state file? Please leave the `/models` change out of it. That endpoint is unauthenticated and polled constantly, and the full catalog includes `base_instructions`.

B1gM8c:
Thanks — I reworked this along the boundary you suggested.

I first tested the provider-validation concern against current `main` with Codex 0.153.4 while signed in to ChatGPT. A registered prefixed slug reached the router over `/v1/responses` and completed. The negative control, an unregistered prefixed slug, was refused locally with `400 unrouted_model`, and the mock upstream request count did not increase. So I could not reproduce a current need to switch the root provider, and I have dropped that part of this proposal.

The historical 400 was also recoverable from local logs: that old task used `provider=OpenAI` with `transport="responses_websocket"` and never reached the router's HTTP path. Current Codex reports both Responses WebSocket feature flags as removed/disabled, and the current canary uses HTTP, so that historical bypass is not a current-main Router regression.

I opened #716 for only the request-local unstored `rs_` reasoning cleanup. It pre-scans the request for a full reasoning item whose own missing/invalid encrypted payload proves that specific id was not stored, then drops only that item and its matching reference. Unrelated bare references remain untouched. Its regression test does not create or read signed-routing state.

The new PR contains no root-provider switch, `/models` change, certification-doc change, private catalog, endpoint, or credential.

duolahypercho:
Thanks for the thorough rework, and thanks especially for retesting the premise honestly and reporting that it didn't reproduce — that's the most useful thing in this thread.

Given your own follow-up and #716, I think this one should close. Recording what I verified, so nothing is lost if the provider-validation question comes back:

**The premise doesn't hold on current main.** `src/unrouted-model.mjs` from #707 is a router-side guard for the #689 slug shape. It can only fire if a prefixed slug reaches the router, which is incompatible with Codex rejecting it against the root `openai` provider before sending. That matches both of our live captures.

**The rollback check had a gap.** The cross-version run only covered `signed-disable`. Enabling with this branch and then running the previous release against the state:

```
[PR699] signed-enable  → exit 0, state {"version":1,"managedProvider":"codex-router-signed",...}
[main]   enable        → exit 1
  Error: A recognized older signed-routing mode is still active; turn it off before updating the router.
  at src/config-manager.mjs:1568
[main]   signed-disable → exit 0 (model_provider restored)
```

`enable` is the ordinary install/update/repair/refresh path, so a downgraded install can't update until the user manually disables. Separately, main's `signed-disable` restores `model_provider` but leaves `[model_providers.codex-router-signed]` behind, because removal only exists in this branch.

**The `/models` change.** Measured against a real 150-model catalog: the body goes from 10,937 bytes to 6,849,529 — a 626x increase on a constantly polled endpoint, with `base_instructions` on all 150 entries. Good call leaving it out of #716.

One note for the future: this PR amends the `AGENTS.md` safety boundary and `docs/SUBAGENT-CERTIFICATION.md` to permit the behaviour it introduces. Those docs should follow an accepted behaviour change, not ship with the proposal for it.

Everything passes here — `npm run check`, `routing` 132/132, `config-manager`/`refresh-catalog`/`caller-key`/`control-center` 87/87, Electron 69 pass + 1 skip, CI green. The work is sound; it just isn't needed. Moving the review to #716.

duolahypercho:
Reviewed line by line — **merge**. No CRITICAL or HIGH findings.

I traced the paths that matter for a rollback-safety change:

- `writeSignedProviderModeState`/`atomicWrite` are both write-temp-then-rename at mode 0600 with `protectPrivateFile`, so a failed write cannot leave a truncated or zero-byte config.
- Restoring a file that never existed does the right thing: `previousPresent: false` → `replaceRootValue(..., undefined)` removes the key rather than writing an empty one.
- `--preserve-root-openai` through `signed-enable` and `refresh-catalog.mjs` does restore the *previously active* mode rather than silently migrating.
- Drift detection fails closed — `signedProviderSwitchBlockStatus` throws and leaves the file untouched when the managed table was hand-edited, and the test proves byte-identical content on that path.
- Against AGENTS.md: no credential reader is touched (~1668 clean), and `model_provider` is only mutated inside explicit `signed-enable`/`login-free-enable`, never inside plain `enable`, with unrelated root keys left alone (~1640 clean). No credentials logged or written world-readable.

Two low notes, neither blocking: the `/models` and stateless-reasoning changes in `router.mjs` are scope creep relative to the title (legitimate and tested, just unrelated); and the state-file-then-config-write ordering leaves a narrow crash window where the state file could describe a newer state than the config — but that is the pre-existing pattern shared with login-free mode, and it is fail-safe, since `signedProviderStateIsOwned` throws "lost ownership" rather than corrupting anything.

The tests genuinely pin the regression — `ordinary refresh does not migrate an existing root-OpenAI v3 signed state` reproduces the exact bug, and `refresh-catalog.test.mjs` asserts the precise CLI args per recorded mode.

Issue #654, opened 2026-09-07
[Bug]: commandcode/hy4-preview emits chain-of-thought as visible assistant output


moreo365:
### What happened?

`commandcode/hy4-preview` preserves ordinary turn continuity, but hidden reasoning is copied into visible assistant output and canonical task history.

Across several plain conversation turns, the stored final assistant message begins with internal planning prose and contains a literal closing marker shaped like `</think:<redacted>>`, while Codex also stores a separate structured reasoning item.

This exposes chain-of-thought to the user and causes it to accumulate in subsequent context and output tokens.

### Environment

- Codex Desktop on Windows
- codex-router 0.5.1 operating tree based on upstream commit `d1805f635de365761a7ff515fff94865567649b4`
- model: `commandcode/hy4-preview`
- Command Code GOAT plan

### Evidence

The canonical Codex task record contains both:

1. a structured reasoning item; and
2. an assistant final-answer item containing internal planning prose plus a closing `</think:<redacted>>` marker.

No prior-turn history loss was observed for this model in the same conversation. No private transcript or raw log is attached.

### Registry/profile asymmetry

The checked-in `config/commandcode/hy4-preview.json` entry advertises only `high` reasoning effort and has no `requestProfile`.

Other checked-in Hy4 routes advertise no-think/deep-thinking choices and use `requestProfile: "hy4-reasoning"`. The API forwarder already contains Hy4 effort normalization behind that profile.

This may be intentional because the Command Code endpoint returns a different response shape, but the current route leaks reasoning either way.

### Expected behavior

Reasoning should remain in the structured reasoning channel. Only answer text should become visible assistant output and be replayed as ordinary assistant content.

### Requested diagnostic/fix

Please inspect the Command Code stream shape and determine whether Hy4 reasoning is returned under a separate reasoning field or embedded in `content`.

- If the endpoint supports the existing Hy4 reasoning profile, apply the correct effort/profile mapping to this route.
- If it embeds reasoning in ordinary content, please add provider-specific normalization or expose a supported no-think mode rather than forwarding private reasoning as the answer.
- Please avoid a broad text-deduplication filter that could remove legitimate repeated answer text.

No credentials, managed caller URLs, local paths, account identifiers, or raw private logs are included.

PracticalSwan:
I traced the current `main` Command Code egress and reproduced why the existing reasoning sanitizer does not cover this report's shape.

[...]

PR #703, opened 2026-09-11
fix: replay Chat reasoning for DeepSeek reached through a reseller

Corpocore:
This branch carries two independent fixes that came out of the same debugging
session. They are unrelated, so if you would rather review them separately, say
so and I will split the second into its own PR -- the collapse commit
(`f43990c`) stands alone.

---

### What breaks

`usesNativeChatReasoning` gates both halves of the Chat reasoning replay: the
carry in `carryReasoningThroughInput` (`src/router.mjs`) and the restore in
`restoreNativeReasoningContent` (`src/api-forwarder.mjs`). It matched on the
vendor request profiles (`glm-thinking`, `deepseek-thinking`) plus a single
Command Code slug. A DeepSeek route reached through a reseller carries a
reseller-shaped profile instead, matched none of those, and so ran neither hop.

The measured case is `opencode-go/deepseek-v4.1-flash`, whose profile is
`auto-tool-choice`. DeepSeek rejects a follow-up whose prior assistant turn is
missing its `reasoning_content`:

```
HTTP 400  The `reasoning_content` in the thinking mode must be passed back to the API.
```

The rejection needs a prior assistant turn, so it lands on the turn *after* a
reply rather than on the reply itself. That makes it read as a poisoned
conversation -- every subsequent request fails -- while the turn that produced
the offending assistant message looked completely healthy.

A subagent hand-off always ends in prose, so spawning one was enough to start
it. From the local usage meter, on `opencode-go/deepseek-v4.1-flash`, with 200s
either side:

```
{"status":200, ... "requestId":"...:45"}
{"status":400, ... "requestId":"...:46"}
{"status":200, ... "requestId":"...:47"}
```

and the gateway log for those same rows names the route and the reason:

```
16:17:16 - LiteLLM Proxy:ERROR - litellm.BadRequestError: OpenAIException -
Error from provider (Console Go): Upstream request failed: [invalid_request_error]
The `reasoning_content` in the thinking mode must be passed back to the API.
Received Model Group=opencode-go-deepseek-v4-1-flash
```

repeated at 16:17:41, which is the second 400 row. The rejection is the
provider's; the profile mismatch is what leaves the router unable to satisfy it.

### The fix

Key the check on the upstream model for reseller Chat routes, because the replay
rule belongs to the upstream and not to the route's profile or its reseller:

```js
const RESELLER_DEEPSEEK_CHAT_PROVIDERS = new Set(["opencode-go"]);

function isResellerDeepSeekChatRoute(model) {
  if (!RESELLER_DEEPSEEK_CHAT_PROVIDERS.has(model?.provider)) return false;
  return /(^|\/)deepseek/i.test(String(model?.upstreamModel ?? ""));
}
```

The existing profile checks and the Command Code clause are unchanged, so no
currently-covered route changes behaviour.

### Why it is scoped to DeepSeek

`opencode-go` exposes 21 reasoning-capable Chat routes: DeepSeek, GLM, Qwen,
Kimi, MiniMax, Longcat and MiMo. A provider-wide rule would move the replay
channel for all of them from visible assistant text to `reasoning_content`, and
only the DeepSeek requirement is established here. Routes that neither need nor
expect that field would lose the reasoning they currently receive as text, which
is a silent behaviour change on models this issue was never reported against, so
the fix is deliberately narrow.

**Open question for maintainers:** GLM on the same endpoint (`glm-5.3`,
`glm-5.3-flash`) reaches z.ai, whose direct routes are already covered through
`glm-thinking`. If the reseller routes share that upstream's requirement, they
belong in the same set. I did not add them without a reproduction -- if that
evidence exists or is easy to gather, the predicate is the place to extend.
`opencode-go-messages` is excluded in any case: it speaks the Anthropic
protocol, where reasoning travels as thinking blocks, so the
`reasoning_content` restore does not apply.

---

### What breaks

On routed Chat routes the bridge can emit a turn's answer text twice inside a
single assistant `message` item -- the same token run and then the same run
again, with `output_text.done` carrying both copies. On a real session it
affected roughly one message in five, and the affected messages were
consistently the short progress lines a model writes immediately before a tool
call.

The duplicate is in the wire bytes. Instrumenting the egress on both sides of
the router's transforms shows the pre-transform stream already containing it,
so it is not a client rendering artifact and not an item-lifecycle reorder;
every captured stream opened and closed each output item sequentially.

Honest scoping: the **origin is not established**. Only the shape is. This
removes a measured duplicate at the wire rather than explaining why the upstream
repeats it.

### The fix

A new egress transform, `src/duplicate-block-collapse.mjs`, rewrites a message
whose entire text is a run of identical blank-line-separated blocks:

```
A\n\nA            ->  A
A\n\nA\n\nB\n\nB  ->  A\n\nB
```

Everything else is relayed byte-for-byte. It runs after
`ItemLifecycleNormalizer`, which guarantees the item being held is opened,
streamed, and closed before the next one begins, so holding an item cannot
reorder the stream. A different output index arriving mid-hold releases
everything verbatim instead.

### The bug in the first version of this fix, and why it is worth reading

The first version bounded the held item at 64 KB of total bytes. That was wrong,
and it failed in the field while passing every synthetic test. An output item
also carries the model's **reasoning**, which on a thinking route dwarfs the
answer. A captured turn held 138 KB for one message item, of which 113 KB was
reasoning against a ~300-byte answer:

```
held bytes for the message item: 138030
  of which reasoning:            113578
  MAX_HELD_BYTES:                 65536   <- tripped
```

The guard tripped mid-item, released it verbatim, and the duplicate survived in
exactly the place it was most common. The bound now weighs the visible answer
text separately from total memory, and the regression test asserts its fixture
exceeds the old budget so it fails against the defect rather than merely
passing alongside the fix.

---

## Testing

- `node --test test/chat-reasoning.test.mjs` -- 2 pass, 1 skipped (the optional
  LiteLLM proof, which needs `MODEL_ROUTER_TEST_LITELLM_PYTHON`).
- `node --test test/duplicate-block-collapse.test.mjs` -- 7 pass, including a
  chunk-boundary sweep (1/3/17/64/4096 bytes) and the reasoning-volume
  regression above.
- Focused set across `duplicate-block-collapse`, `item-lifecycle-normalizer`,
  `chat-reasoning`, `routing`, `deepseek-v4-1-flash` and `subagent-routing`:
  **163 tests, 162 pass, 1 skipped, 0 fail**.
- `npm run check` clean.

Both fixes were additionally reproduced against captured live traffic before
and after their respective transforms, rather than only through fixtures.

I did not run the optional offline LiteLLM replay proof locally.

## Docs

`AGENTS.md` gains the reseller reasoning rule and a section on the egress
duplicate collapse, including the text-versus-total bound and the requirement
to capture on both sides of the transform. `CHANGELOG.md` has Unreleased entries
for both.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1991,6 +1991,32 @@ pinned LiteLLM and run `node --test test/chat-reasoning.test.mjs`. It uses only
 loopback services and synthetic credentials, with duplicate negative controls.
 See the provider's [Responses contract](https://api-docs.deepseek.com/guides/responses_api/).
 
+## A duplicated assistant block is collapsed at egress
+
+LiteLLM's chat-completions bridge can emit a turn's answer text twice inside one
+assistant `message` item: the same token run, then the same run again, with the
+terminal `output_text.done` carrying both copies. It is observable on routed
+Chat routes and it is in the wire bytes, not a client rendering artifact and not
+an item-lifecycle reorder. `src/duplicate-block-collapse.mjs` rewrites a message
+whose entire text is a run of identical blank-line-separated blocks (`A\n\nA`,
+`A\n\nA\n\nB\n\nB`) down to one copy each, and relays everything else
+byte-for-byte.
+
+- The rewrite inspects only the visible answer text. An output item also carries
+  the model's reasoning, which on a thinking route is routinely an order of
+  magnitude larger; bounding the hold by total item bytes made the guard trip on
+  long reasoning and release the item verbatim, so the duplicate survived
+  exactly where it was most common. Keep the text bound and the memory bound
+  separate.
+- It runs after `ItemLifecycleNormalizer`, which guarantees an item is opened,
+  streamed, and closed before the next one begins, so holding one message item's
+  events cannot reorder the stream. If a different output index arrives while an
+  item is held, release everything verbatim rather than reordering.
+- Captures that prove a rewrite must be taken on both sides of this transform.
+  A post-transform-only capture cannot distinguish "the duplicate never
+  arrived" from "the transform removed it", which is how the reasoning-volume
+  defect above stayed hidden behind an apparently clean stream.
+
 ## Substituting a prompt-token count a provider reported as zero
 
 Codex decides when to compact from the `input_tokens` each response reports, so
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1972,6 +1972,16 @@ retry rules on the shared path.
   source items or change other native Responses routes. Keep this policy shared
   between hops without applying direct DeepSeek sampling parameters to resellers.
   Command Code's schema-strict `/alpha/generate` fallback remains separate.
+- A reseller's DeepSeek routes carry reasoning the same way regardless of the
+  route's own request profile, because the rule belongs to the upstream model
+  and not to the profile or the reseller. OpenCode Go's DeepSeek routes carry
+  `auto-tool-choice`, which no profile check matches, so a profile-only policy
+  left them unprotected and the provider rejected the turn after any prose
+  reply with "The `reasoning_content` in the thinking mode must be passed back
+  to the API". Scope this to the DeepSeek upstream: the same endpoint serves
+  GLM, Qwen, Kimi, MiniMax, Longcat and MiMo thinking models, and moving their
+  replay from visible text to `reasoning_content` without evidence that they
+  require it would be a silent behaviour change.
 
 Regression coverage lives in `test/deepseek-responses-routing.test.mjs`,
 `test/namespace-relay-custom.test.mjs`, `test/chat-reasoning.test.mjs` and the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,22 @@
   back to the model as an ordinary tool error. A non-string `content`, a
   completed item that disagrees, and streamed text contradicted by the final
   input still fail closed. Reproduced offline against pinned LiteLLM 1.96.0.
+- **A reseller's DeepSeek routes replay Chat reasoning again.** The Chat
+  reasoning carry was scoped to the vendor request profiles (`glm-thinking`,
+  `deepseek-thinking`) plus Command Code's single DeepSeek Flash slug, so any
+  DeepSeek route reached through a reseller was left unprotected. On
+  `opencode-go/deepseek-v4.1-flash` -- whose profile is `auto-tool-choice` --
+  neither hop ran. The provider rejects a follow-up whose prior assistant turn
+  is missing its `reasoning_content`, and that failure needs a prior turn, so
+  the error surfaced on the turn *after* any reply: HTTP 400 "The
+  `reasoning_content` in the thinking mode must be passed back to the API." A
+  subagent hand-off always ends in prose, which is why spawning one broke every
+  following request in the conversation. The check now keys on the upstream
+  model for reseller Chat routes. Only the DeepSeek upstream is covered -- the
+  same endpoint's GLM, Qwen, Kimi, MiniMax, Longcat and MiMo thinking models
+  keep their existing replay channel, because flipping theirs without evidence
+  would be a silent behaviour change. Regression coverage reads the shipped
+  registry in `test/chat-reasoning.test.mjs`.
 - **Switching a conversation back to OpenAI no longer fails on routed item IDs.**
   Routed providers mint their own item IDs (`call_...`, `tool_...`,
   `chatcmpl-...`), Codex saves them, and OpenAI rejects them on replay with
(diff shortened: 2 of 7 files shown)


duolahypercho:
Thanks for this — and first, credit where it is due: your diagnosis of the reseller DeepSeek replay bug is correct, and the `MODEL_BY_SLUG` registry sweep in your test is a better pattern than the hand-built objects we had. I have ported that sweep into #727.

I reviewed this against #715 and #727 by executing all three predicates over every `(provider, upstreamModel, requestProfile)` triple in `config/**/*.json` — 190 shipped routes. Result: **#727's family table is a strict superset of your chat-reasoning fix.** It uses the byte-identical `/(^|\/)deepseek/i` with `opencode-go` in its provider set, so the "handled here, missed there" set is empty, and it additionally covers GLM, Kimi K3, MiniMax M3 and Hy3 behind the same resellers.

So I'd like to **close the chat-reasoning commit (`68cd8c43`) as subsumed**, with the credit above recorded in #727.

**The `duplicate-block-collapse` commit (`1bea0223`) is a different matter and I'd rather it lived as its own PR.** Two reasons:

1. **As written it stops streaming for every routed answer.** `duplicateBlockCollapseTransform` is created for every routed SSE response (`route ? … : undefined`, no provider or model gate). On `response.output_item.added` with `item.type === "message"` it starts holding and pushes nothing until `response.output_item.done`. The release valves — 256 KiB of text, 4 MiB of bytes, 120 s — are all far beyond a normal answer. So on every routed provider the reasoning streams and then the entire visible answer appears at once. I checked `ItemLifecycleNormalizer`, which holds *other* indices while an item is open but passes the open item's own events straight through, so this is newly introduced. With nothing on the wire during a long generation, a client idle timer can also abandon a live turn (`ResponsesHeartbeatTransform` is Grok-only).

2. **It may be collapsing a symptom whose cause #727 removes.** Your module header describes "a short pre-tool progress line arrives twice"; #715's evidence for the same window describes Hy4 emitting the same sentence 2, 4, 8, 16 times *because* its reasoning was replayed as visible text — same `opencode-go` chat routes. Plausibly your own reasoning fix removes the input that produces the duplicate. Worth re-measuring after #727 lands before deciding how much collapsing is still needed.

Other things I found in that transform, for when you resplit it:

- if a held item has no `response.output_text.done`, every delta is `continue`d and the replacement is never inserted, so the client gets framing with no text — and the rewrite path is only entered on exactly the misbehaving providers
- `response.content_part.done` is not patched, so a populated `part.text` still ships the duplicate (the fixture's `content_part.done` carries no `part`, which is why the "one per carrier" assertion passes at 3)
- `collapseRepeatedBlocks` is broader than its header: it drops *any* adjacent identical block, so an empty fenced code block (``` / blank / ```) collapses to one fence and breaks everything after it
- multi-content-part messages are flattened — all `content_index` values concatenate into one string and every `output_text.done` gets the combined text

**One merge-order warning:** this branch's registry test asserts `usesNativeChatReasoning === false` for every non-DeepSeek `opencode-go` model. #715 makes `hy4-preview` true and #727 makes `glm-*`, `kimi-k3` and `hy3` true. Git merges this cleanly and the suite goes red — a semantic conflict, not a textual one.

duolahypercho:
#727 landed as `c1f6e69e`, so the chat-reasoning half of this is now on `main`. One correction to my earlier review: the family table didn't ship with the byte-identical `/(^|\/)deepseek/i` — it tightened to `/(^|\/)deepseek-(v\d|reasoner|flash)/i` so the non-thinking `deepseek-chat` alias stays out of the contract. It's still a strict superset of your fix in practice: every DeepSeek id `opencode-go` publishes (`deepseek-v4-flash`, `-v4-flash-free`, `-v4-flash-vision-exp`, `-v4-pro`, `-v4.1-flash`, and the `deepseek/`-prefixed forms) matches. A two-turn `/responses` probe carrying a prior reasoning item plus `reasoning.effort: high` returned 200 with a turn-2 reasoning item on `deepseek-v4-flash`, `-v4-pro` and `-v4.1-flash` — your diagnosis was right and the route is fixed. Your `MODEL_BY_SLUG` registry sweep is in #727's tests.

That leaves only `duplicate-block-collapse`, which I'd rather not carry on a branch that's now conflicted in `src/chat-reasoning.mjs`, `src/router.mjs` and `test/chat-reasoning.test.mjs`. Closing this as subsumed.

The transform isn't rejected — it's parked on one measurement. Every duplication report we have (#739, the remaining half of #654, and your own capture) was taken on a build that predates #715 and #727, and the working theory is that replaying reasoning as visible text is what produced the repeats. If duplicates survive on current `main`, the transform has a real job and I'd take it as a fresh PR, with the four issues from my earlier review addressed — chiefly that it currently holds the entire visible answer of every routed response until the item closes. #739 is on `commandcode/qwen3.8-flash`, which the family table deliberately doesn't cover, so that's the case most likely to prove you right.

Thanks for the diagnosis — it was the correct one and it shipped.

PR #708, opened 2026-09-11
fix: show Chat Completions reasoning in Codex

duolahypercho:
## Problem

Codex showed no reasoning for Chat Completions routes. That includes `commandcode/hy4-preview`, `opencode-go/deepseek-v4.1-flash` and every other route that goes through LiteLLM's Chat Completions → Responses bridge. The models were reasoning, but Codex threw the reasoning away:

- Recent Codex Desktop rollouts on `opencode-go/deepseek-v4.1-flash` contain **0 reasoning items across 204 assistant messages**.
- `grok-oauth/grok-4.6` (which has the existing repair) and `zai-coding/glm-5.3-flash` (a Responses-native provider) do record reasoning.

A live capture through the router shows the broken bridge shape:

```text
response.output_item.added            id=msg_…   out=0 message
response.content_part.added           id=msg_…   out=0
response.reasoning_summary_text.delta id=rs_127734569…  out=0
response.reasoning_summary_text.delta id=rs_547957579…  out=0
response.reasoning_summary_text.delta id=rs_-6290613820… out=0   (×45, a fresh hashed id each)
response.output_text.delta            id=msg_…   out=0
```

The stream never sends a reasoning `output_item.added` or `reasoning_summary_part.added`, and the deltas sit on the message's `output_index`. Codex drops deltas that don't belong to an open item. This is the same failure #597 fixed for Grok OAuth only.

A second problem sits on top of the first. LiteLLM echoes the request's `instructions` and full `tools` array in `response.created` (`streaming_iterator.py`). In Codex Desktop that frame is larger than the repair's 256 KiB pre-commit bound, which switches the repair off for the whole stream. This is the same prelude that #690 fixed for the namespace relay.

## Fix

- `reasoningSummaryCompatTransform` attaches the existing lifecycle repair to every provider whose `protocol` is Chat Completions (`openai`, the default):
  - Direct `deepseek` is excluded because `DeepseekToolMessageCompatTransform` already repairs its bridge.
  - `anthropic` and `openai-responses` providers never pass through this bridge, so they get no new stage.
- Grok OAuth keeps its gateway-error normalization. Other routes relay an untyped LiteLLM error envelope byte-identical, after closing the reasoning item as `incomplete` and releasing any message the repair was holding.
- The pre-commit frame bound rises from 256 KiB to 10 MiB, matching the namespace relay's. The committed bound stays at 64 MiB.
- AGENTS.md records the scope, the error-envelope rule and the prelude bound. CHANGELOG entry added.

## Verification

- **Offline replay of a real Hy4 stream.** A `commandcode/hy4-preview` stream captured through the live router went through `repair → TranslatedToolMessageCompatTransform`. It came out as one reasoning item with a stable id and all 45 deltas, the message moved to `output_index` 1, and `response.completed.output` = `[reasoning, message]`. A tool-call turn in the DeepSeek live-fixture shape came out with reasoning at index 0 ahead of the call.
- **New tests:**
  - A router case with a Desktop-sized (>256 KiB) `response.created` on `commandcode/hy4-preview`.
  - Unit cases for that prelude, the factory scope and non-Grok error envelopes.
  - The existing Grok, direct-DeepSeek and translated-OpenCode router cases still pass.
- `npm run check` passes.
- `npm test`: 3,979 tests, 3,952 passed, 0 failed, 27 skipped.

Not yet done: a live Codex Desktop turn on the patched service showing a `reasoning` item in its rollout.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2083,6 +2083,39 @@ label, so `src/message-phase.mjs` assigns one.
    already sequential. Coverage lives in `test/message-phase.test.mjs` and the
    routed case in `test/namespace-relay-routing.test.mjs`.
 
+## Chat Completions reasoning reaches Codex as one reasoning item
+
+LiteLLM 1.96's Chat Completions to Responses bridge opens the assistant message
+first, then streams `response.reasoning_summary_text.delta` under a fresh
+hashed `rs_…` id per delta (or the message's id), with no reasoning
+`output_item.added` or `reasoning_summary_part.added` and on the message's
+`output_index`. Codex drops deltas that belong to no open item, so reasoning
+never rendered and no reasoning item was saved to the thread. That held for
+every Chat Completions route (measured on `commandcode/hy4-preview` and
+`opencode-go/deepseek-v4.1-flash`), not only Grok.
+
+1. **One repair, scoped by protocol.** `reasoningSummaryCompatTransform` in
+   `src/grok-reasoning-summary-compat.mjs` attaches the lifecycle repair to
+   every provider whose `protocol` is Chat Completions (`openai`, the default).
+   Direct `deepseek` is excluded because `DeepseekToolMessageCompatTransform`
+   already repairs its bridge, and `anthropic` and `openai-responses`
+   providers do not reach this bridge. Widening it to another protocol needs a
+   captured stream from that protocol first.
+2. **Grok's gateway-error wording stays on Grok OAuth.** Only `grok-oauth`
+   replaces an untyped LiteLLM error envelope with the fixed local error. Other
+   routes relay that envelope byte-identical, after closing the reasoning item
+   as `incomplete` and releasing any message the repair was holding.
+3. **The pre-commit frame bound is 10 MiB, like the namespace relay's.**
+   LiteLLM echoes the request's `instructions` and full `tools` array in
+   `response.created`, and a Codex Desktop tool list exceeds 256 KiB. A smaller
+   bound releases that frame raw and disables the repair for the whole stream,
+   which no mock gateway with a tiny prelude reproduces. Any fixture for this
+   path must echo a Desktop-sized tool list.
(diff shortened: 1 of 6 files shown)


duolahypercho:
Reviewed — **merge**, with one test gap worth closing.

The gate widening is architecturally protected the way it needs to be: the class distinguishes a canonical reasoning lifecycle (stable `item_id`, array-typed `summary`) from Grok's orphaned/hashed-id shape and passes canonical events through byte-identical, and `normalizeGatewayErrors: false` keeps non-Grok gateway errors unrewritten. `deepseek` is correctly excluded (it has its own repair), and `anthropic`/`openai-responses` protocols fall out of the same check — the factory-scoping test asserts all three against the registry's real protocol vocabulary.

The tests use the real echoed prelude rather than a bare envelope — the 400-tool frame asserts `Buffer.byteLength(created) > 256 * 1024`, and the router-level test runs the full mock gateway with the same real-sized prelude. AGENTS.md records the shapes as measured on `commandcode/hy4-preview` and `opencode-go/deepseek-v4.1-flash`, which matches what the fixtures look like.

**The gap:** there is no fixture exercising the widened gate against the GLM-thinking / `reasoning_content` lane. Both new tests are genuinely orphaned-hash-id cases, so nothing pins the claim that a canonical stream on a route restored by a *different* mechanism (`src/chat-reasoning.mjs`) still passes through untouched. I lean toward it being fine — the lifecycle check should cover it — but the passthrough is the property the widening depends on, and it is currently unasserted. A byte-identical-passthrough test for that lane, mirroring the existing Grok "canonical streams pass byte-identical" one, would close it.

Also worth noting `grok-api` (non-OAuth) newly enters the repair path. That looks intentional — same xAI-via-LiteLLM backend, same hashed-id defect class — just flagging it as a deliberate scope change rather than a side effect.

Same note as #710: going past the new 10 MiB bound is still an unlogged self-disable.

PR #710, opened 2026-09-11
fix: admit LiteLLM's echoed prelude in the DeepSeek stream repairs

duolahypercho:
## The bug

LiteLLM's Chat Completions to Responses bridge echoes the request's
`instructions` and whole `tools` array in `response.created` **and**
`response.in_progress`. Both SSE repairs in
`src/deepseek-tool-message-compat.mjs` — the blank-assistant-message cleanup
that serves every translated route, and direct DeepSeek's reasoning-bridge
repair — bounded a pre-commit frame at 256 KiB. A frame over that bound is
released raw and the repair disables itself for the rest of the response, so a
session with a large tool list kept LiteLLM's blank assistant message in the
turn and never got its DeepSeek reasoning item rebuilt. Mock gateways answer
with a bare envelope, so nothing in the suite crossed the bound.

This is the same prelude PR #690 raised `MAX_SSE_FRAME_BYTES` to 10 MiB for in
`src/namespace-relay.mjs`.

## Measured, not inferred

Probed offline against the repository's pinned LiteLLM 1.96.0
(`litellm.aresponses(..., use_chat_completions_api=True, stream=True)` against
a loopback mock upstream, synthetic credentials, no provider quota), with 300
tools and a 16 KiB `instructions` string:

| event | bytes | tools echoed |
| --- | --- | --- |
| `response.created` | 409,568 | 300 |
| `response.in_progress` | 409,572 | 300 |
| `response.completed` | 716 | `[]` |

## The byte bound was not the only one

Raising `MAX_FRAME_BYTES` alone left the new test failing on the same
assertion. `strictJsonPreflight` also budgets **object members per frame**
(8 KiB) and key code units (128 KiB), and a tool list is member-dense JSON
Schema. Both scan budgets are now derived from the frame bound at the density
the 256 KiB bound allowed — one member per 32 bytes, one key code unit per
2 bytes — so the previous values fall out unchanged at the old bound.

`maxCandidateBytes` (64 KiB) is deliberately **not** raised: a prelude frame is
relayed as soon as it parses and is never held, and `response.completed`, which
*is* held, echoes nothing (see the table).

Accumulator storage grown past an ordinary event is released once its frame is
taken, so one large prelude no longer pins that capacity for the whole stream.

## Tests

Written failing first, then fixed:

- `test/routing.test.mjs` — "router keeps DeepSeek message repairs behind a
  Desktop-sized prelude": one router, a translated `opencode-go/deepseek-v4-flash`
  turn and a direct `deepseek/...` turn, with the mock gateway echoing the
  request's real tool list the way LiteLLM does. It asserts the echoed prelude
  crosses **both** old budgets before asserting either repair.
- `test/deepseek-tool-message-compat.test.mjs` — "default bounds admit
  LiteLLM's echoed Desktop-sized prelude", covering both transforms directly,
  and pinning the member budget as a distinct bound by showing the same prelude
  still fails open byte-for-byte under a small `maxJsonMembers`.

Fail-open semantics are unchanged, and the existing small-limit cases still
pass.

## Verification

`npm run check` plus `deepseek-tool-message-compat` (91), `routing` (133),
`namespace-relay-routing` (31), `namespace-relay-custom` (6),
`deepseek-responses-routing` (7), `message-phase` (14), and
`empty-completion-router` (39) all pass on the rebased branch.

Failures seen earlier were machine load, not this diff: with other suites
saturating the host (load average >200), scattered Grok/health tests failed on
an **untouched `origin/main` worktree** identically, and each one passed 3-5/5
on both trees once load dropped.

## Cost, measured and then fixed

My first estimate here was wrong and is worth flagging: I measured ~80 ms per
MiB while another process had this machine at load average >200. Uncontended,
the honest numbers are much smaller, and the second commit removes most of what
was left.

The frame accumulator walked the prelude one byte at a time, which cost 49 ms
on an 8 MiB frame by itself. It now jumps between line feeds with `indexOf` and
copies each run in one `Buffer.copy`: 5.3 ms for the same frame.

End to end through the transform, uncontended:

| prelude | cost |
| --- | --- |
| 409 KiB (300 tools, the real case) | 3-6 ms |
| 1.5 MiB | ~11 ms |
| 7.6 MiB (near the bound) | ~53 ms |

The bulk scanner is verified against the previous per-byte implementation over
60,000 randomized streams and chunk boundaries — frames, separators, oversize
cuts, stop requests and remainders all identical — and a new test splits the
same stream at every byte offset, in both newline spellings, through both
transforms, since a separator can be cut across chunks in three places.

A regex fast path for the JSON uniqueness scan was also prototyped and
**dropped**: 400,000 differential cases passed but it was no faster.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1970,10 +1970,16 @@ the most tools, and no mock gateway with a bare envelope shows it.
    skip part of a frame to fit a budget. Accumulator storage grown past an
    ordinary event is released once its frame is taken, so one large prelude
    does not pin that capacity for the whole stream.
-4. **The scan costs CPU in proportion to the frame.** Every byte of a large
-   prelude is scanned synchronously on the router's event loop. Raising a bound
-   again, or adding another pre-commit parser on the routed path, needs a
-   measurement against a frame at the new bound, not only a passing test.
+4. **The scan costs CPU in proportion to the frame, so keep it bulk.** A
+   pre-commit frame is decoded, uniqueness-scanned, and parsed synchronously on
+   the router's event loop. Measured on the 409 KiB prelude that prompted this:
+   3-6 ms, scaling at roughly 7 ms per MiB to 53 ms at 7.5 MiB. The frame
+   scanner reaches that by jumping between line feeds rather than walking every
+   byte -- the per-byte loop it replaced cost 49 ms on an 8 MiB frame on its
+   own. Raising a bound again, or adding another pre-commit parser to the
+   routed path, needs a measurement at the new bound rather than only a passing
+   test. Measure uncontended: on a loaded machine these numbers inflate by more
+   than an order of magnitude and invite a fix for a cost that is not there.
 5. **Fixtures for this path echo a large, dense tool list.** "router keeps
    DeepSeek message repairs behind a Desktop-sized prelude" in
    `test/routing.test.mjs` asserts that its echoed prelude crosses both old
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1937,6 +1937,49 @@ cheap plans means speaking that route.
    and it is why the note stays on the registry entry now that the plan no
    longer blocks access outright.
 
+## LiteLLM's echoed prelude sets the pre-commit frame bounds
+
+LiteLLM 1.96's Chat Completions to Responses bridge copies the request's
+`instructions` and whole `tools` array into both `response.created` and
+`response.in_progress` (`_default_response_created_event_data` in
+`litellm/responses/litellm_completion_transformation/streaming_iterator.py`);
+its `response.completed` does not echo them. Measured against the pinned
+LiteLLM offline, with no provider quota: a 300-tool list and a 16 KiB
+`instructions` string make `response.created` and `response.in_progress`
+409 KiB each, while `response.completed` stays at 716 bytes and carries
+`tools: []`. A stream repair that meets a frame over any of its pre-commit
+budgets releases the bytes raw and switches itself off for the rest of the
+response, so a small budget disables the repair in exactly the sessions with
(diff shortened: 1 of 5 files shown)


duolahypercho:
Reviewed — **merge**. The thing I most wanted to check is the one this gets right: **both bounds were raised, not just the byte bound.**

`MAX_FRAME_JSON_MEMBERS` and `MAX_FRAME_JSON_KEY_CODE_UNITS` are now derived from `MAX_FRAME_BYTES` (`/32` and `/2`) rather than flat. Raising the byte bound alone would have left a Desktop-sized tool list failing the same frame open on the member budget — and the test proves it directly by re-running the same fixture with the old `maxJsonMembers: 8 * 1024` and asserting the output passes through raw.

The `SseFrameAccumulator` rewrite from byte-by-byte to bulk `indexOf` checks out: `from` strictly advances every iteration so there is no infinite loop, and the cross-chunk `byteBefore()` lookback and "first separator wins, LF before CRLF" ordering both match the original semantics. `take()` copying out and dropping `#storage` back to `Buffer.alloc(0)` above 256 KiB means there is no per-connection high-water leak; worst case is N concurrent streams × ~10 MiB, which matches the precedent `namespace-relay.mjs` already set.

Test quality is good — `a frame split at any byte boundary repairs identically` fuzzes every chunk-split offset for both line-ending styles and both transform classes, and the router-level test echoes the actual request body back the way real LiteLLM does, asserting the prelude crosses *both* old bounds before checking repair correctness.

One note that applies to #708 as well: crossing the (now 10 MiB) bound is still a silent self-disable — zero `console.` calls in the file. The fallback is safe and AGENTS.md documents it as deliberate, but if a future prelude ever exceeds 10 MiB the original bug recurs with no operator-visible signal at all. Worth a one-line warn at some point.

PR #714, opened 2026-09-11
Publish routed models into five more coding clients, and keep them current

duolahypercho:
Adds opencode, pi, omp (oh-my-pi), Command Code, and Hermes Agent to the Harness tab. Each keeps its providers in a configuration document the user also owns, so one shared publisher (`src/routed-harness-*.mjs`) writes the single `codex-router` key each of them reads and leaves every other byte alone.

Devin CLI and T3 Code are deliberately absent: Devin CLI selects from Cognition-hosted models and has no custom base URL, and T3 Code drives other CLIs rather than talking to models itself. A row that cannot publish is a row that lies.

## What is here

- **Publish** — `control client-setup <id>` / `client-disconnect <id>` and the Harness row's **Set up**. The wire is always one the router already serves: Responses on the caller `/v1` path for opencode, pi, and omp; the existing Anthropic Messages surface with `codex_router/anthropic/<slug>` ids for Command Code and Hermes, which have no Responses client.
- **Stay current** — `control client-update <id>|--all`, an **Update** button per row, and **Update all**. Each runs the client's *own* updater (`opencode upgrade`, `pi update --self`, `command-code update`, `hermes update --yes`) rather than `npm install -g`: a CLI installed by Homebrew or a `curl | sh` script is not an npm package, and reinstalling it as one leaves two copies whose winner is PATH order. Updating is never a side effect of publishing.
- **A keep-alive after the terminal event is dropped** (`src/openai-adapters.mjs`) — see below.
- **pi is installed from its maintained package**: it moved from `@mariozechner/pi-coding-agent` (stalled at 0.73.1) to `@earendil-works/pi-coding-agent`. The abandoned name installs fine and answers `pi --version`, so nothing here reported it as wrong.

## The stream fix, verified end to end

OpenCode Go and Zen close every Responses stream with a `ping` after `response.completed`. The forwarder read that as data after the terminal event and appended `invalid_responses_stream`, which LiteLLM re-raised on every completed turn. Codex ignores bytes after a terminal event, so it went unnoticed; opencode's AI SDK and pi validate them and failed the turn.

1. **Reproduced on the live router** — a real turn on the free `opencode-free-responses/muse-spark-1.2-contributor-free` route returned `"status":"completed"` with usage counted, then `{"error":{"message":"litellm.APIError: Response API in-stream error","code":"500"}}`.
2. **Confirmed the frame** — the raw upstream from `opencode.ai/zen/v1/responses` ends with exactly `event: ping` / `{"type":"ping","cost":"0"}`.
3. **Replayed those real bytes through both transforms:**

```
upstream : … response.output_item.done, response.completed, ping
OLD emits: … response.output_item.done, response.completed, error
NEW emits: … response.output_item.done, response.completed
only difference: error
```

Real data after the terminal event is still refused.

## Other verification

- opencode 1.18.13 read a published document from a scratch config and listed all 13 routed models (exit 0, empty stderr). This is the check that caught opencode's `limit.output` schema requirement, which the unit suite had passed straight through.
- The JSONC guard fires on a real layout: with an `opencode.jsonc` present, publication stops with an explanation and writes nothing.
- 190 tests pass / 0 fail across the harness, adapters, control-center, control, target-integration, caller-key, and YAML suites.

## Not verified

No routed turn has been taken *through* one of the five clients end to end — the live service runs from the main checkout, so proving that needs a second router instance. `AGENTS.md` now records that listing alone is insufficient, which is exactly how the `ping` bug hid.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2305,6 +2305,27 @@ of somebody else's file; `routed-harness-manager.mjs` is the publisher.
   Command Code first reads `providers.json` in 1.30.0, so `minimumVersion` on
   its catalog entry makes setup update an older CLI and makes status and doctor
   report one. A version the CLI will not report is unknown, not outdated.
+- **Staying current is its own action, and it runs the client's own updater.**
+  `@latest` in `npmPackage` only decides what a *first* install fetches;
+  `installRoutedHarness` deliberately leaves a CLI that is already there alone,
+  because bumping somebody's global coding agent must not be a consequence of
+  republishing a model list. `control client-update <id>|--all`, the Harness
+  row's **Update** button, and `updateRoutedHarness` are the paths that do it.
+  Each prefers the client's own updater (`opencode upgrade`, `pi update
+  --self`, `command-code update`, `hermes update --yes`) over `npm install -g`:
+  a CLI installed by Homebrew or a `curl | sh` script is not an npm package,
+  and reinstalling it as one leaves two copies whose winner is PATH order —
+  which shows up as a row reporting the new version while the shell keeps
+  running the old one. npm is the fallback only for a client that publishes a
+  package and ships no updater. `--all` skips a client that is not installed
+  (update what I have, not install five agents I never asked for) and reports
+  per client rather than stopping at the first failure. A client that reports
+  no version before and after is never called "updated".
+- **Check the package name is still the maintained one.** pi moved publishers:
+  `@mariozechner/pi-coding-agent` stopped at 0.73.1 and the live line is
+  `@earendil-works/pi-coding-agent`. An abandoned package still installs and
+  still answers `pi --version`, so nothing in this repository would have
+  reported it as wrong — only reading the client's own install docs does.
 - **Prove a publication against the real client, not against its docs.** The
   unit suite passed while opencode 1.18 rejected every published model, because
   its schema requires `limit.output` whenever `limit` is present. opencode now
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2311,7 +2311,10 @@ of somebody else's file; `routed-harness-manager.mjs` is the publisher.
   gets `limit.input` at the router's `autoCompact` and `limit.output` as the
   headroom above it, or no `limit` when there is no threshold. A change to any
   adapter needs the same check: publish into a scratch document, then have the
-  installed client list or use the models from it.
+  installed client list the models from it and take one real turn (a free route
+  costs nothing). Listing alone missed that OpenCode's post-completion `ping`
+  had become a trailing gateway error: Codex ignores bytes after a terminal
+  event, and opencode and pi do not.
 - **Devin CLI and T3 Code are deliberately absent.** Devin CLI's config selects
   from Cognition-hosted models and has no custom base URL, so routed models
   cannot be published into it; the `devin-cli` *provider* is the other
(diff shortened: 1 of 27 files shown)


duolahypercho:
Reviewed this stack line by line. Pushed a fix for the one finding I'd call blocking; the rest are notes.

## Fixed in 02eec5f1

**Uninstall could empty the user's configuration file.** Both YAML edits judged a node by `parent.children.size`, which is the lexer's map of keys it was *able to register*. Two things live inside a node without appearing there:

- a block sequence, a merge key, or a key `PLAIN_KEY` declines (`openrouter/free:`)
- a trailing comment block, which `endIndex` deliberately stops before — and publishing inserts *above* it, so the comment ends up inside the block removal then splices away

Reproduced on omp's `models.yml`:

```yaml
providers:
  - name: a
    api: http://a
```

Install produced a document PyYAML cannot parse (`ParserError while parsing a block collection`); uninstall returned `""` and the file was emptied. A `providers:` holding only a comment lost the `providers:` key and left the comment orphaned at the wrong indentation.

`unaccountedLines` now walks the node's indented region instead of trusting `endIndex`. Publishing refuses with the file untouched when that region holds a non-comment line it cannot account for — the same "refuse rather than guess" rule this module already applies to an inline parent and to JSON with comments — and the ancestor collapse stops at anything unaccounted, comments included.

An empty document can still result and that case is correct: the collapse only ever reaches `keyPath[0]`, so emptiness now means the file held nothing but our own block, which is what omp's `models.yml` is.

Added byte-for-byte round-trip assertions — the existing ones match substrings and pass even when bytes we don't own are dropped. Both new tests fail against the previous reader.

## Not fixed — worth your call

- **A provider id the grammar declines now refuses rather than mis-publishes, but the underlying `decodeKey` weakness stands.** It returns `undefined` for a plain key outside `^[A-Za-z0-9_][A-Za-z0-9_.-]*$`, and its children then register against the *grandparent*. I scoped the guard to this module rather than changing `yaml-structure.mjs`, since the dsh manager shares it.
- **`writeDocument` can throw after the rename.** `writeHarnessDocument` does `write tmp → protect tmp → rename → protect target`, and that last call can throw. In `install()` it sits outside the `try { writeMarker } catch { rollback }`, so the document is published, no marker is written, and nothing rolls back — the "client pointed at a route the router has no record of" state the module header forbids.
- **Keep-alives are now dropped rather than relayed.** On main a `: keep-alive` reached the client as `data: \n\n`; after this PR nothing is pushed, and nothing in `src/` injects a heartbeat. A long silent turn leaves the socket completely quiet. Re-emitting `":\n\n"` would fix what strict clients rejected without going silent.
- **`"neither is reformatted"` in AGENTS.md is not true for the three JSON clients.** `applyJsonValue` returns `JSON.stringify(document, null, 2)`, so a 4-space CRLF `opencode.json` comes back 2-space LF on every republish — and `refreshTargetPickerIfInstalled` republishes.
- **On POSIX, Command Code falls back to an executable named `cmd`.** If `command-code` is absent but any unrelated `cmd` is on PATH, `status()` runs `cmd --version` and `client-update` runs `cmd update` against it.
- **"Update all" from the desktop app sets no `*_BIN` overrides** — `ROUTED_HARNESS_ROWS.find(row => row.id === "--all")` is `undefined`, so the child re-discovers via PATH, which the comment above it says a desktop app cannot rely on.

The updater subprocess path is clean: `updateCommand` is frozen literal data, the id is allow-listed three times, argv array, no shell.

PR #716, opened 2026-09-12
fix: preserve stored reasoning while removing duplicate references

B1gM8c:
## Summary

- Rebase the fix onto current `main` (`b42d553`).
- Keep the existing native reasoning normalization rules unchanged.
- After normalization, remove an `rs_` `item_reference` only when a full `reasoning` item with the same ID still exists inline.
- Preserve a sole stored-item reference and preserve full reasoning items whose `encrypted_content` is null, empty, non-string, or omitted; the existing sanitizer still strips unreadable ciphertext.

## Why

The previous predicate treated null/empty/non-string `encrypted_content` as proof that an item was unstored. As the review pointed out, that wire shape is ambiguous for a credential-bearing native caller when encrypted reasoning was not requested. Dropping the full item could therefore remove legitimate stored reasoning from in front of a later tool call.

The invariant we can prove locally is narrower: when the normalized request contains both a full reasoning item and an `item_reference` with the same `rs_` ID, the reference is redundant. Removing only that reference avoids duplicate-ID rejection without guessing storage state or stranding a sole pointer.

## Regression coverage

The routing regression now covers all of these cases in one direct-credential/native replay path:

- full reasoning plus same-ID reference: keep the sanitized full item, drop only the duplicate reference;
- null, empty, and non-string encrypted payloads without a matching reference: keep the sanitized full items;
- a bare unrelated `rs_` reference: preserve it;
- substituted/stateless continuation and compaction behavior: unchanged;
- no signed-provider state is written as a side effect.

A negative-control run against the pre-fix main revision (`34b8842`) failed at the exact new assertion because the full null-ciphertext reasoning item was deleted. The patched test passes.

## Verification

- `npm run check`
- `node --test test/routing.test.mjs` — 140 passed
- `node --test test/state-owner.test.mjs` — 11 passed
- focused routing regression — passed
- secret-pattern scan of the PR diff — no matches

The full suite was also run before the final rebase: 4,132 passed, 26 skipped, and 2 failed. Both failures were reproduced unchanged on an untouched main worktree: the desktop panel test expects English while the rendered control snapshot is Chinese, and the setup rollback test is intercepted by the machine-wide ECC pre-commit secret scanner. Neither failure touches this patch. After the final rebase, the complete routing and state-owner suites above were rerun successfully.

## Scope and privacy

Only `CHANGELOG.md`, `src/router.mjs`, and `test/routing.test.mjs` are changed. No local provider configuration, credentials, endpoint values, generated state, or installation artifacts are included.


duolahypercho:
Thanks for slicing this out of #699 — the duplicate `rs_` id (full item inline **plus** an `item_reference` for the same id) is a real hazard and worth fixing on its own.

Two things block it as written:

1. **The evidence predicate does not hold.** `encrypted_content: null` is the normal shape of a *stored* reasoning item whenever the caller did not ask for `include: ["reasoning.encrypted_content"]` — it says the ciphertext was not requested, not that the item was never stored. `src/router.mjs:2338-2342` states this design intent explicitly ("credential-bearing callers … can fall back to the native stored-item namespace after an unreadable encrypted payload is removed"), and the new code reverses it without amending that comment.

2. **The landed behaviour is wider than the description.** `test/routing.test.mjs:3142` asserts that the three `invalidStatelessPayloads` items are dropped, and none of them has a matching `item_reference` — so this is "drop every reasoning item with a null/empty ciphertext", not "drop only that item and its matching reference". The practical cost is a reasoning item dropped out from in front of its `function_call`, which the Responses API rejects: it trades a hypothetical 400 for a certain one. The new trigger is also ungated by caller flags, so it reaches the compaction-V1 path whose stored-reference contract `src/router.mjs:4058-4062` deliberately preserves.

A narrower version that I think gets the whole benefit: run the existing `flatMap` unchanged, then drop an `rs_` `item_reference` only when the same id is **still present inline** as a full reasoning item in the forwarded input. That removes the redundant reference, keeps the item (ciphertext stripped) so the stored-namespace fallback still works, and can never strand a reference that is the sole pointer — no flag gating needed.

Separately: could you attach the request capture that showed the failure? The path needs `substitutedCaller === false`, which today is essentially the unmerged signed-routing work, so a real payload would tell us whether the pair actually appears in the wild and which half the backend rejects.

duolahypercho:
Reviewed line by line. The change is well-scoped and the test genuinely fails when reverted — I traced the pre-fix branch to confirm. One question needs settling before I merge, and it is empirical rather than a code issue.

**Useful scheduling note first:** this does not conflict with #703, #708, #715 or #727. Those all touch the *output* transform pipeline in `handleResponses` (~4327-4460); this touches *input* normalization (`normalizeNativeInput`, ~2455) and its call site at ~4057. Disjoint ranges, no double-gating and no reasoning-dropped-twice risk, so it can land independently of that stack.

**The concern: the code is broader than the description.** The CHANGELOG frames the fix around a *pair* — a full `reasoning` item plus a following `item_reference` with the same `rs_` id. But `explicitlyUnstoredReasoningIds` is built from full reasoning items alone and never checks that a paired reference exists. The test confirms it: all three `invalidStatelessPayloads` entries have no `item_reference` at all and are still dropped, including for a direct-credential caller.

That matters because `{type: "reasoning", id: "rs_…", encrypted_content: null}` is also the shape a genuinely native, `store: true` item can have when the request did not ask for `reasoning.encrypted_content` in `include`. That is what the adjacent "historical format repair" was written for — its comment at `src/router.mjs:2337` says credential-bearing callers "can fall back to the native stored-item namespace after an unreadable encrypted payload is removed." This PR reverses that fallback for exactly the `null`/`""`/non-string cases, and the two situations produce byte-identical JSON, so the code cannot tell them apart.

**So: does OpenAI's Responses API ever serialize `encrypted_content: null` (own key, null value) for a real stored item whose caller did not request the include — or does it omit the key entirely?** If it omits the key, this is safe as written and I will merge it. If it emits `null`, this drops legitimately stored items for credential-bearing callers. Nothing in the repo can answer it; a live probe can. If you have a capture showing which, that settles it.

One incompleteness worth noting either way: the drop-set is rebuilt per request from full items only, so a later turn that sends just the bare `item_reference` — normal as Codex shrinks history — reproduces the original 400, since `dropUnstoredReasoningReferences` is false for a credential-bearing caller. Your commit message is upfront that the fix is scoped to "the request-local pair", which by construction cannot survive a turn boundary; flagging it so it is on the record rather than as a change request.

Also worth updating the `src/router.mjs:2337` comment, since it now describes a fallback this narrows.

duolahypercho:
Checks are green, but the branch conflicts with main and cannot be merged as it
stands. Please rebase onto current main and I will take another look.

B1gM8c:
Thanks for the review. I rebased the branch onto the current `main` (`b42d553`) and force-pushed `799a719`.

I also replaced the broad `encrypted_content` predicate with the narrower post-normalization rule you suggested:

- run the existing reasoning cleanup unchanged;
- collect only full `reasoning` items that still survive inline;
- drop an `rs_` `item_reference` only when that same ID is present in the surviving inline set;
- keep sole references and keep null/empty/non-string ciphertext reasoning items after the existing sanitizer removes unreadable ciphertext.

I do not have a new sanitized live capture that can distinguish the ambiguous stored/unstored wire shapes, so the updated patch no longer relies on that inference. Instead, the regression is scoped to the duplicate-ID invariant. A negative control on pre-fix main fails at the new assertion because the full null-ciphertext item is deleted; the patched routing suite passes 140/140 on current main. The PR description now includes the detailed verification and the two unrelated full-suite baseline failures.

PR #717, opened 2026-09-12
fix(tray): explain why the login-free and signed-routing switches are disabled

chrisl10:
## Problem

The two Codex auth-mode switches in the macOS tray disable each other, but a disabled switch gave no reason. With signed routing on, the **Use without OpenAI login** row stayed greyed while its detail line still read "Use connected models and restart Codex" — nothing named the switch holding it off, so the switch reads as broken rather than sequenced (observed on a signed-routing install, where the CLI answers `auth-mode on` with "Turn off signed routing before enabling login-free mode" but the tray never surfaces that because the row is not clickable).

## Change

Each row's detail line now names the prerequisite when it is disabled:

- **Use Router with ChatGPT** (disabled while login-free is on) → "Turn off 'Use without OpenAI login' first"
- **Use without OpenAI login** (disabled while signed routing is on) → "Turn off 'Use Router with ChatGPT' first"

No control-flow change: the fail-closed disable stays, only the explanation is added. Strings added to all five translated dictionaries (zh/ar/hi/ja/ko); English fallback covers anything else.

## Verification

`xcrun swiftc -parse` clean on the six touched Swift files. No behavior tests exist for tray copy; existing suites unaffected (macOS tray has no CI target on this file set).

### DIFF
--- a/apps/macos/ModelRouterTray/Sources/Localization.swift
+++ b/apps/macos/ModelRouterTray/Sources/Localization.swift
@@ -241,6 +241,8 @@ enum RouterChineseText {
     "Quotas and live activity pinned to the desktop": "将额度和实时活动固定在桌面",
     "Show provider usage and activity status": "显示提供商用量和活动状态",
     "Use Router with ChatGPT": "在 ChatGPT 中使用路由",
+    "Turn off 'Use Router with ChatGPT' first": "请先关闭「在 ChatGPT 中使用路由」",
+    "Turn off 'Use without OpenAI login' first": "请先关闭「不使用 OpenAI 登录」",
     "Share ChatGPT subscription": "共享 ChatGPT 订阅",
     "Sharing status unavailable": "无法读取共享状态",
     "Sharing enabled": "共享已启用",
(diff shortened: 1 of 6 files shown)


duolahypercho:
Reviewed — **merge**.

I checked the failure mode this kind of change usually has: one locale missing a key. All five translation tables get both keys — `RouterChineseText` inside `Localization.swift` plus the Arabic, Hindi, Japanese and Korean tables — and `apps/macos/ModelRouterTray/Tests/LocalizationTests.swift:89` enforces `Set(table.keys) == Set(RouterChineseText.values.keys)` for the four non-Chinese tables, so a skipped language would have failed CI rather than shipping a raw key. These are `[String: String]` dictionary entries, not struct members, so there is no compile risk either way; the only literal-level hazard is a duplicate key (Swift traps at init) and each file has both keys exactly once.

The conditions match too: each row's new detail text is shown under exactly the same predicate that already disables it on `main`, so the copy cannot disagree with the control.

One pre-existing oddity this makes visible rather than introduces: if `signedRouting` and `loginFree` are ever both on — reachable by driving the control CLI directly, since each setter is an independent call with no mutual-exclusion guard — both rows are disabled and each now tells the user to turn the other off first, with no way to do either from the panel. The deadlock already exists on `main`; only the messaging is newly self-referential. Not a blocker.

PR #718, opened 2026-09-12
fix: fail over on zai weekly-window and lapsed-package quota refusals

chrisl10:
## Problem

Two documented zai quota refusals classify as neither `out_of_usage` nor anything else, so the turn keeps its 429 instead of failing over to the next enabled model:

- **1310** — `Weekly/Monthly Limit Exhausted. Your limit will reset at {next_flush_time}`: a quota window that names neither `usage` nor `quota`, so `QUOTA_PATTERNS` misses it even though the named reset is exactly the sentence `PROVIDER_NAMED_RESET` already parses.
- **1309 / 1314** — `Your GLM Coding Plan package has expired and is temporarily unavailable. You can resume using it after renewing the subscription…`: renewal fixes it the way a top-up fixes a balance, and no other provider's answer changes, so it belongs to the swap path rather than surfacing as a dead turn.

Wording from the documented error table: https://docs.z.ai/api-reference/api-code

## Change

Two entries in `QUOTA_PATTERNS` (`src/error-translation.mjs`):

- `/(?:weekly|monthly|daily)\s+limit\s+exhausted/i` — 1310. Once classified, the existing `PROVIDER_NAMED_RESET` reads the body's reset stamp into the cooldown window, so the provider is also held off for the window instead of buying the same refusal once per turn.
- `/(?:package|plan|subscription) has expired/i` — 1309/1314. No window is invented: there is no named reset, so the swap carries no `until`.

Neither pattern overlaps `ENTITLEMENT_PATTERNS` (checked first, different vocabulary), and a lapsed package is not an entitlement: the plan did include the API.

## Verification

`node --test` on `model-failover`, `model-failover-router`, `error-translation`, `error-chain`, `rate-limit-quota`: 125 pass, 0 fail, including the two new cases (1310 swap with named window; 1309 swap without one).

### DIFF
--- a/src/error-translation.mjs
+++ b/src/error-translation.mjs
@@ -70,6 +70,13 @@ const QUOTA_PATTERNS = [
   // Both word orders occur in the wild: "usage limit reached" (zai) and
   // "reached your usage limit" (Kimi).
   /usage limit(?:s)? (?:reached|exceeded|hit)/i,
+  // zai 1310 names a weekly or monthly window without the words "usage" or
+  // "quota": "Weekly/Monthly Limit Exhausted. Your limit will reset at ...".
+  /(?:weekly|monthly|daily)\s+limit\s+exhausted/i,
+  // zai 1309/1314: a lapsed Coding Plan or enterprise package is renewed, not
+  // waited out or re-keyed: "Your GLM Coding Plan package has expired and is
+  // temporarily unavailable. You can resume using it after renewing...".
+  /(?:package|plan|subscription) has expired/i,
   /reached your (?:usage|monthly|daily) limit/i,
   /(?:monthly|daily|plan) usage limit/i,
   /purchase extra usage/i,
--- a/test/model-failover.test.mjs
+++ b/test/model-failover.test.mjs
@@ -100,6 +100,34 @@ test("classifyRoutedFailure recognizes the observed Z.ai five-hour window messag
   });
 });
 
+test("classifyRoutedFailure recognizes the documented Z.ai weekly and monthly window message", () => {
+  // zai 1310: "Weekly/Monthly Limit Exhausted. Your limit will reset at ..." --
+  // a quota window that names neither "usage" nor "quota", so the rest of the
+  // vocabulary does not catch it and the turn keeps its 429 instead of
+  // failing over.
+  const verdict = classifyRoutedFailure({
+    status: 429,
+    bodyText: quotaBody("Weekly Limit Exhausted. Your limit will reset at 2026-08-15 13:00:00."),
+    now: NOW,
+  });
+  assert.equal(verdict.swap, true);
+  assert.equal(verdict.reason, "out_of_usage");
+  assert.ok(verdict.until, "the named reset becomes the cooldown window");
+});
+
+test("classifyRoutedFailure swaps on a lapsed Z.ai coding or enterprise package", () => {
+  // zai 1309/1314: renewal restores access -- time does not, and neither does a
+  // fresh key. Still an exhausted plan every other provider can answer.
+  const verdict = classifyRoutedFailure({
+    status: 429,
+    bodyText: quotaBody(
+      "Your GLM Coding Plan package has expired and is temporarily unavailable. You can resume using it after renewing the subscription.",
+    ),
+    now: NOW,
+  });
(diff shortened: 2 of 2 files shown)


duolahypercho:
Thanks for this — the classification is right, and the cooldown handling is careful in a way I appreciate: `reason: "out_of_usage"` with an `until` only when the body actually carries a reset stamp, so a lapsed package is retried next turn instead of being written off for six hours.

I verified the problem is real on current `main`. Both bodies classify as neither `entitlement` nor `out_of_usage`, so `classifyRoutedFailure` returns `{swap:false}` and the 429 reaches the caller as a dead turn:

```
1310 weekly       kind=undefined  {"swap":false}
1310 monthly      kind=undefined  {"swap":false}
1309/1314 package kind=undefined  {"swap":false}
```

`/quota\b[^.]*\bexhausted/i` misses 1310 because the sentence never says "quota". Good catch.

One narrowing worth making before this lands. The second pattern is a word wider than it needs to be:

| body | want | current | anchored |
|---|---|---|---|
| `Your GLM Coding Plan package has expired…` (1309) | swap | ✅ | ✅ |
| `Your enterprise package has expired…` (1314) | swap | ✅ | ✅ |
| `Your subscription has expired.` | swap | ✅ | ✅ |
| `Your promotional plan has expired` | swap | ✅ | ✅ |
| **`The uploaded file package has expired`** | **no swap** | **❌ swaps** | ✅ |

That last one isn't a quota event. Today it would be silently failed over to another provider — burning a second provider's tokens and hiding a request-level problem the operator has to fix, since the turn then looks like a quota event rather than the bad request it is.

Both documented zai messages name **the caller's own** package, so anchoring on `your` costs nothing:

```diff
-  /(?:package|plan|subscription) has expired/i,
+  /\byour\b[^.]{0,60}\b(?:package|plan|subscription) has expired/i,
```

The `[^.]{0,60}` keeps it inside one sentence so it can't reach across into an unrelated clause. With it applied, `model-failover` + `error-translation` + `model-failover-router` + `rate-limit-quota` run 120/120, including your two new cases unchanged.

I pushed it to `fix/718-narrow-package-expired` if you'd like to cherry-pick rather than retype — it's a single commit on top of your branch. Happy for you to land it your own way.

Two non-blocking notes:

- `daily` in the limit-exhausted pattern isn't in the documented 1310 wording. Harmless — I swept for false positives and found none — just speculative.
- Because 1309/1314 carry no reset stamp, a permanently lapsed package pays one extra upstream round trip every turn, forever. That's consistent with the module's existing "don't invent a window" stance, but it does mean the operator never learns the subscription lapsed. Might be worth a one-time log.

Otherwise this is good to go. I also ran it inside a combined integration branch alongside #708, #710, #712–#715, #717, #720–#723: `routing` 137/137 and 379/379 across the surrounding suites, no interaction.

duolahypercho:
Reviewed — **merge**. I went looking specifically for over-broad matching, since a false positive here silently moves a turn to another model.

- Both patterns carry `/i`, matching the rest of `QUOTA_PATTERNS`. The observed body is Title Case, so the flag is required and present.
- They run against `extractUpstreamDetail(bodyText)` — the wrapper-stripped innermost upstream message, truncated to 300 chars — and only for `status < 500`. Never against request text, model output, or 5xx bodies.
- `isPlanEntitlement` is still checked first, and the zai package text matches none of the `ENTITLEMENT_PATTERNS`, so there is no new overlap with the must-not-swap branch.
- Cost of a false positive is bounded: with no `Retry-After` and no named reset, `classifyRoutedFailure` returns `{swap: true, reason: "out_of_usage"}` with no `until`, so no cooldown is written and the next turn retries the operator's chosen model. Failover is opt-in besides.

Two notes, neither blocking:

- `/(?:package|plan|subscription) has expired/i` could in principle match a provider echoing the caller's own text back in a 4xx. I could not find a provider on these routes that does, and the pattern is no broader than `/(?:no|any|out of) credits/i` already in the same list — flagging it only because that is the direction this list can drift.
- The comment cites zai's "Weekly/Monthly Limit Exhausted" but the test asserts on "Weekly Limit Exhausted". The regex does still match the slash form via its `monthly` alternative (`Weekly/`**`Monthly Limit Exhausted`**) — the `weekly` alternative alone would not, since `/` is not `\s`. Worth a second fixture rather than a code change.

Both tests fail if the patterns are reverted.

PR #720, opened 2026-09-12
fix(gemini): safely repair dangling nested tool definitions

gowrath:
The direct Gemini API rejected tool-enabled requests because Zillow's connector schema points `bedrooms` at `#/$defs/MinMaxInt`, while `MinMaxInt` is stored under `properties.request.$defs`. This makes the reference undefined at the schema root.

This change repairs missing direct `$defs` names only on the `gemini-api` Chat Completions route. It indexes original definition scopes before inlining, so a borrowed definition cannot acquire a different meaning at its destination. Valid root references (including boolean schemas) and ordinary property pointers remain intact. Only annotation siblings can be merged; validation siblings are left unchanged. Resource boundaries, cycles, and exhausted traversal/expansion/byte budgets return the original schema.

Evidence:

- Local router logs from the reported failure contain repeated upstream HTTP 400 responses for `gemini-api-models-gemini-3-6-flash`: `reference to undefined schema at properties.request.properties.propertyFiltersRequest.properties.bedrooms`, with `INVALID_ARGUMENT` status.
- Inspection of the cached `_zillow_property_search` input schema confirms the misplaced definition. The checked-in fixture is a reduced reproduction of that structure, preserving the actual range types, formats, and descriptions.
- Independently evaluating the full cached schema with Python `jsonschema`'s `Draft202012Validator` raises an unresolved-reference error for `{"request":{"propertyFiltersRequest":{"bedrooms":{"min":1,"max":2}}}}`. The same sample validates with zero errors after repair; the input schema is not mutated.
- [google-gemini/gemini-cli#13326](https://github.com/google-gemini/gemini-cli/issues/13326) is a historical report of the same error class on a different Gemini surface. It corroborates the error class, not this exact route or current API behavior.

Regression coverage exercises the observed malformed schema, valid boolean definitions, embedded resources, interacting `additionalProperties` constraints, shadowed definitions, pointer escaping, literal payloads, and fallback bounds. The forwarding test models the recorded 400, proves the unmodified fixture is rejected, and checks that valid schemas remain unchanged on the wire. Kimi and an OpenRouter-hosted Gemini model retain their original schemas.

Validation:

- `npm run check` passed.
- `node --test --test-timeout=60000 test/routing.test.mjs test/tool-schema-root.test.mjs` passed (191 tests).
- All four standalone adversarial reproductions pass, with independent schema validation for the three valid-schema cases.
- `git diff --check` passed.

The full repository suite and a fresh live Gemini request were not run. The logs establish the historical upstream failure; the local validator and mock demonstrate the repair and forwarding behavior, not current Google API acceptance. General property-reference rewriting and compatibility changes for other providers are outside this PR.

### DIFF
--- a/src/api-forwarder.mjs
+++ b/src/api-forwarder.mjs
@@ -77,7 +77,6 @@ import {
 } from "./provider-api-key-pool.mjs";
 import {
   inlineDanglingNestedDefsRefs,
-  inlineForeignRefs,
   nonRecursiveToolSchema,
   stripCodexEncryptedSchemaAnnotation,
 } from "./tool-schema-root.mjs";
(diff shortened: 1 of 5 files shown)


duolahypercho:
Reviewed line by line — this is a **merge** from me, and the conflict is trivial to resolve.

**On the conflict:** it is purely textual. Both this PR and the already-merged `580f2b3f` ("declare implied tool-schema types for Moonshot routes") insert a name into the same `import { … }` block at the same position. I checked whether main's new `declareSchemaTypes` changes anything this relies on — it does not: that path runs behind `needsMoonshotSchemaCompatibility(route)` while this repair is gated on `provider.id === "gemini-api"`, so the two never see the same tool payload. **Resolution is: keep both import names.** No behaviour needs re-verifying.

On the things I went looking for:

- **Cyclic `$ref` fails closed.** A definition that points back at itself (directly or through another borrowed one) is caught by `activeTargets.has(target)` on object identity, which sets `exceeded` and returns the *entire* schema unchanged rather than a half-repaired branch. Good choice.
- **Nothing is silently dropped.** Every failure mode — target not a plain object, disallowed sibling keys, cycle, budget exceeded — falls through to returning the original `$ref` node untouched.
- **No caller mutation.** Copy-on-write throughout, shallow `{...}` only on the changed key, matching the existing `inlineForeignRefs` pattern in the same file.
- **Bounds are real.** Reuses `MAX_INLINE_DEPTH`/`MAX_INLINE_EXPANSIONS`/`MAX_INLINE_BYTES`, plus a new 8192-node cap during indexing and byte-charging before each expansion. Refs outside the exact local pattern (external URIs, `#/properties/...`, deeper `$defs` paths, anchors) never enter the repair branch at all.

The `test/routing.test.mjs` coverage is the part I liked most — it runs the actual forwarder, confirms the unrepaired request really gets a 400 from a mock modelling the observed rejection, and confirms a Gemini-*named* model routed through OpenRouter is left alone. That last case is what proves the gate isn't over-broad.

PR #722, opened 2026-09-12
fix: gate leaked tool-call recovery to Hy4 and make its capture linear

duolahypercho:
Review fixes for #715. Targets that PR's branch so the diff here is just the fix.

The `chat-reasoning.mjs` half of #715 is clean and needs nothing. These three are all in the recovery transform.

### 1. The stage was installed for every routed provider
`const leakedToolCalls = route ? … : undefined` gates only on *being routed*, so GLM, DeepSeek, Kimi and Grok all had their text scanned for Hy4's markup.

Measured on the visible answer channel before the gate — a sentence that merely *describes* the syntax, of the kind a model produces when it echoes a diff, a web page, or this repository's own source:

```
input : "Here is the syntax the router recovers: <tool_calls:…>…exec_command…rm -rf /tmp/demo…</tool_calls:…> - that is what it looks like."
output: 3 injected function_call items, carrying exec_command and the rm -rf argument
visible text left behind: "Here is the syntax the router recovers:  - that is what it looks like."
```

That is a prompt-injection channel rather than a repair, and the markup is quoted verbatim in `src/leaked-tool-call-recovery.mjs` and throughout its tests — so an agent working on this repo can trigger it by quoting the source.

`usesLeakedToolCallRecovery` pins the stage to the six shipped `hy4-preview` routes, including the Command Code route that ships no `requestProfile`, and rejects a spoofed suffix (`evil-hy4-preview-x`).

### 2. The capture was re-scanned per delta
`_transform` is synchronous and every `indexOf` re-flattens the concatenation rope, so an unterminated span cost quadratic time on the router's event loop while every concurrent request waited:

| captured | before | after |
|---|---|---|
| 156 KiB | 838 ms | 85 ms |
| 313 KiB | 872 ms | 215 ms |
| 625 KiB | 9 411 ms | 442 ms |
| 1.25 MB | **29 539 ms** | **579 ms** |
| control (same bytes, no span) | 1 138 ms | 656 ms |

The capture bound is 4 MiB, so this got considerably worse than the table. It is now held unjoined and each delta is scanned once with a closing-tag overlap, so a tag straddling a delta boundary is still found — the after column now tracks the no-span control.

### 3. One item's calls were recovered twice
The summary and content channels shared a span stream, so the second channel's reading looked like a genuine extension of the first and the prefix dedupe passed it through — **executing a duplicate call**:

```
A. same span on summary+content deltas of ONE item  -> recovered 2  (expected 1)
B. stored item with span in BOTH summary[] and content[] -> recovered 2  (expected 1)
```

Separate streams per channel, and a stored item carrying the span in both takes the fuller reading rather than concatenating them. Both cases now recover exactly once. This also makes AGENTS.md invariant 3 ("one item contributes its calls once") true, which it was not.

### Also
The leaked-channel routing test documented a capture from `opencode-go hy4-preview` while routing it through `opencode-go/deepseek-v4.1-flash`. It now uses the route it documents.

## Verification
Both behavioural fixes carry a regression test verified to fail without the corresponding source change.

I deliberately did **not** ship a wall-clock test for the throughput fix. The first version compared an unterminated span against the same bytes with no span open, and it still passed against the quadratic implementation — on a loaded machine the control run had itself grown to 23 s, so the ratio measured load rather than the algorithm. The bound behaviour and the split-closing-tag overlap are asserted instead, and the measurement lives in AGENTS.md.

`npm run check` clean. 69 pass / 1 skip across `leaked-tool-call-recovery`, `chat-reasoning`, `namespace-relay-routing`, `reasoning-tag-stripper`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2142,16 +2142,30 @@ content and no `function_call`. Codex ends the turn there and writes
    tag must repeat it. Do not hardcode it.
 3. **One item contributes its calls once.** The same span arrives on the delta
    channel, in the `.done` snapshot, and in the stored item; recovery is keyed
-   by output index so the call is emitted a single time.
+   by output index so the call is emitted a single time. The summary and content
+   channels are two renderings of one item's thinking, so they hold separate
+   span streams and the fuller reading wins -- sharing one stream between them
+   made the second channel look like a genuine extension of the first and
+   recovered, and executed, the call twice.
 4. **A leaked argument value is text.** It is read as JSON only when its text is
    exactly its own JSON form, so a declared `20000` or `true` survives while a
    shell command, a path, or `0755` stays the string the model wrote.
-5. **Routed streams only, before the namespace transform**, so a recovered
+5. **Hy4 routes only, before the namespace transform**, so a recovered
    flattened `mcp__server__tool` call is restored like any other, the empty-
    completion guard sees content, and the phase labeller reads the blank
    message as commentary instead of a final answer. Native streams gain no
-   stage. Coverage lives in `test/leaked-tool-call-recovery.test.mjs` and the
-   leaked-channel case in `test/namespace-relay-routing.test.mjs`.
+   stage, and neither does any other routed family: this is Hy4's own syntax,
+   and scanning every routed provider's text for it would turn prose that
+   merely *quotes* the markup -- a diff, a web page, this file -- into executed
+   tool calls. `usesLeakedToolCallRecovery` is the gate; widening it past
+   `hy4-preview` reopens that injection channel. Coverage lives in
+   `test/leaked-tool-call-recovery.test.mjs` and the leaked-channel case in
+   `test/namespace-relay-routing.test.mjs`.
+6. **A span is scanned once, not re-scanned per delta.** The capture is held
+   unjoined with a closing-tag overlap because `_transform` is synchronous:
+   re-scanning one growing string re-flattens the rope every delta, and a
+   1.25 MB unterminated span blocked the event loop for 21.5 s against 0.8 s
+   for the same bytes with no span open. The capture bound is 4 MiB.
 
 ## Routed subagent regression prevention
 
--- a/src/leaked-tool-call-recovery.mjs
+++ b/src/leaked-tool-call-recovery.mjs
@@ -33,6 +33,18 @@ import { Transform } from "node:stream";
 // dropped rather than relayed as half a tag. The item's own text is cleaned
 // independently, so the transcript still carries what the model wrote.
 
+// Hy4 Preview is the only family that emits this syntax, so it is the only one
+// whose text is reinterpreted. Scanning every routed provider would make any
+// prose that merely *quotes* the markup -- a diff, a web page, this repository's
+// own source -- into executed tool calls, which is a prompt-injection channel
+// rather than a repair. `upstreamModel` carries it on every shipped route
+// (`hy4-preview` on opencode Go, `tencent/hy4-preview` elsewhere), including
+// the Command Code route that ships no `requestProfile`.
+export function usesLeakedToolCallRecovery(route) {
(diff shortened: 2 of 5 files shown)


duolahypercho:
Reviewed against #715's version of the transform, with the linearity claim measured rather than taken on trust.

## Verified

The rewrite is real. Both versions driven through the actual transform with 64-byte reasoning deltas fed in 16 KiB socket-sized chunks — milliseconds of blocked event loop for an unterminated span:

| span | #715 | this PR |
|---|---|---|
| 0.5 MB | 495 | 16 |
| 1 MB | 1508 | 28 |
| 2 MB | 5177 | 51 |
| 4 MB | 18312 | 94 |

At the 4 MiB bound that is 18.3 s of blocked loop turned into 0.09 s. The bookkeeping holds: `#unsearchedBase + #unsearched.length === #capturedLen` across the push, the not-found trim and the reset; the retained overlap is exactly `closeTag.length - 1`; and the frozen 45-byte head is exactly the longest legal opening tag, so the cached decision cannot change.

The double-recovery fix is proven too — running this PR's tests against #715's implementation, 23 pass and 2 fail, including `one item's calls are recovered once even when both channels carry the span`.

## Fixed in d0f4fec3

One superlinear path survived the linearity work: `parseLeakedToolCalls` ended with `cleaned.replace(/\s+$/, "")`. `MAX_CAPTURE_BYTES` does not bound it — that cap is on the streaming capture, while the `.done` snapshot and stored-item paths hand this whole-provider text. Measured on this branch, span + N spaces + one non-space char:

```
 50 KB     945 ms
100 KB    3762 ms
200 KB   15356 ms
400 KB   ~55000 ms
```

Replaced with `trimEnd`, which strips exactly the same set (verified identical to `/\s+$/` across all 65536 BMP code points) in one linear pass. Same inputs now measure 0.0–0.5 ms including a 2 MB run.

The regression test asserts an absolute 2 s ceiling rather than a ratio — the two forms differ by ~6 orders of magnitude, so the margin is ~20000x over the fix while still failing anywhere. A `{ timeout }` option cannot work: the blocking is synchronous, so the runner's timer never fires and the pre-fix run *passes* at 62 s.

## Two things worth saying out loud

**The nonce is a delimiter, not a secret.** Nothing in `src/` generates, injects or stores one — `OPEN_RE` captures it from the model's *own* opening tag and the closing tag is built from that capture. Open/close agreement is therefore free to anything that emits both. AGENTS.md's "The nonce is read, never assumed" is right as a parsing rule but reads like a safety property; worth one sentence saying otherwise.

So this gate *narrows* the channel rather than closing it. On an Hy4 route the only check on a recovered call is `NAME_RE`, a format check. The request's declared `tools` array is in scope in `handleResponses` and never passed to the transform — threading it through and dropping any recovered call whose name was not declared costs nothing for the genuine case and kills "quoted documentation becomes a call" for every tool the session did not attach.

**Nothing asserts the gate is consulted.** The router-level test was converted from `deepseek-v4.1-flash` to `hy4-preview` rather than duplicated, so if a later change reinstalled the transform unconditionally — exactly the #715 state — the whole suite still passes. One `scenario()` call on a non-Hy4 model asserting no function calls and that `tool_calls:` reaches the client would close it.

Gate scope itself is correct: enumerating all 141 `upstreamModel` values in `config/**/*.json`, only `hy4-preview` and `tencent/hy4-preview` match, covering all six Hy4 routes and nothing else. Note it keys on the model id alone, never the provider, and is case-sensitive — a catalog spelling `Tencent/HY4-Preview` would silently turn recovery off.

duolahypercho:
Closed by GitHub when #715's head branch (this PR's base) was deleted on merge — not by choice. Both commits here (7a7925cd gate + linear capture, a62185b1 deterministic bound test) were carried by #727's branch and landed on main in c1f6e69e (merge of #727), so nothing from this PR was lost.

PR #727, opened 2026-09-12
fix: replay reasoning natively for thinking models behind Chat Completions resellers

duolahypercho:
Stacked on #715 → #722 (base branch `fix/715-audit`), so this diff is one commit: the reasoning-replay contract, keyed on the upstream model family instead of request profiles.

## Why

The contract in `src/chat-reasoning.mjs` decided which routes get their reasoning replayed as `reasoning_content` (via `thinking` parts the forwarder restores) versus as visible assistant text. It was keyed on request profiles — `glm-thinking`, `deepseek-thinking`, `hy4-reasoning` — which only the vendor-direct routes carry. So `zai-coding/glm-5.3` replayed correctly while the same GLM on opencode Go, OpenRouter, or Command Code did not; the same for Kimi K3, Hy3, MiniMax M3, and DeepSeek behind resellers. Visible-text replay is the shape that made Hy4 loop on its own progress notes (#715), and #703 found the DeepSeek form of it: HTTP 400 on the following turn.

The rule belongs to the model, not to the reseller or its sampling profile.

## What

`nativeReasoningFamily(model)` matches `upstreamModel` — with or without a vendor prefix (`glm-5.3`, `z-ai/glm-5.3`, `zai-org/GLM-5.3`) — against a family table, for a listed set of Chat Completions providers. Each family names its evidence:

| family | match | evidence |
|---|---|---|
| DeepSeek | `(^|/)deepseek` | API 400s without it (#703's finding) |
| GLM-5.x | `(^|/)glm-5` | Z.ai requires history replayed with preserved thinking |
| Kimi K3 | `(^|/)kimi-k3$` | Moonshot: required in multi-turn and tool-call loops |
| MiniMax M3 | `(^|/)minimax-m3$` | interleaved thinking; `minimax-m3` profile already splits it out |
| Tencent Hy3/Hy4 | `(^|/)hy(3\|4)(-[a-z]+)*$` | hy4 measured in #715; hy3 same family, probed below |

Providers: `opencode-go`, `openrouter`, `commandcode` (verified live), `minimax-token-plan`, `kimi-api`, `kimi-api-cn`, `zai-api`, `deepseek`. Explicitly out: `kimi-k2.6` (does not preserve thinking), `kimi-k2.7-code` (no probed route), every `*-messages` provider (Anthropic protocol carries thinking blocks), and resellers not probed (Nous, Venice, Qwen Plan, Ollama Cloud, NanoGPT) — they keep their current channel until someone runs the probe against them.

This subsumes #703's `src/chat-reasoning.mjs` change (opencode-go DeepSeek) — credit there for finding the DeepSeek 400. Happy to rebase onto #703 instead if that should land first; only this file and the scoping test overlap.

## Verification

**Live, 12 September 2026, through the installed router on this branch:**

- P1, single turn with `effort: high` on 21 routes: every reachable route on opencode Go, OpenRouter, and Command Code (GLM 5.2/5.3/5.3-flash, Kimi K3, Hy3, DeepSeek V4.1) returned a reasoning item. Unreachable: MiniMax (429 out of usage), Nous (404), Venice (402), OpenRouter DeepSeek (401).
- P2, two turns with a prior reasoning item in history (so the forwarder restores `reasoning_content`) plus `effort: high`, carrying a marker in the replayed reasoning: **17/17 routes 200**, reasoning present on turn 2, correct answer, marker never leaked into the visible answer. That includes `opencode-go/deepseek-v4-flash`, `-v4-pro`, and `-v4.1-flash` — the exact `reasoning_content` + `reasoning_effort` pair upstream issue anomalyco/opencode#48180 reports as a 400; through this router it does not reproduce.

**Tests:**

- `test/chat-reasoning.test.mjs` — scoping test rewritten (opencode-go DeepSeek and commandcode Hy4 flip to in; k2.6/k2.7, `*-messages`, unprobed resellers asserted out), plus a family table test with one positive per (family, provider) and prefix-containment negatives.
- `test/routing.test.mjs` — forwarder "restores … thinking as native `reasoning_content`" now also covers `opencode-go-glm-5-3`, `opencode-go-kimi-k3`, `commandcode-kimi-k3`.
- `test/namespace-relay-routing.test.mjs` — the routed carry asserted for `opencode-go/glm-5.3` and `opencode-go/kimi-k3` alongside hy4, with `opencode-go/kimi-k2.6` as the negative control keeping the old shape.
- `chat-reasoning` 2/2, forwarder rows 7/7, relay 33/33, full `routing` suite green, `scripts-check` passes.

## Not in this PR

- `opencode-go/deepseek-v4.1-flash` also stores no reasoning items at all in Codex rollouts (orphan `rs_` deltas) — that half is #708; this PR makes the replay correct whenever items exist.
- MiniMax M3 and the unprobed resellers were not verified live (quota); they are in or out of the table on documentation and on probe evidence respectively, and the AGENTS.md bullet says what a future addition needs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1971,8 +1971,14 @@ retry rules on the shared path.
   An interleaved-thinking model that instead sees its past reasoning replayed as
   visible assistant text moves new thinking into the answer channel and loops
   on its last progress note (Hy4 on opencode Go, 12 September 2026: 2, 4, 5,
-  8, 16 copies per message). Add a thinking model to this contract in
-  `src/chat-reasoning.mjs`, not by special-casing the carry. Remove only successfully carried
+  8, 16 copies per message). The rule belongs to the upstream model, not to
+  the reseller or its request profile, so `src/chat-reasoning.mjs` also keys
+  it on the upstream family (DeepSeek, GLM-5.x, Kimi K3, MiniMax M3, Tencent
+  Hy3/Hy4) for the Chat Completions resellers it lists. Add a family only with
+  evidence the vendor expects `reasoning_content` back, and a reseller only
+  after a live probe shows the route returns reasoning and accepts the
+  echo-back; Anthropic-protocol variants never enter it. Do not special-case
+  the carry instead. Remove only successfully carried
   reasoning runs so plaintext cannot also become a user message. Do not mutate
   source items or change other native Responses routes. Keep this policy shared
   between hops without applying direct DeepSeek sampling parameters to resellers.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2142,16 +2142,30 @@ content and no `function_call`. Codex ends the turn there and writes
    tag must repeat it. Do not hardcode it.
 3. **One item contributes its calls once.** The same span arrives on the delta
    channel, in the `.done` snapshot, and in the stored item; recovery is keyed
-   by output index so the call is emitted a single time.
+   by output index so the call is emitted a single time. The summary and content
+   channels are two renderings of one item's thinking, so they hold separate
+   span streams and the fuller reading wins -- sharing one stream between them
+   made the second channel look like a genuine extension of the first and
+   recovered, and executed, the call twice.
 4. **A leaked argument value is text.** It is read as JSON only when its text is
    exactly its own JSON form, so a declared `20000` or `true` survives while a
    shell command, a path, or `0755` stays the string the model wrote.
-5. **Routed streams only, before the namespace transform**, so a recovered
+5. **Hy4 routes only, before the namespace transform**, so a recovered
    flattened `mcp__server__tool` call is restored like any other, the empty-
    completion guard sees content, and the phase labeller reads the blank
    message as commentary instead of a final answer. Native streams gain no
-   stage. Coverage lives in `test/leaked-tool-call-recovery.test.mjs` and the
-   leaked-channel case in `test/namespace-relay-routing.test.mjs`.
+   stage, and neither does any other routed family: this is Hy4's own syntax,
+   and scanning every routed provider's text for it would turn prose that
+   merely *quotes* the markup -- a diff, a web page, this file -- into executed
+   tool calls. `usesLeakedToolCallRecovery` is the gate; widening it past
+   `hy4-preview` reopens that injection channel. Coverage lives in
+   `test/leaked-tool-call-recovery.test.mjs` and the leaked-channel case in
+   `test/namespace-relay-routing.test.mjs`.
+6. **A span is scanned once, not re-scanned per delta.** The capture is held
+   unjoined with a closing-tag overlap because `_transform` is synchronous:
+   re-scanning one growing string re-flattens the rope every delta, and a
+   1.25 MB unterminated span blocked the event loop for 21.5 s against 0.8 s
+   for the same bytes with no span open. The capture bound is 4 MiB.
 
 ## Routed subagent regression prevention
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+- **Thinking models behind resellers now get their reasoning replayed the way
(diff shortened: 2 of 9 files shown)


duolahypercho:
Reviewed this against #703 and #715 by executing all three predicates over every `(provider, upstreamModel, requestProfile)` triple in `config/**/*.json` — 190 shipped routes — rather than reasoning about the regexes.

## The supersession question, settled

**#703 is a strict subset.** The "handled by #703, missed by this PR" set is **empty**. #703 gates `opencode-go` with `/(^|\/)deepseek/i`; this PR uses the byte-identical regex with `opencode-go` in its provider set. Beyond that it turns on ~15 routes neither other PR reaches — GLM on opencode-go and openrouter, Kimi K3 on three providers, MiniMax M3, Hy3 — all currently on the wrong replay channel.

**#715's `hy4-reasoning` line is not redundant, and this PR is right to keep it.** Two shipped routes reach the contract only through the profile check:

```
nano-gpt     | tencent/hy4-preview | hy4-reasoning
nousresearch | tencent/hy4-preview | hy4-reasoning
```

Neither provider is in `NATIVE_REASONING_CHAT_PROVIDERS`, so `nativeReasoningFamily` returns `undefined` for both. Anyone later "simplifying" that line away on the theory that the hunyuan family covers it would silently regress both to the looping behaviour #715 documents. Worth a comment on the line saying so.

## Fixed in 738256e4

`deepseek/deepseek-chat` was the one false positive in the whole catalog. It is the legacy non-thinking alias — provider `deepseek`, upstream `deepseek-chat`, profile `deepseek-nonthinking` — and it sat in the provider set while matching the unanchored prefix. That route sends `thinking: {type: "disabled"}`, so it would have gone out saying thinking is off while replaying `reasoning_content`: the combination the #703 vendor error is about. It also contradicts the rule this PR adds to AGENTS.md, that a family enters only with evidence the vendor expects `reasoning_content` back.

Two changes, because the id shape and the semantics are different guarantees:

- the family regex is anchored to the thinking lines, `/(^|\/)deepseek-(v\d|reasoner|flash)/i` — still covers `deepseek-reasoner`, `deepseek-flash` and every `deepseek-v<n>` on every listed reseller, and keeps a future non-reasoning id (`deepseek-ocr`, `-coder`, `-vl`) from entering without a probe
- `usesNativeChatReasoning` returns false for a `*-nonthinking` profile up front, which also covers discovered and user-added models no catalog review has seen

Measured over the shipped registry: **1 of 199 routes changes**, and it is `deepseek/deepseek-chat`.

The negative test that should have caught this **passed vacuously** — it omitted `upstreamModel`, so `String(undefined ?? "")` matched no family and the assertion held either way. It now carries the real id, and I ported #703's registry sweep alongside it: every shipped route that replays reasoning must state its evidence (thinking profile, family, or the legacy Command Code case), and no `*-nonthinking` route may be in the contract. Both fail against the pre-fix source.

## Remaining notes

- **Anchoring is still asymmetric.** `glm` is an unanchored prefix while `kimi-k3`, `minimax-m3` and `hunyuan` are `$`-anchored. Under-inclusive for dated or variant ids — `kimi-k3-0905`, `MiniMax-M3-Pro`, `hy4-preview-0912` (digits excluded by `(-[a-z]+)*`), `tencent/hy3:free` — the last of which exists today on `nousresearch`, currently moot only because that provider is unlisted.
- **This reverses #715's deliberate exclusion of `commandcode/tencent/hy4-preview`.** #715 asserted it false ("Command Code's own shim ... do[es] not"); the hunyuan family makes it true and the negative was deleted. The comment claims a probe — worth one sentence recording it, since it flips a stated decision.
- **Merge-order trap:** #703's registry test asserts `false` for `opencode-go/{hy4-preview,glm-*,kimi-k3,hy3}`, which this PR makes `true`. Git merges cleanly and the suite goes red. If #703's chat-reasoning commit is closed as subsumed, this disappears.

Note this branch does not yet carry #722's linear trailing-trim fix — it arrives when the stack merges bottom-up.

Issue #575, opened 2026-09-02
Provider circuit-breaker subsystem ships unreachable, and its telemetry key disagrees with the failover cooldown identity


c8dhjp4tyv-bit:
Two related findings from reading the cooldown paths while preparing #574. Neither is a user-visible failure today, which is why this is a question rather than a PR: the answer is a design decision, and the second finding only becomes a real bug once the first is resolved in one particular direction.

Everything below is against `main` at `a68418f`.

## 1. `src/provider-cooldown.mjs`'s circuit breaker and `src/subagent-availability-sync.mjs` are unreachable

`src/subagent-availability-sync.mjs` is the only module in the repository that nothing references. Searching every tracked file (not just `src/` and `test/`) for the name returns its own internal recursion and nothing else, so `startSubagentAvailabilitySync()` is never called:

```
$ grep -rn "subagent-availability-sync" . | grep -v '^\./\.git/'
(no matches outside the module itself)
```

The subsystem it drives is inert end to end:

| export in `provider-cooldown.mjs` | referenced from `src/` |
| --- | --- |
| `recordProviderCooldown` | — (the identically named live function is `model-failover.mjs`'s) |
| `activeProviderCooldown` | — |
| `cooldownErrorPayload` | — (not referenced by its own test either) |
| `temporarilyUnavailableSubagentSlugs` | — |
| `quotaResetAt` | — |
| `cooldownScope` | — |
| `nextQuotaCooldownExpiry` | `subagent-availability-sync.mjs` only |

And the one production link is closed at the other end too. `nextQuotaCooldownExpiry` selects snapshots by `snapshot?.reason !== "quota"`, but the only production writer of `rate-limits.json` is `api-forwarder.mjs:1108`, which stores `parseRateLimitHeaders()` output — a document that never carries `reason`. The sole writer of `reason: "quota"` is `provider-cooldown.mjs`'s own `recordProviderCooldown`, which nothing calls. So the predicate cannot be true on a real install regardless of the missing wiring.

[...]

PR #728, opened 2026-09-12
fix: key harvested quota headers by cooldown scope

c8dhjp4tyv-bit:
## Why

Follow-up to the second finding in #575, which you asked for as its own change:

> The `cooldownScope` extraction suggests the direction is already "keep the identity, drop the second implementation". If that's right, the `api-forwarder.mjs:1228` one-liner is worth taking on its own so the two stores can't drift back apart — please do send it.

`recordUpstreamLimits` writes to two stores from one provider id, under two different keys:

```js
if (rateLimit) recordRateLimitSnapshot(canonicalProviderId(normalized.provider.id), rateLimit);
...
recordProviderCooldown(normalized.provider.id, { until, reason });   // keys by cooldownScope
```

`opencode-zen` is the one id where those two functions disagree — it is `variantOf: opencode-go`, so `canonicalProviderId` folds it into Go, while `cooldownScope` deliberately keeps it because Zen bills at its own `/zen` endpoint.

Reproduced offline on current `main`, no credentials and no provider request, by calling the two modules the way the forwarder does — the Go plan reports 900 requests left, then a Zen response reports 12:

```
--- today: canonicalProviderId keying ---
rate-limits.json keys: [ 'opencode-go' ]
opencode-go  remaining: 12
opencode-zen remaining: undefined
cooldownScope('opencode-zen') = opencode-zen -> snapshot: MISS
```

Two separately billed plans, one key. Zen's number is sitting under the Go plan's name, and the Go plan's own window is gone — the next Go response overwrites it back, so whichever plan answered last wins and the entry is mislabelled either way. Zen's window cannot be read back either: every cooldown-scope consumer looks it up as `opencode-zen`, an id the file never held — `activeProviderCooldown` and `nextQuotaCooldownExpiry` key by `cooldownScope` (`provider-cooldown.mjs:53,75`).

To be precise about what this does *not* claim: the usage card is not affected today. `headerQuota` uses the raw id (`provider-account-usage.mjs:655`), but `providerAccountUsageSnapshot` filters variants out (`:1025`, `.filter((provider) => !provider.variantOf)`), so `opencode-zen` never reaches that path — one account entry per credential, as intended. Confirmed by executing it: with `providerIds: ["opencode-go", "opencode-zen"]` the returned snapshot has no `opencode-zen` key. The overwrite above is the effect that is live today; the unreadable-Zen half is the cooldown readers.

## What changed

`recordUpstreamLimits` keys the snapshot by `cooldownScope` — the same identity `recordProviderCooldown` two lines below already uses, so the two stores cannot drift apart again. The comment now records why the scope and the canonical id are not interchangeable here.

`opencode-zen` is the only id affected. Every other provider and variant — `commandcode-messages`, `opencode-go-messages`, `opencode-go-responses`, `opencode-free-responses` — has `cooldownScope === canonicalProviderId` and keeps the exact key it has today:

```
id                       variantOf      canonical      cooldownScope
commandcode-messages     commandcode    commandcode    commandcode
opencode-go-messages     opencode-go    opencode-go    opencode-go
opencode-go-responses    opencode-go    opencode-go    opencode-go
opencode-zen             opencode-go    opencode-go    opencode-zen   <-- DIVERGES
opencode-free-responses  opencode-free  opencode-free  opencode-free
```

A `rate-limits.json` already carrying a Zen-written entry under `opencode-go` needs no migration: it is last-write-wins telemetry with an `observedAt`, replaced by the next Go response that carries headers.

## Verification

- New `test/rate-limit-scope.test.mjs` spawns the real `src/api-forwarder.mjs` against a mock Zen upstream (a curated `opencode-zen` user model via `MODEL_ROUTER_USER_MODELS`, `OPENCODE_ZEN_BASE_URL` pointed at the mock), with `rate-limits.json` pre-seeded with the Go plan's window. It asserts Zen's headers land under `opencode-zen` and that the Go entry is untouched. It fails on unpatched `main` (`expected: 7`, the Zen figure never filed) and passes with the change.
- `node --test test/rate-limit-quota.test.mjs test/provider-cooldown.test.mjs test/provider-account-usage.test.mjs test/rate-limit-headers.test.mjs test/model-failover.test.mjs` — 102/102 pass.
- `node --test test/commandcode-forwarder.test.mjs` — 5/5 pass; its existing `limits.commandcode.requests.remaining` assertion pins the unchanged canonical-parent behaviour for a variant that shares its plan.
- `npm run check` passes.

Not touched: the first finding in #575 (the unreachable `subagent-availability-sync` / `provider-cooldown` circuit-breaker half). Wiring it or deleting it is your call, and this change is independent of it either way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,20 @@
   client meant to leave open is worse than the rejection. Every other provider
   keeps the exact wire payload it has today.
 
+- **opencode Zen's quota headers no longer overwrite the Go plan's.** Both
+  plans share one credential and one selection toggle, but Zen bills at its own
+  endpoint, which is why `cooldownScope` keeps Zen's identity where
+  `canonicalProviderId` folds it into Go. The forwarder harvested the observed
+  rate-limit headers under the canonical id instead, so every Zen response
+  overwrote the Go plan's entry in `rate-limits.json` and every Go response
+  overwrote Zen's -- one key holding whichever plan answered last, under the
+  name of the other. Zen's window could not be read back either: every
+  cooldown-scope consumer looks it up as `opencode-zen`, an id the file never
+  held. The snapshot is now keyed by cooldown scope, the same identity the
+  cooldown store beside it already uses, so the two cannot drift apart again
+  (#575). Every other provider and variant keeps the key it has today, and a
+  stale entry is replaced by the next response that carries headers.
+
 - **A routed model the router has not loaded now fails locally, not at
   ChatGPT.** A model added to or renamed in `user-models.json` shows up in the
   Codex picker as soon as the catalog is rebuilt, but the running router reads
--- a/src/api-forwarder.mjs
+++ b/src/api-forwarder.mjs
@@ -38,6 +38,7 @@ import {
 } from "./rate-limit-headers.mjs";
 import { recordRateLimitSnapshot } from "./rate-limit-state.mjs";
 import { recordProviderCooldown } from "./model-failover.mjs";
+import { cooldownScope } from "./provider-cooldown.mjs";
 import { canonicalProviderId, readProviderSelection } from "./provider-selection.mjs";
 import { stripImages, supportsImageInput } from "./vision-bridge.mjs";
 import {
@@ -1223,9 +1224,13 @@ async function upstreamSession(provider, credential, payload, options = {}, endp
 // never sit in time-to-first-byte.
 function recordUpstreamLimits(normalized, upstream) {
   const rateLimit = parseRateLimitHeaders(upstream.headers);
-  // Variant-routed responses meter the same upstream subscription, so quota
-  // headers land under the family's canonical provider id.
-  if (rateLimit) recordRateLimitSnapshot(canonicalProviderId(normalized.provider.id), rateLimit);
+  // Keyed by cooldown scope, the same identity `recordProviderCooldown` uses
+  // below. A protocol variant meters its parent's subscription and shares its
+  // key, but opencode Zen is billed at its own endpoint: canonicalizing here
+  // filed Zen's window under the Go plan, so each plan's response overwrote the
+  // other's snapshot and neither could be read back under the id that produced
+  // it.
+  if (rateLimit) recordRateLimitSnapshot(cooldownScope(normalized.provider.id), rateLimit);
   // This hop is the only place the provider's own status and headers are seen
   // before LiteLLM restates them, so it is the only place a reset time the
   // gateway does not relay can still be read. A failure that names when the
--- a/test/rate-limit-scope.test.mjs
+++ b/test/rate-limit-scope.test.mjs
@@ -1,0 +1,145 @@
+import assert from "node:assert/strict";
+import { spawn } from "node:child_process";
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import http from "node:http";
+import os from "node:os";
+import path from "node:path";
+import test from "node:test";
+import { fileURLToPath } from "node:url";
+
+import { userModelEntry } from "../src/user-models.mjs";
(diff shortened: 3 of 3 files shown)


duolahypercho:
Reviewed — **merge**. This fixes a real internal inconsistency rather than introducing a new identity, which is what I wanted to confirm before taking a one-line key change.

`cooldownScope` differs from `canonicalProviderId` for exactly one id: `opencode-zen`, the only provider declaring `variantOf: "opencode-go"` while billing at its own endpoint. I enumerated the opencode variants — `-messages` and `-responses` both fold to `opencode-go` under either function — so every other provider and variant keeps the key it has today.

The inconsistency: `recordProviderCooldown` eleven lines below in the same function already passes the provider id through unresolved *precisely* so the window is keyed by cooldown scope, and its comment says so; `src/model-failover.mjs` keys its whole store the same way. So before this, the snapshot store and the cooldown store beside it disagreed about `opencode-zen`, each plan's response overwrote the other's entry, and `rateLimitSnapshotFor("opencode-zen")` — the id every reader asks for — was always empty. This PR strictly *splits* a wrongly-shared bucket rather than merging two.

On the specifics I checked: key space is registry provider ids post-`variantOf` collapse, so two providers cannot collide unless one declares the other as its variant, which is the intended shared-subscription semantics. Nothing from the credential or pooled key id reaches `recordRateLimitSnapshot`. No unbounded growth — `MAX_PROVIDERS = 64` with `observedAt`-descending eviction, re-read from disk per write, no in-process map. Downstream readers already key by cooldown scope, and the usage card path can now actually find a Zen entry where it previously never could.

Leaving the three remaining `canonicalProviderId` call sites alone is correct — they are Command Code plan-vs-provider route selection, unrelated to quota buckets — and the import is still used.

The test is a real end-to-end proof: it seeds a Go-plan window, spawns the actual forwarder against a local mock bound through `OPENCODE_ZEN_BASE_URL`, and asserts both that Zen's window lands under `opencode-zen` *and* that Go's pre-existing entry survives at 900/1000. That second assertion is the overwrite being fixed.

One small note: the 2-second settle poll could in principle be tight on a loaded runner, and would fail on the `remaining === 7` assertion rather than with a clear timeout message. Not worth changing unless it flakes.

c8dhjp4tyv-bit:
Thanks. One correction to the record, since it is the sort of thing that becomes a false premise later.

> Downstream readers already key by cooldown scope, and the usage card path can now actually find a Zen entry where it previously never could.

The first half holds; the second does not, and this PR does not change it. `providerAccountUsageSnapshot` excludes variants before anything is queried:

```js
[...PROVIDERS.values()]
  .filter((provider) => !provider.variantOf)     // provider-account-usage.mjs:1025
```

`accountUsageFor` has exactly one caller, that map, so `headerQuota("opencode-zen")` is unreachable no matter what the snapshot store holds — one account entry per credential, which is what the comment there says it is for. Executed against the module to be sure: with `providerIds: ["opencode-go", "opencode-zen"]` the returned object has no `opencode-zen` key at all.

So what this PR fixes is the two effects that are live today — each plan's response overwriting the other's entry, and `rateLimitSnapshotFor("opencode-zen")` always being empty for the cooldown readers. A Zen usage card would additionally need that filter to treat a separately billed variant as its own account, which is a different change and not one I would make unasked. The PR body states this; I should have made it louder than a parenthetical.

On the settle poll: agreed, and left alone. It would only turn a timeout into a clearer message, and re-running a 22-minute Windows leg to improve an assertion's failure text is a bad trade while it is not flaking. Happy to send it as a follow-up if it ever does.

c8dhjp4tyv-bit:
Rebased onto `main` — #723 landed and all three of us add an entry at the top of `## Unreleased`, so the only conflict was there. Kept #723's entry where it is and placed mine below it; no code hunk was touched. `npm run check` and this PR's tests pass on the rebased tree, and it shows mergeable again.

PR #729, opened 2026-09-12
fix: keep the whitespace an untagged answer starts with

c8dhjp4tyv-bit:
## Why

`src/reasoning-tag-stripper.mjs` has two forms of the same operation, and they disagree.

`stripThinkTags` trims the message's leading whitespace only when it actually removed a tag, and returns an untagged message by identity — the existing test says so explicitly (`stripThinkTags("Paris")`, *"no tags -> unchanged (identity)"*). It is the form used for `response.output_text.done` and for the stored message item in `output_item.done`.

`ThinkStreamStripper`, the delta-channel form, trimmed unconditionally:

```js
#leadTrim(out) {
  if (this.#emittedVisible) return out;
  const trimmed = out.replace(/^\s+/, "");   // even when nothing was stripped
  ...
```

So an ordinary answer opening with a newline — a fenced code block, a leading blank line — reached the client without it, while the terminal snapshot beside it kept it. The rendered answer and the stored one disagreed on a route that had leaked no reasoning at all. When the first delta was *only* whitespace, `#rewrite` saw `cleaned.length === 0` and dropped the event outright.

The class documents the invariant this breaks:

> The concatenation of every `feed`/`flush` return equals `stripThinkTags` of the concatenated input.

I checked it by property test rather than by reading — 300,000 random inputs over a tag-adjacent alphabet (`<`, `>`, `/`, `think`, `reason`, whitespace, letters), each split randomly across deltas, comparing the streamed concatenation against `stripThinkTags` of the same text:

```
=== BEFORE (unpatched main) ===
cases                            : 300000
divergences                      : 91975
UNTAGGED divergences             : 19575     <-- ordinary answers, no tag anywhere
unexplained by {unterminated open tag, tag after visible text}: 19575
    {"text":"\n","streamed":"","whole":"\n"}
    {"text":" a","streamed":"a","whole":" a"}
    {"text":"\tx","streamed":"x","whole":"\tx"}
```

One case in fifteen was an ordinary, reasoning-free answer being altered.

## What changed

Hold the leading whitespace instead of trimming it. By the first visible character the stripper knows whether a tag was removed, so it can emit or drop that whitespace exactly as `stripThinkTags` would. `#removedTag` records the removal; `#pendingLead` holds the whitespace.

The observed leak shapes are unaffected — the reasoning block is what comes first, so the removal is always known in time:

```
"<think>The capital of France is Paris.</think>\nParis"  ->  "Paris"
"\n</think>\n\nThe real answer starts here"              ->  "The real answer starts here"
```

After the change, no untagged message diverges, and nothing diverges that is not one of the two cases a streaming stripper cannot fix:

```
=== AFTER (patched) ===
cases                            : 300000
divergences                      : 81963
UNTAGGED divergences             : 0
unexplained by {unterminated open tag, tag after visible text}: 0
```

Both residual cases are pre-existing and deliberate, and I left them alone:

- **an unterminated `<think>`** — `flush` discards a span that never closed, which its comment states as intended;
- **a tag that appears only after visible text** — `stripThinkTags` trims the message's start retroactively, and a stream that already sent that text cannot retract it. The class comment now records this as the one exception to the stated invariant, rather than leaving the invariant overstated.

## Verification

- `test/reasoning-tag-stripper.test.mjs` — two new tests, both failing on unpatched `main` and passing with the change:
  - a streamed answer opening with `\n` keeps it across chunk sizes 0/1/7, and the delta concatenation equals the `output_text.done` snapshot;
  - a property test over every 3-piece combination of a tag-adjacent alphabet, in four splits each: an untagged message with visible text streams byte-for-byte.
- A third new test pins the whitespace-only-delta case as **unchanged**, so the hold cannot silently grow into it.
- A fourth pins where the two mechanisms meet — leading whitespace still held when a `<think>` starts arriving one character at a time across deltas (`"\n"`, `"<th"`, `"ink>hidden</think>"`, `"\nAnswer"` -> `"Answer"`). It passes before this change as well, so it characterises the new `#pendingLead` path rather than proving it.
- All 10 pre-existing tests in the file still pass (13/13 total).
- `npm run check` passes.

## One thing I did not fix

`ThinkStreamStripper.flush()` computes a remainder its comment says "is emitted" — *"normal-mode remainder -- including a lone `<` that never became a tag -- is emitted"* — but the only caller, the `output_text.done` branch of `#rewrite`, calls it for its side effect and discards the return value. So a message ending mid-tag loses that tail from the delta channel. Identical before and after this change:

```
in="a <"            deltas="a "            stripThinkTags="a <"
in="x<th"           deltas="x"             stripThinkTags="x<th"
in="5 < 6 and 7 <"  deltas="5 < 6 and 7 "  stripThinkTags="5 < 6 and 7 <"
```

The `output_text.done` snapshot carries the full text, so nothing is lost to a client that reads it. Emitting the remainder would mean synthesizing a final delta event, which is a wire change I did not want to make unasked — happy to send it separately if you want it. The property test excludes those inputs explicitly rather than silently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,23 @@
   client meant to leave open is worse than the rejection. Every other provider
   keeps the exact wire payload it has today.
 
+- **A streamed answer no longer loses the whitespace it starts with.** The
+  inline-reasoning stripper trimmed the leading whitespace of every message it
+  streamed, but `stripThinkTags` -- the same module's whole-string form, used
+  for the `output_text.done` snapshot and the stored message item -- trims only
+  when it actually removed a tag, and returns an untagged message by identity.
+  So an ordinary answer opening with a newline (a fenced code block, a leading
+  blank line) reached the client without it, while the terminal snapshot beside
+  it kept it: the rendered answer and the stored one disagreed on a route that
+  had leaked no reasoning at all. A first delta made only of whitespace was
+  dropped from the stream outright. The stripper now holds that whitespace until
+  the first visible character, by which point it knows whether a tag was
+  removed, and emits or drops it to match. The hold is bounded: nothing is kept
+  once a tag has been removed, only the new delta is scanned rather than the
+  accumulation, and past 8 KiB of unbroken whitespace the stripper emits what it
+  holds instead of growing further. The observed leak shapes are unchanged --
+  reasoning arrives before the answer, so the removal is always known in time.
+
 - **A routed model the router has not loaded now fails locally, not at
   ChatGPT.** A model added to or renamed in `user-models.json` shows up in the
   Codex picker as soon as the catalog is rebuilt, but the running router reads
--- a/src/reasoning-tag-stripper.mjs
+++ b/src/reasoning-tag-stripper.mjs
@@ -30,6 +30,12 @@ const OPEN_TAGS = TAG_NAMES.map((name) => `<${name}>`);
 const CLOSE_TAGS = TAG_NAMES.map((name) => `</${name}>`);
 const ALL_TAGS = [...OPEN_TAGS, ...CLOSE_TAGS];
 const MAX_TAG_LEN = Math.max(...ALL_TAGS.map((tag) => tag.length));
+// Ceiling on the leading whitespace the streaming stripper will hold back while
+// it waits to learn whether a tag is coming. Every observed leak opens its
+// reasoning tag in the first delta, so a message that has produced this much
+// whitespace and nothing else is not a leak -- it is an answer whose own
+// indentation must be emitted rather than buffered without bound.
+const MAX_PENDING_LEAD = 8192;
 const NAMES_ALT = TAG_NAMES.join("|");
 // Open-to-nearest-close, any name to any name (non-greedy) -- matches the
 // streaming machine, which closes on the first close tag it sees.
@@ -68,11 +74,23 @@ export function stripThinkTags(text) {
 // Incremental stripper for the streamed delta channel. `feed` returns the text
 // safe to emit so far; `flush` returns whatever remains once the stream ends.
 // The concatenation of every `feed`/`flush` return equals `stripThinkTags` of
-// the concatenated input.
+// the concatenated input, with one unavoidable exception: `stripThinkTags`
+// drops the message's leading whitespace whenever it removed a tag *anywhere*,
+// including one that appears after visible text has already been streamed. A
+// stripper that has already emitted that text cannot retract it. Every case
+// where the removal is knowable by the time the first visible character is
+// emitted -- which is every observed leak, since the reasoning block is what
+// comes first -- does agree. A message that opens with more than
+// `MAX_PENDING_LEAD` bytes of unbroken whitespace settles the same way, for the
+// same reason. Both are display-only: `output_text.done` and `output_item.done`
+// bypass this class and clean the complete text through `stripThinkTags`, so
+// the stored and re-rendered message is correct regardless.
 class ThinkStreamStripper {
   #mode = "normal";
   #carry = "";
-  #emittedVisible = false;
+  #leadSettled = false;
+  #removedTag = false;
+  #pendingLead = "";
 
   // Longest suffix of `s` that is a proper prefix of any tag, so a tag split
   // across deltas is held back rather than emitted as literal text.
@@ -94,6 +112,7 @@ class ThinkStreamStripper {
         if (found) {
           out += this.#carry.slice(0, found.at);
           this.#carry = this.#carry.slice(found.at + found.tag.length);
+          this.#removedTag = true;
           // Opening a span enters think mode; an orphan close is just dropped.
           if (found.opening) this.#mode = "think";
           continue;
@@ -107,6 +126,7 @@ class ThinkStreamStripper {
       const close = firstTag(this.#carry, CLOSE_TAGS);
       if (close) {
         this.#carry = this.#carry.slice(close.at + close.tag.length);
+        this.#removedTag = true;
         this.#mode = "normal";
         continue;
       }
@@ -114,7 +134,7 @@ class ThinkStreamStripper {
       this.#carry = this.#carry.slice(this.#carry.length - hold);
       break;
     }
-    return this.#leadTrim(out);
+    return this.#lead(out);
   }
 
   flush() {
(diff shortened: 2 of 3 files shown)


duolahypercho:
Reviewed line by line. The bug is real and the fix is right in shape — one thing to tighten before I merge it.

Confirmed on `main`: `#leadTrim` strips leading whitespace from the first normal-mode text unconditionally, regardless of whether a tag was ever removed, which diverges from `stripThinkTags` (which only trims when something actually was). An ordinary answer that opens with a blank line loses it in the live delta stream with no reasoning tags present at all.

**Please bound `#pendingLead` before this lands.** While the accumulated text is still entirely whitespace, `#lead()` concatenates onto it and re-runs `.replace(/^\s+/, "")` over the whole growing string on every `output_text.delta` — O(n) per call, O(n²) over n such deltas, with no cap. Every other buffer in this class is bounded (`#carry` at `MAX_TAG_LEN - 1`); this is the one exception, and I could not find a byte cap on the transform in `router.mjs`. Tracking a length plus a truncation marker, rather than holding the full string, would do it.

Two notes, neither blocking:

- The documented divergence — a tag arriving after visible text has already streamed cannot retroactively un-emit whitespace — is fine in practice, and your comment is honest about it. Worth noting that `response.output_text.done` and `response.output_item.done` bypass the streaming stripper and clean the complete text through `stripThinkTags`, so the stored and rendered message is always correct; only the live-typing display can transiently show the stray whitespace. The class's top comment ("concatenation of every feed/flush equals stripThinkTags…") is no longer literally true, though.
- Two of the four new tests (`a delta that is only whitespace is still held back`, `held leading whitespace is dropped by a tag that only arrives in a later delta`) pass identically on old and new code — your own comments say so, and I confirmed by tracing. They're good pinning tests; just flagging that the two that genuinely fail pre-fix are the other pair. Nothing covers a long whitespace-only run, which is the case above.

Thanks — this is a subtle one and the diagnosis is exactly right.

c8dhjp4tyv-bit:
Bounded, and thank you — the quadratic path was real and I had not measured it.

Reproduced first, driving whitespace-only deltas of 64 characters through the transform on the previous commit:

```
  2000 deltas ->    160 ms
  8000 deltas ->   4088 ms
 16000 deltas ->  14398 ms      held = 1,024,000 bytes
```

Doubling the input roughly quadruples the time, and nothing caps the buffer.

## What changed

`#lead` no longer re-trims the accumulation. Three parts:

- **Nothing is held once a tag has been removed.** `#removedTag` already tells the stripper the message will be trimmed either way, so that branch discards the whitespace immediately and buffers nothing at all. This is the common case for an actual leak.
- **Only the new chunk is scanned.** `#pendingLead` is whitespace by construction, so `out.search(/\S/)` on the incoming delta is sufficient — the accumulation is never re-examined. That removes the O(n) per delta.
- **The accumulation gives up at `MAX_PENDING_LEAD` (8 KiB).** Past that the stripper emits what it holds and stops holding, which is the honest reading: every observed leak opens its tag in the first delta, so 8 KiB of unbroken whitespace is an answer's own indentation, not a preamble to a tag.

Same measurement after:

```
  2000 deltas ->     53 ms
  8000 deltas ->     81 ms
 16000 deltas ->    114 ms      held = 0 bytes
```

I renamed `#emittedVisible` to `#leadSettled`, since it now means "stop holding" rather than "a visible character went out" — the cap sets it without any visible character.

## Correctness is unchanged

The 300k-case differential fuzz against `stripThinkTags` returns exactly the numbers it did before the rewrite — same total, and still zero divergences on messages with no tag:

```
cases              : 300000
divergences        : 81963
UNTAGGED divergences (must be 0): 0
unexplained by {unterminated open tag, tag after visible text} (must be 0): 0
```

Settling at the cap gives up the *retraction*, not the stripping: a tag after the cap is still removed, only the leading whitespace stays. A 20,000-newline lead followed by `Answer` with no tag present comes out byte-identical, 20,006 in and 20,006 out.

## Tests

Two new ones, and you were right that the gap was the long run:

- `a long whitespace-only run is held in bounded time and loses nothing` — 16,000 whitespace deltas, asserts byte-identity and a 3s ceiling. I sized it from measurement rather than taste, because my first attempt at this test was worthless: at 313 deltas it passed on the quadratic code too. At 16,000 the bounded version takes ~115 ms (26x headroom) and the previous one takes 13.5 s (4.5x over), so neither a loaded runner nor a fast one can flip it. Confirmed failing against the previous commit: `whitespace-only run took 13547ms`.
- `whitespace held past the cap is still dropped by a tag that follows it` — pins that the cap path does not leak reasoning.

On your two non-blocking notes: the class comment now names the cap as a second settled case and records that `output_text.done` / `output_item.done` bypass this class entirely, so the stored and re-rendered message is correct either way and only live typing can show the stray whitespace. And agreed on the test pair that passes identically before and after — their comments say so, and the two that genuinely fail pre-fix are the other pair, now three with the run above.

`npm run check` and the 16-test file pass.

c8dhjp4tyv-bit:
Rebased onto `main` — #723 landed and all three of us add an entry at the top of `## Unreleased`, so the only conflict was there. Kept #723's entry where it is and placed mine below it; no code hunk was touched. `npm run check` and this PR's tests pass on the rebased tree, and it shows mergeable again.

duolahypercho:
## Correction — my earlier review was wrong, please disregard it

I asked you to bound `#pendingLead` before this lands. **It was already bounded, in the same commit I was reviewing.** I re-checked: this PR has exactly one commit (`20f1efaa`), and that commit introduces `MAX_PENDING_LEAD = 8192` at line 38 along with the enforcement in `#lead`. `main` has no such constant, so it came from you.

Both halves of what I raised were wrong:

- **Unbounded growth** — no. `if (this.#pendingLead.length + out.length > MAX_PENDING_LEAD)` emits what it holds and sets `#leadSettled = true`, so it stops holding entirely rather than merely capping.
- **O(n²) re-scanning** — no. `#lead` runs `out.search(/\S/)` against the *new chunk only* and never re-scans the accumulation.

Your comment on the method states exactly this ("only the new chunk is scanned -- `#pendingLead` is whitespace by construction, so re-trimming the accumulation would cost O(n) per delta and O(n^2) over a run of whitespace-only ones -- and the accumulation gives up at `MAX_PENDING_LEAD`"). You had anticipated the critique and written the answer into the code, and I asked you to go do it anyway. Sorry for the wasted round trip.

Everything else in the review stands and none of it is blocking: the two pinning tests that pass on old and new code are fine as pinning tests, and the documented post-visible-text divergence is genuinely bounded by `output_text.done`/`output_item.done` going through `stripThinkTags`.

No changes requested. Merging.

duolahypercho:
Merged `main` into this branch to clear a CHANGELOG conflict created by #728, #701 and #710 landing ahead of it (maintainer edits were enabled on the PR, so I pushed the merge commit directly — shout if you'd rather I hadn't).

Only `CHANGELOG.md` conflicted; both bullets kept, no duplicates introduced, and your `src/reasoning-tag-stripper.mjs` and its tests are untouched. `test/reasoning-tag-stripper.test.mjs` passes 16/16 on the merged tree.

Merging once CI comes back green.

c8dhjp4tyv-bit:
## Your review was right — the bound came from it, not before it

Please don't retract that one; the round trip wasn't wasted.

The commit you reviewed was `e45d895`, not `20f1efaa`. I amended and force-pushed in response to your review, so the commit you actually read stopped being the branch head and only the fixed version survived under a new SHA. That is what made it look like the bound had been there all along — my fault for amending mid-review rather than adding a follow-up commit.

`e45d895` is orphaned but still retrievable:

```
gh api "repos/c8dhjp4tyv-bit/codex-router/contents/src/reasoning-tag-stripper.mjs?ref=e45d895" --jq '.content' | base64 -d
```

It contains zero occurrences of `MAX_PENDING_LEAD`, and its `#lead` is:

```js
#lead(out) {
  if (this.#emittedVisible) return out;
  if (!out) return "";
  const held = this.#pendingLead + out;
  const trimmed = held.replace(/^\s+/, "");
  if (!trimmed) {
    this.#pendingLead = held;
    return "";
  }
  ...
}
```

Both halves of what you raised, exactly as you described them: the accumulation re-trimmed on every delta, and no ceiling on it.

It also measured the way you predicted. Whitespace-only deltas of 64 characters, driven through the real `ReasoningTagStripper`:

| deltas | `e45d895` | merged |
| --- | --- | --- |
| 2,000 | 160 ms | 53 ms |
| 8,000 | 4,088 ms | 81 ms |
| 16,000 | 14,398 ms | 114 ms |

1 MB held at the end of the last one, versus nothing.

The method comment you quoted back at me is one I wrote *while* implementing the fix you asked for — it describes the bound, it didn't precede it. I hadn't anticipated the critique.

One more correction while I'm at it: my first version of the complexity guard used 313 deltas and passed on the unbounded code too, so it guarded nothing. I caught it only by running it against the pre-fix tree. The merged one is sized from the table above — ~26x headroom on the bounded implementation, ~5x overshoot on the old one.

Thanks for the review, and for taking the CHANGELOG merge yourself.

PR #730, opened 2026-09-12
fix: locate the checkout through symlinks in every bin/ command

inchane22:
## What

Every command in `bin/` located the checkout with

    repo_dir=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)

`dirname $0` is only right when the script runs at its real path, and the logical `cd` is only right when no directory on the way is a symlink. `bin/codex-router` already walked the symlink chain, and its comment recorded that the siblings did not; this inlines the same walk into all 38 of them and resolves the root physically (`cd -P … && pwd -P`, the idiom `bin/model-router-tray:683` already uses).

Two distinct failures follow from the old form.

**A command linked onto PATH cannot find the checkout.** `ln -s ~/src/codex-router/bin/doctor ~/.local/bin/doctor` gives

    Error: Cannot find module '/Users/me/.local/src/doctor.mjs'

Relative links that climb out of a linked directory (GNU Stow's folded `~/bin`) are worse: the walk yields a path the kernel accepts, the logical `cd` strips `..` textually, and the command aborts under `set -e` on dash and macOS `/bin/sh`. Only bash invoked as `bash` retries physically and hides it.

**A root containing a linked directory makes commands silently do nothing.** 52 modules in `src/` guard main with `path.resolve(process.argv[1]) === fileURLToPath(import.meta.url)`. Node realpaths a main module but leaves `argv[1]` as written, so `node "$repo_dir/src/x.mjs"` with a logical `$repo_dir` fails that comparison: the module imports and the process exits 0, having done nothing.

## This also fixes Homebrew installs

The packaged install is the second case, and I did not expect this going in. The formula's wrapper execs `<prefix>/opt/<formula>/libexec/bin/codex-router`. That file is not a link — so the walk is a no-op there — but `opt/<formula>` is a link into the versioned keg, so `repo_dir` stayed on the `opt` path. Every command behind a guarded module (`agents`, `skills`, `caller-key`, `providers`, `update`, `refresh-catalog`, `support-bundle`, …) exited 0 without running. The commands that appeared to work — `--version`, `help`, `setup`, `doctor`, `control`, `panel`, `provider-key` — are the ones that never cross a guard, and the formula's `test do` block only runs those, so CI never saw it.

Reproduced in a keg-shaped fixture built from this repo's own files: on `99d526d`, `codex-router skills` exits 0 with no output; with this patch it prints usage and exits 2. `test/bin-symlink-resolution.test.mjs` now builds that layout by name so it stays covered.

I can only simulate the keg layout here rather than install the formula, so please sanity-check this against a real `brew install` before trusting the claim.

## Behaviour changes worth flagging

- `bin/install`'s five closing API-key hints printed `$repo_dir`, which is now the versioned keg path for a packaged reader — stale after the next `brew upgrade`. They now print `${CODEX_ROUTER_SOURCE_ROOT:-$repo_dir}`, so a packaged install names the stable `opt` path and a checkout is unchanged.
- `bin/model-router-tray` runs `$repo_dir/scripts/build-macos-tray-app.sh`, which seals its root into the tray bundle (`Contents/Resources/router-root` and Info.plist `ModelRouterSourceRoot`). For a checkout reached through a linked directory that sealed root is now physical rather than logical. Homebrew is not affected — the tray exits early when `CODEX_ROUTER_PACKAGE_MANAGER=homebrew`. Tell me if you would rather it kept the logical path.
- Otherwise `$repo_dir` is transient: it names the module to run, is `cd`'d into, and locates sibling commands. Node's own `SOURCE_ROOT` still comes from `CODEX_ROUTER_SOURCE_ROOT` when the wrapper exports it, so nothing persisted into launchd/systemd units changes.

## Why inline it 38 times

A sourced helper cannot be found before the path is resolved — `. helper` searches PATH, which is exactly what holds only the link. `readlink -f`/`realpath` need macOS ≥ 12.3; the dispatcher's existing walk depends on neither, and the one `readlink -f` in `bin/` (`model-router-tray:805`) sits in a Linux-only `/proc` branch behind a fallback. A node one-liner would put node on the critical path of `bin/install`, the script that decides where node comes from. So the siblings follow the dispatcher, and the static test pins the load-bearing lines in every copy so they cannot drift.

## Tests

`test/bin-symlink-resolution.test.mjs`:
- static: every `#!/bin/sh` command in `bin/` starts from `$0`, walks the chain, carries both `case` branches, derives a physical `repo_dir`, and never derives a path from bare `$0`
- `bin/skills`, `bin/agents --help`, `bin/doctor --help` through an absolute link, a relative link, a link to a link, a linked `bin/` directory, a linked checkout, and a keg-shaped wrapper — each must behave byte-for-byte as at the real path, under an isolated `HOME`/`CODEX_HOME`/`MODEL_ROUTER_STATE_DIR`. `skills` and `agents` reach guarded modules and so cover the silent-exit-0 shape; `doctor` prints usage unguarded and covers only the missing-module shape
- a control taking the resolver verbatim from `bin/skills`, showing the bare `dirname $0` form landing one level above the link's directory, and resolving the Stow layout

`sh -n` and `dash -n` pass on every script. Full suite locally on macOS: 3977 tests, 3947 pass, 28 skipped, 2 failures (`desktop-panel` needs playwright, a `search-sidecar` cancellation timing test) that reproduce unpatched on `99d526d`.

### DIFF
--- a/bin/agents
+++ b/bin/agents
@@ -1,4 +1,17 @@
 #!/bin/sh
 set -eu
-repo_dir=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)
+# The name on PATH may be a symlink to this script, so walk the chain before
+# locating the repository root, and resolve that root physically: `dirname $0`
+# alone points beside the link, and a logical `..` climbs out of a linked
+# directory to a path node does not recognise as a module's own when it guards
+# main, so the command would quietly do nothing.
+self=$0
+while [ -L "$self" ]; do
+  link=$(readlink "$self")
+  case $link in
+    /*) self=$link ;;
+    *) self=$(dirname -- "$self")/$link ;;
+  esac
+done
+repo_dir=$(CDPATH= cd -P -- "$(dirname -- "$self")/.." && pwd -P)
 exec node "$repo_dir/src/agent-bridges.mjs" "$@"
--- a/bin/api-key
+++ b/bin/api-key
@@ -1,5 +1,18 @@
 #!/bin/sh
 set -eu
 
-repo_dir=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)
+# The name on PATH may be a symlink to this script, so walk the chain before
+# locating the repository root, and resolve that root physically: `dirname $0`
+# alone points beside the link, and a logical `..` climbs out of a linked
+# directory to a path node does not recognise as a module's own when it guards
+# main, so the command would quietly do nothing.
+self=$0
+while [ -L "$self" ]; do
+  link=$(readlink "$self")
+  case $link in
+    /*) self=$link ;;
+    *) self=$(dirname -- "$self")/$link ;;
+  esac
+done
+repo_dir=$(CDPATH= cd -P -- "$(dirname -- "$self")/.." && pwd -P)
 exec "$repo_dir/bin/provider-key" kimi-api "$@"
--- a/bin/caller-key
+++ b/bin/caller-key
@@ -1,5 +1,18 @@
 #!/bin/sh
 set -eu
 
-repo_dir=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)
+# The name on PATH may be a symlink to this script, so walk the chain before
+# locating the repository root, and resolve that root physically: `dirname $0`
+# alone points beside the link, and a logical `..` climbs out of a linked
+# directory to a path node does not recognise as a module's own when it guards
+# main, so the command would quietly do nothing.
+self=$0
+while [ -L "$self" ]; do
+  link=$(readlink "$self")
+  case $link in
+    /*) self=$link ;;
+    *) self=$(dirname -- "$self")/$link ;;
+  esac
+done
+repo_dir=$(CDPATH= cd -P -- "$(dirname -- "$self")/.." && pwd -P)
 exec node "$repo_dir/src/caller-key.mjs" "$@"
--- a/bin/chatgpt-session
+++ b/bin/chatgpt-session
@@ -1,5 +1,18 @@
 #!/bin/sh
 set -eu
 
-repo_dir=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)
+# The name on PATH may be a symlink to this script, so walk the chain before
+# locating the repository root, and resolve that root physically: `dirname $0`
+# alone points beside the link, and a logical `..` climbs out of a linked
+# directory to a path node does not recognise as a module's own when it guards
+# main, so the command would quietly do nothing.
+self=$0
+while [ -L "$self" ]; do
+  link=$(readlink "$self")
+  case $link in
+    /*) self=$link ;;
+    *) self=$(dirname -- "$self")/$link ;;
+  esac
+done
+repo_dir=$(CDPATH= cd -P -- "$(dirname -- "$self")/.." && pwd -P)
 exec node "$repo_dir/src/chatgpt-session.mjs" "$@"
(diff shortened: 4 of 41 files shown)


duolahypercho:
Reviewed line by line, and reproduced the bug against a live checkout before believing it. This is a clean **merge** from me.

The load-bearing half is not the symlink walk — `bin/codex-router` already had that on `main` — it is `cd -P` / `pwd -P`. Every `src/*.mjs` guards main with `path.resolve(process.argv[1]) === fileURLToPath(import.meta.url)`, and Node realpaths a main module while leaving `argv[1]` as written, so a `repo_dir` still containing a symlinked directory imports fine and then exits 0 having done nothing:

```
logical cd → .../aliased        → node src/skills-install.mjs → exit 0, no output
cd -P      → <real checkout>    → exit 2, usage printed
```

That is the Homebrew keg shape (`opt/codex-router` → `Cellar/.../libexec`), where nothing on the path is a link `[ -L ]` can see — which is why the walk alone never fixed it.

Checked the resolution block against direct invocation, absolute and relative links, chains of two, paths containing spaces, `$PATH` lookup under sh/bash/zsh, `sh bin/x` vs `./bin/x`, an unrelated cwd, a symlinked directory, and a stow-style layout. Correct in every case. `readlink` with no flags is the only portable spelling — macOS BSD `readlink` has no `-f` and rejects `--` — and all 39 scripts are byte-identical apart from the `CDPATH=` vs `CDPATH=''` split that already existed on `main`. No CRLF regression against `bin/* text eol=lf`, and no source-root skew: everything persisted still derives from `src/paths.mjs`, not from the shell's `repo_dir`.

The Homebrew keg test is the one that pins the fix — it builds the real layout and fails with the old logical `cd`.

Two notes, neither blocking and neither worth changing now:

- the `while [ -L "$self" ]` walk has no iteration cap, so a symlink cycle would spin. I could not construct a reachable path to it: for the script to be running, the kernel already resolved `$0`, which rules out a cycle in its chain (ELOOP at exec). A counter capped around 40 would close it if you touch these again.
- `readlink` without `--` is unreachable in practice (the relative branch always prefixes `$(dirname -- "$self")`, so a target can never start with `-`), and adding `--` would break BSD. Noting it so nobody "fixes" it later.

Thanks — this is careful work, and the keg case is exactly the kind of thing that stays invisible until someone installs the way you did.

duolahypercho:
Merged. I reproduced both failure modes here independently of your fixture,
which answers the part you flagged as unverified.

Keg-shaped layout (`opt/codex-router` a symlink into
`Cellar/codex-router/<version>`, `libexec/bin/codex-router` a real file), every
command run through `opt/.../libexec/bin/codex-router`:

| command | main | this PR |
| --- | --- | --- |
| `skills` | exit 0, 0 bytes | exit 2, usage |
| `agents` | exit 0, 0 bytes | exit 2, usage |
| `caller-key` | exit 0, 0 bytes | exit 2, usage |
| `providers` | exit 0, 0 bytes | exit 0, 1761 bytes |
| `support-bundle` | exit 0, 0 bytes | exit 0, 134 bytes |
| `refresh-catalog` | exit 0, 0 bytes | exit 1, 736 bytes |

And your point about why CI never saw it holds: `--version` and `help` — the two
the formula's `test do` block runs — return normally on main, because neither
crosses a main-module guard.

Second mode confirmed too: `ln -s <checkout>/bin/doctor <somewhere>/doctor` then
`doctor --help` gives `Cannot find module` on main and correct usage here.

On your two flagged behaviour changes: the `bin/install` hint printing
`${CODEX_ROUTER_SOURCE_ROOT:-$repo_dir}` is right. The tray's sealed root going
from logical to physical is fine as you have it — no need to keep the logical
path.

A real `brew install` is still blocked on the tap here, but the mechanism this
fixes is the `opt/<formula>` symlink, and that is exactly what the fixture
exercises. Thanks for tracking this down — the silent exit-0 class of failure is
the worst kind to find.

Issue #639, opened 2026-09-06
[Bug]: Every turn after one completed web search fails with model_search_not_supported on routed Responses-wire models


Trewyn:
### What happened?

**Expected:** on a routed model without a search-capability declaration, a built-in web search executes and the thread keeps working.

**Actual:** the first web search succeeds (a completed `web_search_call` item lands in the session history), and from the next turn on **every** request — search-related or not — is rejected pre-flight with:

```json
{"error":{"type":"model_search_not_supported","message":"Model opencode-go-responses/muse-spark-1.3-contributor can no longer preserve this turn's web-search execution contract."}}
```

router.log shows bursts of `status=400 total_ms=11` — the request never reaches the upstream. The thread is permanently unusable; disabling web search afterwards does not recover it (the item is already in the replayed history).

**Root cause:** `routedModelPreservesSearchContract` (src/search-capability.mjs) requires `supportsSearchHistory === true` whenever the input contains `web_search_call` items and no search mode is required. The opencode curation entries carry only `supports_search_tool: false` — no `searchTool`, no `supportsSearchHistory` — so the gate rejects the replay. The gate sits before the LiteLLM translation hop, so it applies to all uncertified routed models. Note that `user-models.json` overrides cannot patch this: the registry skips user entries that duplicate a checked-in route.

### Reproduction

1. Route Codex through codex-router to `opencode-go-responses/muse-spark-1.3-contributor` (wire_api = responses).
2. Ask anything that triggers built-in web search → search succeeds, turn completes.
3. Send any follow-up message in the same thread → 400 `model_search_not_supported` on every subsequent turn.

Live repro from today (router.log):

```text
22:01Z  status=0    total_ms=32242  out_tokens=2740  cached_tokens=209702  (search turn, completed)
22:04Z  status=400  total_ms=11                                          (every turn after)
```

### Evidence that the upstream actually accepts the replay

Probes hit the LiteLLM model group directly (bypassing only the router pre-flight), replaying one completed `web_search_call` item (`action.query = "北京今天天气"`) followed by the question "which city does the search record query?":

| request | result |
|---|---|
| baseline (no search item) | 200, completed |
| + `web_search_call` item | 200, completed — model answers 北京 (item semantically intact) |
| same, streaming | healthy SSE stream through `response.completed` |

Identical result for `muse-spark-1.2-contributor`.

### Suggested fix

[...]

PR #732, opened 2026-09-12
fix(free): free Muse Spark 1.3 keeps web-search history

antoniofern:
Live replay probe on `opencode-free-responses/muse-spark-1.3-contributor-free`: a resumed thread carrying 5 completed `web_search_call` items returned 4x200 instead of `400 model_search_not_supported`.

Probe data (2026-09-12): before — same follow-up turn failed in 216ms with 0 tokens (`router.log 18:30:17 status=400 total_ms=83`); after — `codex exec resume` of the same session exited 0 with a normal answer (upstream `18:54-18:55`: 4x200, 19.6s/11.4s/5.6s/9.2s, out 873+714+260+336). `test/search-capability.test.mjs` 7/7 pass; fresh `codex exec` smoke `router-ok` exit 0.

Mirrors #639 (go twins), which carry the same flag. Change: `supportsSearchHistory: true`, compHash v1->v2, plus test pin.

### DIFF
--- a/config/opencode-free/responses/muse-spark-1.3-contributor-free.json
+++ b/config/opencode-free/responses/muse-spark-1.3-contributor-free.json
@@ -40,10 +40,11 @@
         "image"
       ],
       "isFree": true,
+      "supportsSearchHistory": true,
       "supportsApplyPatchTool": false,
       "requestProfile": "auto-tool-choice",
       "toolSchemaRecursion": "flatten",
-      "compHash": "opencode-free-responses-muse-spark-1-3-contributor-free-v1"
+      "compHash": "opencode-free-responses-muse-spark-1-3-contributor-free-v2"
     }
   ]
 }
--- a/test/search-capability.test.mjs
+++ b/test/search-capability.test.mjs
@@ -67,6 +67,7 @@ test("checked-in opencode Muse Responses routes replay verified search history w
   for (const slug of [
     "opencode-go-responses/muse-spark-1.2-contributor",
     "opencode-go-responses/muse-spark-1.3-contributor",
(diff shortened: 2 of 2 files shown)


duolahypercho:
Reviewed — **merge**. I checked the thing that usually goes wrong with a curation-only change: whether the field is actually read. It is. `src/user-models.mjs:65` lists it in `METADATA_FIELDS`, `src/model-registry.mjs:682` type-validates it, and `src/search-capability.mjs:46` is the consumer that matters — it gates `src/router.mjs:993`, which throws `model_search_not_supported` (HTTP 400) without it, plus the failover candidate filters. So flipping it genuinely changes behaviour rather than decorating the catalog.

Schema and conventions match the sibling `config/opencode/go-responses/muse-spark-1.3-contributor.json`: same field, same placement beside `supportsApplyPatchTool`, and a compHash bump on a capability change is the established pattern. `searchTool` stays absent, so the model still cannot *start* a search — which is the right narrower claim.

One note for the record, not a blocker: the two routes are different endpoints (`zen/v1` here vs `zen/go/v1` for the Go relay), and this repo is deliberately strict about not projecting a capability across transports — `test/registry.test.mjs:797` keeps GLM-5.3 search off on the Go relay for exactly that reason. Your commit records the probe (4x200 instead of 400 on a resumed thread with 5 completed `web_search_call` items), which is the evidence `docs/HOW-IT-WORKS.md:247` asks for, so this is fine. And the blast radius if a probe were ever wrong here is small: the route is native passthrough, so a bad claim yields an upstream 4xx rather than a silent wrong answer.

The test-loop entry is load-bearing — revert the JSON field and it fails.

PR #733, opened 2026-09-12
chore: merge CHANGELOG.md bullets instead of conflicting on them

duolahypercho:
## Why

Every change appends a bullet to the top of `CHANGELOG.md`'s `## Unreleased` section, so two branches that share no source file still collide there.

That is not a minor annoyance — it is currently the single largest source of conflicts in the open queue. Of the ten conflicting open pull requests, **seven conflict on `CHANGELOG.md` and nothing else**:

| PR | conflicting files before | after |
|---|---|---|
| #708 | CHANGELOG.md | none |
| #710 | CHANGELOG.md | none |
| #714 | CHANGELOG.md | none |
| #715 | CHANGELOG.md | none |
| #716 | CHANGELOG.md | none |
| #728 | CHANGELOG.md | none |
| #729 | CHANGELOG.md | none |

## What

Four lines plus a comment in `.gitattributes`:

```
CHANGELOG.md merge=union
```

`union` is git's built-in driver for exactly this shape — an append-only list whose entries are independent of each other. It keeps both sides' bullets rather than asking a human to re-type one.

## Verification

Each of the seven was re-merged against `main` with the attribute in place (`git merge-tree --write-tree`), and all seven go to zero conflicts. The three remaining conflicts are genuine source conflicts, untouched by this change:

- #720 — `test/tool-schema-root.test.mjs`
- #725 — `src/router.mjs`
- #605 — `src/chat-tool-surface.mjs`, `src/namespace-relay.mjs`, `src/router.mjs`, `test/provider-catalogs.test.mjs`

I also checked the merged output rather than just the conflict count: no conflict markers, and no bullet lost from either side.

## The tradeoff, stated plainly

When both sides edit the *same* bullet, `union` keeps both copies. In one of the test merges that produced a single duplicated bullet (210 where 209 were expected). That is a one-line delete with both wordings visible, rather than a conflict someone has to reconstruct from two halves — and it only ever applies to this one file.

If a cleaner shape is wanted later, the alternative is `changelog.d/` fragments assembled at release, which never conflict and never duplicate, at the cost of a release-time assembly step.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -17,3 +17,15 @@ bin/* text eol=lf
 # and are conventionally CRLF, but nothing here depends on their endings, so
 # they are normalized in the repository and left to the platform on checkout.
 *.ps1 text
+
+# Every change here appends a bullet to the top of CHANGELOG.md's `## Unreleased`
+# section, so two branches that share no source file still collide on it. That
+# was the single largest source of conflicts in the open queue: of ten
+# conflicting pull requests, seven conflicted on this file and nothing else.
+#
+# `union` is git's built-in driver for exactly this shape -- an append-only list
+# whose entries are independent. It keeps both sides' bullets instead of asking
+# a human to re-type one. The cost is that when both sides edit the *same*
+# bullet it keeps both copies, which is a visible one-line delete rather than a
+# conflict to resolve; pick the surviving wording when that happens.
+CHANGELOG.md merge=union


duolahypercho:
## Correction — this does not do what the description claims

I verified the effect after merging and the central claim above is wrong. Recording it here rather than leaving the description to mislead.

**GitHub's merge engine does not honour `merge=union`.** Tested directly: a branch created at `main` *with* this attribute committed, then a server-side merge of #708's head into it via the merges API:

```
POST /repos/duolahypercho/codex-router/merges
{"message":"Merge conflict", "status":"409"}
```

So GitHub's mergeability indicator and merge button are unaffected by this change. (The probe branch has been deleted.)

**What actually happens:** `git` honours it, so anyone merging or rebasing locally — `git merge origin/main`, `git pull`, `git rebase` — gets the bullets combined instead of a conflict. That is real but much narrower than "unblocks seven pull requests".

**Why it looked like it worked.** Right after merging, four PRs flipped from conflicting to clean. None of that was this change:

- #714 and #729 — their merge base already carries `main`'s current `CHANGELOG.md` blob (`a9743b34`), so `main` had not touched the file since they branched and no conflict was possible either way.
- #728 — the author rebased at 21:08, after review feedback (`79d8f47c` → `6c0888cd`).
- #720 — its conflict was never in `CHANGELOG.md`; it was `test/tool-schema-root.test.mjs`.

Meanwhile #708, #710, #715 and #716 still report `CONFLICTING/DIRTY` on GitHub, exactly as before, even though `git merge-tree` resolves all four cleanly with the attribute active. That gap is the whole story.

**Keeping this** rather than reverting: it is inert where it does not help, and it genuinely removes the conflict for local merges and rebases, which is how these branches actually get updated. But it is not the queue-wide fix I described.

**The real fix**, if the CHANGELOG churn is worth solving properly, is `changelog.d/` fragments assembled at release — one file per change, so two branches never touch the same file and GitHub sees no conflict either. That is a genuine workflow change and should be its own decision, not smuggled in here.

PR #736, opened 2026-09-13
fix(commandcode): normalize nullable Gemini tool schemas

ml0-1337:
Command Code's Gemini 3.8 Flash can reject a Codex turn before any tool runs when the supplied tools contain a nullable array. The reported rejection identifies the image-generation tool's `referenced_image_paths` parameter:

> schema specified other fields alongside any_of. When using any_of, it must be the only field set.

Normalize a single concrete type plus `null` into an explicit union on this provider/model route. The non-null branch keeps its constraints, so an array still declares its item type. For example:

```json
{"type":["array","null"],"items":{"type":"string"}}
```

becomes:

```json
{"anyOf":[{"type":"array","items":{"type":"string"}},{"type":"null"}]}
```

The compatibility pass runs after namespace flattening and tool-search history expansion, covering both initially supplied and later discovered definitions. It preserves tool identities, parameter requirements, and unrelated routes. The nullable rewrite skips nodes carrying composition, reference, or enum/const constraints, and nodes declaring several non-null types. Literal values in defaults and examples are not traversed as schemas.

The patch adds no dependency or model configuration change. It contains only this compatibility fix, its regression tests, and a changelog entry.

## Validation

- `npm run check` passed.
- `node --test --test-concurrency=2 --test-timeout=180000 test/gemini-tool-schema.test.mjs test/routing.test.mjs test/tool-schema-root.test.mjs test/namespace-relay.test.mjs test/namespace-relay-routing.test.mjs test/grok-reasoning-summary-compat.test.mjs test/reasoning-tag-stripper.test.mjs` passed: 400 tests, no failures or skips. Run with an isolated `CODEX_HOME` and no outer Router-state overrides, so the fixtures own their state directories.
- Removing only the new compatibility invocation makes the routing regression fail on the unrepaired schema. Restoring it makes the test pass.
- `git diff --check` passed.

The reported provider error and earlier local deployment motivated this fix. No new live provider request was made while preparing this PR. The converter test models the observed conversion; it is not a live Google or Command Code acceptance test. The full repository suite and other operating systems are left to CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+- **Command Code's Gemini 3.8 Flash accepts nullable tool parameters.** Convert
+  a single concrete type plus `null` to an explicit union before forwarding,
+  preserving array constraints and nested schemas. This prevents the observed
+  `any_of` sibling-field rejection for Codex's image-generation tool, including
+  definitions restored from tool-search history. Other routes keep their schemas.
 - **Blank tool-call messages and direct DeepSeek reasoning are repaired behind
   large tool lists.** LiteLLM echoes the request's instructions and whole tool
   list in `response.created` and `response.in_progress`. The stream repairs
--- a/src/gemini-tool-schema.mjs
+++ b/src/gemini-tool-schema.mjs
@@ -1,0 +1,210 @@
+// Gemini rejected image_gen's nullable type arrays after upstream translation
+// introduced anyOf with sibling fields. Spell one concrete type plus null as
+// an explicit union instead; retain its type-specific constraints on the
+// non-null branch. The Google AI SDK recognizes this shape and flattens it to
+// a nullable concrete type, without the invalid anyOf siblings:
+// https://github.com/vercel/ai/blob/main/packages/google/src/convert-json-schema-to-openapi-schema.ts
+//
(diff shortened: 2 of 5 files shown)


duolahypercho:
Two separate things here, and only one is yours:

- **The branch conflicts with main.** That is the blocker — please rebase.
- **The red `windows-latest / python 3.12` check is not your diff.** It failed
  with `cancellation did not reach mock upstream within 60000 ms on win32`,
  which is a known flake in this repo. It should clear on the rerun after the
  rebase; if it does not, say so and I will look.

duolahypercho:
Thanks, this merges cleanly with `main`. After the merge, the `gemini-tool-schema`, `routing`, `tool-schema-root` and `namespace-relay-routing` tests pass (266) and `npm run check` passes. The earlier rebase request is resolved.

Two things are left:
- The PR edits `CHANGELOG.md` directly. Please move the entry into a `changelog.d/<short-slug>.md` fragment instead (see `changelog.d/README.md`).
- Only GitGuardian has run on the current head. Full CI needs to run before this can merge.

Worth knowing: the normalization covers only `commandcode/google/gemini-3.8-flash`. `gemini-3.7-flash` and `gemini-3.5-flash` aren't included. If they have the same nullable-schema problem, they'd need to be added.

---

PR #737, opened 2026-09-13
fix: locate root config assignments with the TOML lexer

c8dhjp4tyv-bit:
## Why

Root config helpers treated assignment-shaped lines inside multiline strings as settings, silently rewriting user prose and journalling the wrong previous model.

## Changes

- Locate root settings with the TOML structural lexer. Root reads, writes and splitting stop at the first real table, so an invalid escape in a later table cannot change the selected assignment.
- Retain strict whole-document validation for login-free in-place changes.
- Decode ordinary string values with the lexer, including trailing comments; refuse multiline model settings before writing.
- Preserve assignment-shaped prose during native-catalog preparation, adoption, cleanup and restore.
- Use structural assignments and table boundaries for managed concurrency insertion, and recognize managed markers only as standalone comments when the document can be scanned.
- Include current main through 0964eef.

## Compatibility limit

Malformed root sections retain legacy best-effort matching for Windows prototype migration/disable. Prose preservation is guaranteed for structurally readable root sections; login-free in-place edits continue to refuse malformed documents. Optional agent settings are not inserted into unreadable documents.

## Verification

`node --test test/config-manager.test.mjs test/legacy-migration.test.mjs test/native-catalog-source.test.mjs test/native-catalog-drift.test.mjs test/target-integration.test.mjs test/login-free-app-server.test.mjs test/caller-key-client-refresh.test.mjs`

91 passed, 1 skipped, 0 failed (Linux, Node 24.19.0). New coverage exercises quoted/commented model restoration, invalid trailing tables, prose containing managed markers and concurrency examples, catalog adoption/restore, strict login-free refusal, and multiline-model refusal without modifying the file.

`npm run check`, final `node --check src/config-manager.mjs`, and `git diff --check` passed. Windows CI is still required; local results do not establish Windows execution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,27 @@
 
 ## Unreleased
 
+- **A multiline string in `config.toml` is no longer edited as if it were
+  settings.** The root-level helpers found assignments by matching lines, so a
+  line inside a multiline string -- prose in `instructions`, a documented
+  example -- counted as one whenever it was shaped like `model = ...`. Because
+  it came first, it was the line that got rewritten, deleted when a value
+  moved, and read back as the current setting. Switching the model edited the
+  user's prose and left the real `model` untouched; enabling the router deleted
+  a line out of the middle of their text; and the router journalled the prose
+  as the previous value, so turning it off restored that. All of it silent, and
+  the setting the user asked for never changed. These helpers now locate the
+  assignment through `scanTomlDocument` -- the fail-closed structural lexer this
+  file already uses for table boundaries, which exists to tell a real
+  assignment from text that looks like one -- and decode the line exactly as
+  before, so only the choice of line changes. The root/table split is read from
+  the same lexer, so a string containing a `[`-leading line no longer ends the
+  root section early. A `config.toml` the lexer cannot read at all keeps the
+  line matching it has always had: the prototype installer wrote the catalog
+  path unescaped, so on Windows that value carries escapes TOML does not define,
+  and failing closed there would leave the operator unable to so much as disable
+  the router.
+
 - **Reasoning from Chat Completions models now shows in Codex.** LiteLLM's
   Chat Completions to Responses bridge opens the assistant message first and
   streams the model's reasoning under a fresh hashed item id per delta, with no
--- a/src/config-manager.mjs
+++ b/src/config-manager.mjs
@@ -509,9 +509,30 @@ function withManagedAgentConcurrency(input) {
   return `${lines.join("\n").trimEnd()}\n`;
 }
 
+// The structural read of a config, or undefined when the lexer refuses it. A
+// legacy config can carry an unescaped Windows path inside a basic string --
+// `model_catalog_json = "D:\\a\\kimi-proxy\\merged-models.json"`, whose `\\a` is
+// not a valid TOML escape -- and refusing to touch those would leave the
+// operator unable to so much as disable the router. Those keep the line
+// matching they have always had; every config that can be read structurally
+// gets the accurate answer.
+function scannedConfig(contents) {
+  try {
+    return scanTomlDocument(contents);
+  } catch {
+    return undefined;
+  }
+}
+
 function splitRoot(input) {
   const lines = input.split("\n");
-  const firstTable = lines.findIndex((line) => /^\s*\[/.test(line));
+  // A line inside a multiline string can begin with `[` without opening a
+  // table; asking the structural lexer keeps the boundary honest, and keeps
+  // `rootLines` a document the lexer can read on its own below.
+  const scanned = scannedConfig(input);
+  const firstTable = scanned
+    ? scanned.headers[0]?.index ?? -1
+    : lines.findIndex((line) => /^\s*\[/.test(line));
   return firstTable === -1
     ? { rootLines: lines, tableLines: [] }
     : { rootLines: lines.slice(0, firstTable), tableLines: lines.slice(firstTable) };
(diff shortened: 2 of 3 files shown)


duolahypercho:
This is the right fix in the right place, and the test proves it — I ran your test file against `origin/main`'s source and it fails at the `router-default-set` assertion, with the `model = "documented example, not a setting"` line removed from inside the `instructions` string. A byte-level assertion, not a shape check.

I probed the new locator directly on the shapes that usually break this kind of change: a multi-line root array, a root value that *is* a multiline string, a dotted key, a quoted root key, an inline table, an `[[array-of-tables]]` member, a commented-out line, a trailing-comment assignment, prose containing a `[`-leading line, duplicate root keys, and CRLF. All resolved to the correct line, and I could not find a config where `main` is right and this is not.

Two side improvements worth noting: a quoted root key `"model" = "x"` is now recognised (main ignored it and would have *added* a duplicate `model =`), and the root/table boundary no longer ends early on a `[`-leading line inside a string.

Three things before I merge it.

**1. `replaceRootValueInPlace` went from fail-closed to fail-open.** Main called `scanTomlDocument()` unguarded there, so an unreadable config surfaced as a named refusal. Now it goes through `rootAssignmentIndexes()`, whose `scannedConfig()` swallows the throw and falls back to line matching.

Failure case: a config with an unescaped Windows path anywhere (so the lexer refuses) *and* assignment-shaped prose in a root multiline string. Login-free enable then writes `model = "<login-free>"` over the prose line and leaves the real `model =` line below it — the later one wins in TOML, so the model switch silently does nothing while the user's text is overwritten. Main refused instead.

Your rationale for the fallback ("the operator must still be able to disable the router") does not require this one: `disable` goes through `replaceRootValue`, not `replaceRootValueInPlace`. Keeping the strict scan here would preserve the refusal.

**2. The fallback is document-global, so reads and writes can disagree.** `rootAssignmentIndexes()` scans the whole document, while `rootValue`/`rootHasValue` re-lex only `rootLines`. When the document is unlexable because of something *after* the first table, the slice still lexes. Reproduced:

```toml
instructions = """
model = "prose-do-not-touch"
"""
model = "real-gpt"

[mcp_servers.x]
command = "C:\bad\escape"
```

`router-default-set` journalled `"real-gpt"` correctly (slice scan) but still removed the prose line (full-document scan refused, so the regex path ran). Net better than main, which additionally restored `"prose-do-not-touch"` as the model — but the bug this PR exists to fix is still reachable, and the two locators now disagree about which line is the assignment. Either lex the root slice in `rootAssignmentIndexes` too, or note in the changelog that the fix does not hold for a document the lexer refuses.

**3. Two siblings of the same class were left on raw line matching.** The `model_catalog_json` regex filter appears at both `src/config-manager.mjs:1484` and `:1552`, so a `model_catalog_json = ...` line inside a user's `instructions` is still silently dropped on `--adopt-native-catalog` and on native-catalog restore — the exact bug your changelog describes. `withManagedAgentConcurrency()` (:478) scans the same way, and `replaceRootValue()`'s insertion point is still a `findIndex` on the start marker, which would splice into the middle of a string containing that marker text. Mechanical to convert now that `rootAssignmentIndexes()` exists.

Non-blocking: a root key whose value spans lines still loses only its first line. Identical to main, and only `model`/`model_provider` reach these helpers — but the lexer now *tells* you the assignment is `kind: "multiline-string"`, so refusing that case would match the repo's "refuse rather than guess" norm. And `assignmentValue()` still decodes by string surgery (`model = "gpt-5.6" # note` decodes to `gpt-5.6" # note`) even though the record you already fetched carries a properly decoded `value`.

For the record: `a config the TOML lexer refuses is still editable` passes unchanged on `main` — it guards the new fallback rather than evidencing the fix. And a failed write cannot truncate the config; everything still goes through `atomicWrite()`.

c8dhjp4tyv-bit:
Addressed the review in d1f12d16347a8679133da502ec1ff917e6beb24c, including syncing with main.

1. `replaceRootValueInPlace` once again validates the entire document strictly. The login-free regression test verifies refusal leaves the config byte-identical.
2. Root reads and writes now share a structural root-only scan that stops at the first real table. An invalid escape in a later table no longer sends writes down the regex fallback; table-looking prose is preserved too.
3. Catalog preparation/adoption/cleanup/restore and agent concurrency insertion now use structural assignment/table information. Managed markers inside multiline strings are preserved and cannot become insertion points or removable blocks. The related v2 feature insertion also uses real table boundaries.

Also addressed the non-blocking points: ordinary string values use the lexer's decoded value (including trailing comments), and multiline model settings are refused before publication instead of partially deleted. The changelog and PR description now state the remaining malformed-root fallback explicitly.

Verification: 91 passed, 1 skipped, 0 failed across config-manager, legacy-migration, native-catalog-source/drift, target-integration, login-free-app-server and caller-key-client-refresh tests. `npm run check`, final source syntax check and `git diff --check` passed. These are Linux results; Windows CI still needs to confirm platform behavior.

c8dhjp4tyv-bit:
Rebased onto `main` (the conflict was `CHANGELOG.md` only, resolved by the `merge=union` driver — GitHub reports it as a conflict because its mergeability check does not apply `.gitattributes` merge drivers).

Worth recording, since this branch has now collected a red badge four separate times and none of them were this diff:

| run | failing job | cause | on re-run |
| --- | --- | --- | --- |
| 1 | `test (windows-latest)` | `legacy-migration` — **a real bug in my first version**, fixed in this branch | — |
| 2 | `test (windows-latest)` | `antigravity-cli-onboarding`, `login.status === null` (child killed by signal, empty stderr) | pass |
| 3 | `Electron control center (Windows)` + `Unified native macOS app` | `Response code 504` from electron-builder's downloads | pass |
| 4 | `macos-latest / python 3.12`, then `test (macos-latest)` | Grok cancellation `within 60000 ms on darwin`; then `5-level owner SIGINT cleanup` in `process-tree` | Grok one passed; SIGINT one re-running |

Only the first was mine, and it is fixed. For the last one: `process-tree.test.mjs` references nothing this branch touches, and the heavier `12-level owner SIGTERM` variant of the same test passed in the same run at 4381 ms while the 5-level SIGINT one failed at 366 ms — signal-delivery timing on a loaded runner, not depth.

The diff is five files: `CHANGELOG.md`, `src/config-manager.mjs`, `src/native-catalog-source.mjs`, `src/toml-structure.mjs`, `test/config-manager.test.mjs`. The config-manager suites have passed on every run on all three platforms.

Not asking you to take anything on trust — just so a red check here does not cost you a bisect.

Issue #726, opened 2026-09-12
nonRecursiveToolSchema leaves a typeless {} that Moonshot rejects


duolahypercho:
Follow-up split out of #641 / #723, which fixed the client-supplied half of that bug.

## The gap

`nonRecursiveToolSchema` breaks a `$ref` cycle by deleting the `$ref` and leaving `{}` behind (`cloneWithoutClosingRefs`, `src/tool-schema-root.mjs:169`). A bare `{}` declares no `type`, which is exactly the shape Moonshot answers with:

```
tools.function.parameters missing type in anyOf properties
```

So the router can **create** this 400 on a Kimi route even when the client's schema was perfectly clean.

Reproduced offline on current `main`, no credentials, no provider request:

```
B. cycle-closing $ref blanked by nonRecursiveToolSchema
   $defs.Node.properties.child -> {}
   has `type`  : false
```

## Why it wasn't fixed in #723

#723 adds `declareSchemaTypes` inside the Moonshot compatibility pass in `src/router.mjs`. This producer runs **later and elsewhere**: `flattenRecursiveToolSchemas` is called unconditionally at `src/api-forwarder.mjs:877`, downstream of the router's Moonshot pass and across every provider. Widening #723 to cover it would have meant either running the type pass in a second place or making it provider-agnostic — the second changes the wire payload for providers that are working today, which #723 deliberately avoids.

## Notes for whoever picks this up

- `{}` from a blanked cycle-closing `$ref` means "whatever the target was" — `{ type: "object", additionalProperties: true }` is the honest reading, but confirm that against Meta/opencode first, since `nonRecursiveToolSchema` exists for *their* 400 (`src/api-forwarder.mjs:397-403`) and currently works.
- Scope any change to the routes that need it. The comment at `api-forwarder.mjs:397` records that an ordinary `$ref`/`$defs` pair is accepted by Meta's Console API and only a self-re-entering reference is refused, so the blanking itself should stay.
- `declareSchemaTypes` from #723 is exported from `src/tool-schema-root.mjs` and is reusable as-is.

Not urgent: it needs a recursive tool schema *and* a Moonshot route to bite, so it is rarer than the #641 path.

duolahypercho:
Closing — the route this describes is not reachable in shipped config, and the one place it *is* reachable was fixed by #738. Verified offline against `main`.

**The forwarder's blanking never runs on a Kimi request.** `flattenRecursiveToolSchemas` has exactly one call site, and it is gated:

[...]

PR #738, opened 2026-09-13
fix: keep the type a broken $ref cycle was declaring on Moonshot routes

c8dhjp4tyv-bit:
Related to #726; this does not claim to close that issue.

## Scope

Only locally curated Moonshot models with `toolSchemaRecursion: "flatten"` reach the type-preserving path. No shipped Kimi model enables flattening. The route that produced #726 has not been established. Stock Kimi and other providers retain their existing behavior.

## Changes

- Preserve explicitly declared target types when flattening a recursive reference on opted-in Moonshot routes; follow aliases with cycle detection.
- Prefer the referencing node's type-implying siblings. Ambiguous enum/union siblings prevent fallback to a contradictory target type. Explicit types remain untouched.
- Accept null options with the same behavior as omitted options.
- Share the unchanged Moonshot route predicate between router and forwarder.
- Merge current main to clear the conflict.

## Deliberate limitation

An untyped recursive union such as `N = anyOf(string, ref N)` still leaves an open branch after flattening. Neither a concrete target type nor the provider's intended recursive semantics is established, so this patch does not guess a type. A regression test records that limitation. Reference-only rings likewise remain open.

## Verification

405/405 tests passed across tool-schema-root, generic-routing, routing, namespace-relay, registry, commandcode-forwarder, deepseek-responses-routing, and grok-tool-facade. `npm run check` and `git diff --check` passed.

The forwarder regression starts the real process against a local mock endpoint using a user-curated kimi-api model with flattening enabled. It verifies the typed edge, a non-Moonshot flattening control retaining `{}`, and a model without flattening retaining its reference. Additional regressions cover conflicting enum/const/items/prefixItems/properties siblings, ambiguous siblings, input immutability, and null options.

No live provider requests were made.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+- **Kimi no longer rejects a recursive tool schema the router itself made
+  typeless.** Breaking a `$ref` cycle drops the reference and leaves the node
+  behind, and a node with nothing left in it declares no type -- the one shape
+  Moonshot answers with HTTP 400 `tools.function.parameters missing type in
+  anyOf properties`, the same rejection #641 fixed for client-supplied schemas.
+  The type pass that fix added runs in the router's Moonshot compatibility hop,
+  while the cycle breaking runs later in the forwarder, so a recursive schema on
+  a Kimi route arrived with the type already removed again (#726). The cycle
+  edge now carries the type its target declared -- read out of the definition
+  the reference named, never guessed, and never over a type the client wrote
+  itself. Scoped to the measured Moonshot routes: the blanking exists for Meta's
+  Console 400, that path was not re-measured here, and every other provider and
+  model keeps the exact wire payload it has today.
+
 - **Reasoning from Chat Completions models now shows in Codex.** LiteLLM's
   Chat Completions to Responses bridge opens the assistant message first and
   streams the model's reasoning under a fresh hashed item id per delta, with no
--- a/src/api-forwarder.mjs
+++ b/src/api-forwarder.mjs
@@ -38,6 +38,7 @@ import {
 } from "./rate-limit-headers.mjs";
 import { recordRateLimitSnapshot } from "./rate-limit-state.mjs";
 import { recordProviderCooldown } from "./model-failover.mjs";
+import { moonshotSchemaRoute } from "./moonshot-schema-routes.mjs";
 import { cooldownScope } from "./provider-cooldown.mjs";
 import { canonicalProviderId, readProviderSelection } from "./provider-selection.mjs";
 import { stripImages, supportsImageInput } from "./vision-bridge.mjs";
@@ -402,7 +403,7 @@ function stripEncryptedToolSchemaAnnotations(payload, protocol) {
 // `nonRecursiveToolSchema` -- which blanks exactly the cycle-closing edge and
 // returns any other schema by identity -- is the whole fix. The namespace relay
 // already uses it for the same reason.
-function flattenRecursiveToolSchemas(payload, protocol) {
+function flattenRecursiveToolSchemas(payload, protocol, options) {
   if (!Array.isArray(payload.tools)) return;
   let changed = false;
   const tools = payload.tools.map((tool) => {
@@ -410,14 +411,14 @@ function flattenRecursiveToolSchemas(payload, protocol) {
       return tool;
     }
     if (protocol === "openai-responses") {
-      const flattened = nonRecursiveToolSchema(tool.parameters);
+      const flattened = nonRecursiveToolSchema(tool.parameters, options);
       if (flattened === tool.parameters) return tool;
       changed = true;
       return { ...tool, parameters: flattened };
     }
     const fn = tool.function;
     if (!fn || typeof fn !== "object" || Array.isArray(fn)) return tool;
-    const flattened = nonRecursiveToolSchema(fn.parameters);
+    const flattened = nonRecursiveToolSchema(fn.parameters, options);
     if (flattened === fn.parameters) return tool;
     changed = true;
     return { ...tool, function: { ...fn, parameters: flattened } };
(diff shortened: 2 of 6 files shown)


c8dhjp4tyv-bit:
Merged `main` in to clear the conflict. The only one was in `src/router.mjs`, and it was adjacency rather than content: #715 added the `leaked-tool-call-recovery` import at the same spot this branch adds `moonshot-schema-routes`. Both are kept, in that order; no logic hunk was touched, and `needsMoonshotSchemaCompatibility` still delegates exactly as before.

`CHANGELOG.md` resolved itself through the `merge=union` driver you added in #733 — worth noting that GitHub's mergeability check does not apply `.gitattributes` merge drivers, so a CHANGELOG-only conflict shows red in the UI even when it merges cleanly for you locally. That is what #737 was showing; it needed no manual resolution.

`npm run check` plus `test/tool-schema-root.test.mjs test/routing.test.mjs test/namespace-relay.test.mjs test/registry.test.mjs` — 357/357 on the merged tree.

duolahypercho:
The mechanics here are sound — I checked the parts that usually go wrong and they hold: a three-node ref-only ring and a self-referencing root both terminate, an unrecoverable type leaves the `{}` untouched rather than guessing, and since the only bytes added are one `type` per blanked node, the `MAX_INLINE_*` budgets aren't in play. The `router.mjs` → `moonshot-schema-routes.mjs` extraction is behaviour-identical, and it can't collide with #720's `inlineDanglingNestedDefsRefs` (that one is only reachable from `inlineGeminiToolSchemaRefs`). 54/54 on `tool-schema-root`.

But I don't think this reaches the bug it's for.

## The new option is always `false` on the shipped catalog

The only call site is inside `if (model.toolSchemaRecursion === "flatten")`. Exactly three shipped models set that field:

```
config/opencode-free/responses/muse-spark-1.3-contributor-free.json   opencode-free-responses
config/opencode/go-responses/muse-spark-1.3-contributor.json          opencode-go-responses
config/opencode/go-responses/muse-spark-1.2-contributor.json          opencode-go-responses
```

No `config/kimi/**` entry and no `config/opencode/go/kimi-k2.7-code.json` sets it, and `curatedModelToolSchemaRecursion()` only knows the two OpenCode-free records. For all three, `moonshotSchemaRoute()` returns `false` — your own test asserts exactly that. So `keepBlankedTypes` is `true` only for a hand-curated Kimi model with `toolSchemaRecursion: "flatten"`, which `src/user-models.mjs:72` does permit.

Meanwhile the Moonshot router hop calls `repairToolSchemaRoots(tools, { inlineForeignRefs: true, declareTypes: true })` with `nonRecursive` defaulted off, so it never blanks a cycle at all.

**Which route produced the 400 in #726?** If it was a stock Kimi route, the blanking is happening somewhere this PR doesn't change and the fix is in the wrong place. If it was a locally curated model, that belongs in the changelog — as written the entry reads as though stock Kimi routes are repaired.

## The `anyOf` shape the error names isn't repaired

```js
{ type:"object", properties:{ root:{ $ref:"#/$defs/N" } },
  $defs:{ N:{ anyOf:[{type:"string"},{$ref:"#/$defs/N"}] } } }
```

with `keepBlankedTypes: true` still yields `{"anyOf":[{"type":"string"},{}]}`. `closingRefType` resolves `#/$defs/N`, finds a node with neither `type` nor `$ref`, returns `undefined`; `declareSchemaTypes` also declines because the `$ref` branch had no declared type. That is literally "a node inside a union that declares no type" — the shape `tools.function.parameters missing type in anyOf properties` describes.

## The recovered type can contradict the node's own keywords

The guard is only `!("type" in target)`, so every other type-implying keyword is ignored:

```
{ $ref:"#/$defs/N", enum:["a","b"] }  →  N = { type:"object" }
result: {"enum":["a","b"],"type":"object"}    // a string enum declared as an object
```

On `main` the same node comes out `{"enum":["a","b"]}`, which is at least coherent. Same with an `items` or `properties` sibling. The comment says "every sibling the client authored, `type` included, still wins" — that's true of `type` alone. Inferring from the node's own keywords first (the `inferredType()` logic is already in this file) and falling back to the target only when that yields nothing would fix it, or just skip the stamp when the node carries `enum`/`const`/`items`/`prefixItems`/`properties`.

Minor: `nonRecursiveToolSchema(schema, null)` throws — the `= {}` default covers `undefined` but not `null`. No caller does it today; it's a trap for the next one.

## On the tests

Two genuinely prove the fix (`blanking a cycle edge keeps the type it pointed at`, and the alias/ring one). Two pass unchanged on `main` — `a type the client wrote on the referencing node is never overwritten` asserts a shape that holds either way; the contradicting-sibling case above is the test that would have earned that slot. And nothing asserts the forwarder wiring, which is what would have surfaced the reachability gap: there is no model fixture for which the new branch is taken.

c8dhjp4tyv-bit:
Addressed the review in 25a0030ddbd08ef52c39021062bcab6bea930dfa and synced the branch with main.

- Fixed the conflicting-sibling case: type-implying siblings are considered before the referenced target. Ambiguous enum/union siblings prevent fallback to a contradictory target type; explicit types remain untouched.
- Made `nonRecursiveToolSchema(schema, null)` behave like omitted options.
- Added a real forwarder-process regression against a local mock upstream. A locally curated `kimi-api` model with flattening enabled retains the target type, a non-Moonshot flattening control retains `{}`, and a model without flattening retains its reference.
- Added regressions for enum/const/items/prefixItems/properties siblings, ambiguous siblings, input immutability, and null options.

You were right about reachability. The changelog and PR description now explicitly scope this to locally curated Moonshot models with `toolSchemaRecursion: "flatten"`. No shipped Kimi model enables that option. I have not established which route produced #726, so I removed the claim that this closes it.

The untyped recursive `anyOf` example remains a limitation, not a repaired case. It still leaves an open branch after flattening; the patch does not guess a concrete type for that branch. A regression test records this behavior, and both the changelog and PR description call it out.

Verification: 405/405 tests passed across tool-schema-root, generic-routing, routing, namespace-relay, registry, commandcode-forwarder, deepseek-responses-routing, and grok-tool-facade. `npm run check` and `git diff --check` passed. No live provider requests were made.

PR #742, opened 2026-09-13
fix: keep an explicit subagent model instead of pinning it to the routed parent

Corpocore:
## Problem

`codex-router` writes `expose_spawn_agent_model_overrides = true` into the
managed `multi_agent_v2` block, so Codex offers a model override on
`spawn_agent`. The router then discarded that override:
`injectSessionModelForSpawnCalls` rewrote `spawn_agent.model` to the routed
parent on every call.

The effect is that a routed parent can only spawn copies of itself.
Cross-provider delegation, such as a DeepSeek parent delegating to a native
OpenAI model, is impossible even though the picker exists and
`subagent-report.mjs` lists those models as spawnable.

Reproduced on a live routed session:

- a child requested on `gpt-6-astra` reported `deepseek/deepseek-v4-flash`
- a child requested with a bogus model id also ran as the parent, with no
  client-side validation error

## Change

Keep an explicit subagent model; only a call that omits the model inherits the
parent. This mirrors what the router already does for `create_thread`.

There is no enum to police here. Codex ships `model` on the tool schema as a
plain string:

```json
{"type":"string","description":"Model override for the new agent. Omit unless an explicit override is needed."}
```

so `sanitizeSpawnAgentModel` never fires against a current client. Codex
validates the value itself and rejects anything outside its list:

```
Unknown model `totally-bogus-model-xyz` for spawn_agent. Available models:
gpt-6-astra, gpt-5.6-sol, gpt-5.6-terra, gpt-5.6-luna, gpt-5.5
```

An invented name therefore still cannot cross a billing boundary. The sanitizer
keeps its guard for a client version that does ship an enum: the item it clears
arrives without a model and inherits the parent as before.

## Testing

- `test/subagent-model-inherit.test.mjs` covers a kept explicit model, the
  absent/empty/non-string cases that still inherit the parent, and the
  wire-level preserve-or-inherit split driven through the real
  `flattenNamespaceTools` -> `buildNamespaceLookups` ->
  `rewriteNamespaceResponsePayload` path.
- `test/namespace-relay.test.mjs`: the existing pin test now asserts that an
  advertised override survives and that only an unadvertised one is pinned.
- `node --test` over the namespace-relay, namespace-relay-routing,
  subagent-completion, subagent-model-inherit and catalog suites: 259 pass, 0
  fail.
- `node scripts-check.mjs` passes.
- Live from a routed parent after the change, the router served child turns on
  `gpt-6-astra` and on `gpt-5.6-luna` per its own request log. Before the change
  the same calls were served on the parent.

### DIFF
--- a/.claude/skills/codex-subagents/SKILL.md
+++ b/.claude/skills/codex-subagents/SKILL.md
@@ -118,6 +118,19 @@ what Codex is allowed to choose.
 For parallel work, ask for several in one message so they run concurrently rather
 than in sequence.
 
+## Delegating from a routed parent to a native OpenAI model
+
+The router writes `expose_spawn_agent_model_overrides = true` into the managed
+`multi_agent_v2` block, so Codex offers a model override on `spawn_agent` for
+every session, routed parent included. Name the child model when you ask:
+
+> spawn a subagent on gpt-6-astra to audit src/router.mjs
+
+Codex checks the name against its own override list and rejects anything it does
+not offer, so an unavailable model fails before a turn is spent. The router
+keeps the override instead of pinning the child back to the parent's model; only
+a spawn call that omits the model inherits the parent.
+
 ## When it goes wrong
 
 **Model not offered in Codex.** Run the report. Not listed → `subagents set <slug> on`,
--- a/docs/HOW-IT-WORKS.md
+++ b/docs/HOW-IT-WORKS.md
@@ -395,7 +395,11 @@ unprotected, or left over for a model that is no longer eligible, the installed
 picker is republished to restore the coupled catalog/agent state. An unreadable
 or foreign Codex transport is never repaired speculatively.
 
-Only registry-proven models are advertised as native v2 spawn-agent overrides.
+Only registry-proven models are advertised as native v2 spawn-agent overrides,
+and an explicit spawn model is kept: a child runs on the model the operator
+picked from Codex's own list instead of being pinned back to the routed parent.
+Codex checks that value against its list before the call is dispatched, so an
+override cannot name a route the operator was never offered.
 The Settings tab (desktop panel and macOS tray) exposes two local accordions:
 **Subagent models** can withhold or re-enable proven models, while **Model
 picker** controls visibility. Local settings never promote an unverified model
--- a/src/namespace-relay.mjs
+++ b/src/namespace-relay.mjs
@@ -607,13 +607,24 @@ function isSubagentSpawnCall(item) {
   return item.namespace === undefined && item.name === `collaboration${NAMESPACE_DELIMITER}spawn_agent`;
 }
 
-// Inject the session model into local create_thread calls that omitted it and
-// pin every spawn_agent call to the routed parent. A parent not offered by the
-// client then fails closed at tool validation instead of silently crossing a
-// billing boundary.
-// `model` is the routed session's model (route.slug). Returns a rewritten
-// item when a local create_thread carries no explicit model or a spawn_agent
-// model differs from the routed parent; otherwise returns the item untouched.
+// Inject the session model into local create_thread calls that omitted it, and
+// into spawn_agent calls that carry no model of their own.
+//
+// An explicit subagent model wins. Codex ships the override as a plain string
+// on the tool schema rather than a schema enum, and validates the value against
+// its own advertised list, so anything that reaches here is the operator's own
+// delegation choice and not a value the parent model invented. Rewriting it
+// back to the routed parent discarded that choice and made a cross-provider
+// subagent impossible from any routed session, which is the one thing
+// `expose_spawn_agent_model_overrides = true` exists to allow.
+//
+// `sanitizeSpawnAgentModel` still drops a value outside the advertised set when
+// a client version does ship one, so that guard is unaffected: the item it
+// clears arrives here without a model and inherits the parent as before.
+//
+// `model` is the routed session's model (route.slug). Returns a rewritten item
+// only when the call carries no model of its own; otherwise returns the item
+// untouched.
 export function injectSessionModelForSpawnCalls(item, model) {
   if (!isSpawnModelCall(item)) return item;
   if (typeof model !== "string" || !model) return item;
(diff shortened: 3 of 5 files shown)


duolahypercho:
Thanks for this — the frustration it's addressing is legitimate, and the implementation is clean. But I can't take it as written, for a reason that goes beyond the policy question.

## The safety argument is self-cancelling

The new comment rests the case on `sanitizeSpawnAgentModel` still policing the client's advertised set, while also asserting that "Codex ships the override as a plain string on the tool schema rather than a schema enum."

Those two statements cannot both hold. `sanitizeSpawnAgentModel` returns the item untouched when there is no set to check against:

```js
const allowed = lookups.spawnAgentModels;
if (!(allowed instanceof Set) || allowed.size === 0 || typeof item.arguments !== "string") {
  return item;
}
```

And `SPAWN_AGENT_MODELS` is only populated when the schema carries a `const`/`enum`/`anyOf`. **No enum means no guard at all.** Combined with the change here — `if (typeof args.model === "string" && args.model) return item;` — an arbitrary string reaches Codex verbatim. Verified against a client declaring `model: { type: "string" }`:

```
advertised set: undefined
result model:   totally-made-up/whatever-9000
```

Same input on `main` yields the routed parent. So in exactly the configuration the comment describes as normal, there is no validation left.

## It also reverses a policy landed deliberately

`ea3dbb59` (#629) and `b944249d` (#632) pin routed subagents to the parent specifically so a spawn cannot silently cross a provider or billing boundary. `b944249d` added one regression case — a `spawn_agent` naming `gpt-5.6-terra` while the parent runs `opencode-go/deepseek-v4-flash` — and this PR renames it to `advertisedCrossProvider` and flips the expectation.

Worth being precise about why membership in the enum isn't a defence: that set is Codex's v2-eligible override window, built from `multi_agent_version` (`src/catalog.mjs:727`). It says nothing about provider. So "in the enum" has never implied "on the same credential", and there's no downstream affinity check when the child request lands — I grepped for one.

## What would make this mergeable

1. **Keep the parent pin when `spawnAgentModels` is absent or empty**, so the guard is never silently a no-op. That alone closes the hole above and is independent of the policy debate.
2. **Constrain the override to the parent's own provider**, or show that Codex rejects an unadvertised override before dispatch. Either gives the cross-provider case a real answer.
3. **Update `skills/codex-router/SKILL.md`** — not just `.claude/skills/`. That file is installed into `$CODEX_HOME/skills` by `src/skills-install.mjs` and verified by `doctor`, and it still tells the routed model that subagents "are always pinned to the routed parent's model because their tool arguments are model-generated and must not silently cross a provider or billing boundary." Right now the prompt context would contradict the behaviour.
4. Reference #629 and #632, and add the `AGENTS.md` / `CHANGELOG.md` entries, so the reversal is on the record as intentional.

Also worth adding a test for the no-enum client — that's now the only configuration where an invented model reaches Codex unchecked, and nothing covers it. Note `a wire model outside the advertised enum is still pinned to the parent` passes on both branches, so it can't distinguish them.

Your two new tests are real — both fail against `main`, so the behaviour is genuinely pinned. And `flattenNamespaceChild` and the custom-tool identity registration are untouched, which I checked.

duolahypercho:
Agreed on the problem, and it is worse than the description suggests: because
`sessionModel` is `route?.slug` (src/router.mjs:4331), the pin fires for every
routed parent, so choosing a subagent model from a routed session has never had
any effect — the child always ran on the parent's model. From a native GPT
parent the override already worked, since `injectSessionModelForSpawnCalls`
returns early on an empty session model. So this is the routed case, which is
the case the `codex-subagents` skill is about. Worth fixing.

One narrowing before I merge it.

`sanitizeSpawnAgentModel` is conditional: src/namespace-relay.mjs:2024 returns
the item untouched when `lookups.spawnAgentModels` is absent or empty. On a
client that advertises no list, nothing validates `args.model`, and that
argument is written by the parent model rather than chosen by the operator. As
written, this PR turns that case from fail-closed into pass-through, which is
what the old comment was protecting against.

Please keep an explicit model only when it was actually validated — when
`spawnAgentModels` is a non-empty Set — and fall back to the current pin when
there is no advertised list to check against. That preserves the delegation this
PR is for without widening the unvalidated path. A test for each side of that
branch, and a CHANGELOG entry, and I will merge it.

duolahypercho:
Correction to my previous comment — do not implement it as written yet.

I asked you to keep an explicit model only when `spawnAgentModels` is non-empty.
That assumed Codex advertises the override as a schema enum, which is what this
repo's own fixtures assume (`clientRoutedTools()` ships
`anyOf: [{ type: "string", enum: [...] }, { type: "null" }]`). Your PR says the
opposite — a plain string — and Codex 0.153.2's binary carries the available
models as prose, in a tool description ending "Available models: ", with no
enum I can find next to the `model` property.

If you are right, `schemaStringValues` (which only reads `const`/`enum`)
collects nothing, `SPAWN_AGENT_MODELS` is never populated, and
`sanitizeSpawnAgentModel` is already a no-op in production. Two consequences:

1. My narrowing would make this PR do nothing at all — "keep only when
   validated" would never keep anything.
2. The pin you are removing is currently the *only* thing standing between a
   parent-model-authored `model` argument and the spawn, which makes the
   fail-closed question more live than I first said, not less.

So before either of us writes code: can you confirm which shape the live client
sends, by capturing the actual `spawn_agent` tool definition the router receives
from a Codex with `expose_spawn_agent_model_overrides = true`? If it carries an
enum, your patch is fine as-is and the sanitizer already covers it. If it is a
plain string, we need a router-side allow-list — an explicit model accepted when
it resolves to a model the router knows, pinned otherwise — rather than a
blanket accept.

The underlying bug is real either way and I want this fixed: with
`sessionModel = route?.slug`, a subagent model chosen from a routed parent has
never had any effect.

duolahypercho:
Merged as-is. Ignore both of my review comments above — the bug this fixes is
real and has been silently sending every subagent spawned from a routed parent
to the parent's own model, and that outweighs the fail-closed question I was
circling.

The open question is now a follow-up, not a blocker: if the live client
advertises the override as a plain string rather than an enum, the router has no
allow-list to check an explicit model against. If you do happen to capture that
tool definition, post it here and I will open an issue for a router-side check.

Thanks for the fix, and for the skill and HOW-IT-WORKS updates alongside it.

PR #743, opened 2026-09-13
Set default-tier model aliases in Claude Code launcher

Hunselvm:
## Problem

Claude Code sessions routed through the router keep dying inside spawned subagents with:

```
Agent terminated early due to an API error: There's an issue with the selected model
(claude-opus-5[1m]). ... (error type model_not_found, HTTP 404, model sent to the
API: claude-opus-5)
```

The existing `CLAUDE_CODE_SUBAGENT_MODEL` pin does **not** cover this case. Two distinct agent paths ignore it:

- **Built-in agents** (Explore, Plan, ...) — they resolve `opus`-tier requests through the default-tier alias (`ANTHROPIC_DEFAULT_OPUS_MODEL`), not the subagent override.
- **User-defined agents** whose frontmatter pins `model: opus` — same alias path.

With neither the alias nor the subagent var matching, Claude Code falls back to the literal Anthropic id `claude-opus-5`. The router serves neither, so the request goes upstream verbatim and 404s — killing every spawned agent of that type while the main loop keeps working (its model is pinned correctly). Verified by spawning an Explore agent and a general-purpose agent in the same session: general-purpose succeeded, Explore 404'd.

## Fix

Set the three default-tier aliases to `catalog.defaultModel` alongside the existing pins, so every opus/sonnet/haiku-tier resolution inside a routed session lands on a served model.

Verified locally by importing `claudeRouterEnvironment()` with the real caller secret — all six model vars now emit the routed default.


duolahypercho:
Good catch on the underlying problem — built-in agents and `model: opus` frontmatter resolving to a literal `claude-*` id the router doesn't serve is a real 404, and the fix is in the right place.

Two things I'd want changed before merging, both about respecting choices the user already made.

**1. The new block ignores the gates the existing one honours.** `ANTHROPIC_MODEL` is set only when `!explicitModel(args) && !saved.startsWith("codex_router/anthropic/")`, and there's a test pinning that intent (`an explicit or already-saved router model wins over the launcher default`). The new block has no such gate, so the session splits across two models. Verified with `--model codex_router/anthropic/deepseek/test` plus the same `settings.model`:

```
ANTHROPIC_MODEL:               undefined                                  <- user's deepseek choice wins
CLAUDE_CODE_SUBAGENT_MODEL:    codex_router/anthropic/openai/gpt-test
ANTHROPIC_DEFAULT_OPUS_MODEL:  codex_router/anthropic/openai/gpt-test
```

A user who picked deepseek via `/model` gets every subagent and tier call on the catalog default instead, silently. Honouring at least `saved` would keep the session coherent.

**2. Caller-supplied values are clobbered.** All five assignments are unconditional over `{...environment}`, so an incoming `ANTHROPIC_SMALL_FAST_MODEL=user-picked-cheap` comes back as the catalog default. Someone who deliberately points the background tier at something cheap loses it with no warning. `env.X ??= …` fixes both the 404 and this.

Related: the small-fast/haiku tiers exist for cheap background work, and `claudeDefaultModel` sorts on `priority`, which routed models don't carry — so the winner is effectively the alphabetically-first routed slug. That selection is pre-existing, but this widens its blast radius from "the default session model" to "every background and subagent call."

**On the good side**, and I checked these specifically: nothing is persisted into a Claude-owned file (all process-local, so the AGENTS.md rule holds), and it does **not** hardcode a model — `catalog.defaultModel` is derived from the user's own selected, credentialed models, and drift is already reported by `claudeCatalogDrift` / `claude doctor`.

**No test fails if this is reverted** — the suite passes unchanged on both branches. Four assertions on `claudeRouterEnvironment` would cover it: all five set with empty settings/args (the fix itself); the subagent and tier variables tracking a different saved model; a preset `ANTHROPIC_SMALL_FAST_MODEL` surviving; and none set when `catalog` is undefined. The middle two fail on the current diff, which is exactly why they're worth having.

One more, non-blocking: I couldn't confirm from this checkout that all five variable names are honoured by the installed Claude Code — none appear elsewhere in the tree, and the repo already carries a version caveat for `ANTHROPIC_DEFAULT_MODEL` ("only exists in Claude Code 2.1.236+"). A version-gated variable would be inert rather than harmful, so this is a docs gap, not a defect — but worth a note beside the others.

duolahypercho:
Thanks — the diagnosis is right: subagents and the default-tier aliases do fall
through to literal Anthropic ids the router does not serve. A few things before
this can land.

**1. Fold this into the block above it.** The new `if (catalog?.defaultModel)`
sits directly beneath a block guarding on the same condition. One block.

**2. The new block skips the guards the existing one respects.**
`ANTHROPIC_MODEL` is only set when `!explicitModel(args)` and the saved
`settings.model` does not start with `codex_router/anthropic/`. The alias vars
honour neither, so `claude --model X` runs the session on X while every subagent
and every built-in agent goes to `catalog.defaultModel` instead. A user with an
Anthropic model pinned in settings gets the same split. Whatever the intended
behaviour is, it should be the same decision for the session and its children.

**3. `ANTHROPIC_SMALL_FAST_MODEL` should not take the default model.** That
variable exists for cheap background work — titles, summaries. Pinning it to the
routed default puts all of it on the heavy model. If the reasoning is that the
router serves nothing small and fast, say so in the comment and pick the
cheapest served model rather than the default.

**4. Tests and CHANGELOG.** `test/claude-code-launcher.test.mjs` already covers
`claudeRouterEnvironment`; please add cases for the new variables, including the
`--model` and saved-settings interactions in point 2. A CHANGELOG entry under
Unreleased matches the rest of the repo.

CI is green.

PR #744, opened 2026-09-13
fix: background Windows job owners are kept windowless

brianbell-x:
Background Windows commands requested with `stdio: "inherit"` could allocate a console for their PowerShell Job owner. Hiding was previously disabled for every inherited stream; even after the TTY correction, libuv's `CREATE_NO_WINDOW` was still omitted whenever a standard descriptor was inherited. The original child-only probe therefore missed the owner-console gap reported in review.

For hidden Windows calls without a TTY, three fresh pipes are now used and stdin, stdout, and stderr are forwarded. Output is forwarded as raw bytes through asynchronous descriptor writes: synchronous Windows `process.stdout.write()` was demonstrated to block deadlines when downstream was left unread. Partial writes are completed, output is drained before successful settlement, and early stdin closure is handled.

Direct inheritance is retained for actual TTYs, explicit `windowsHide: false`, and POSIX. The existing Job Object, barrier, and process-tree cleanup mechanisms are preserved. The static PowerShell guard's comment has been narrowed to its actual coverage.

Validation:

- On Windows 11 with Windows Terminal Preview configured as the default terminal, zero console-window handles were verified for both the contained command and its actual PowerShell owner. The exact new owner regression was failed against the previous PR commit `4cc5a08`, as expected. A visible Terminal flash was not independently established by that probe.
- Native stream regressions were passed on Node 22.23.1 and Node 24.19.0: 3 MiB binary output per stream with delayed readers and immediate caller exit; 2 MiB input checksum; early input closure; and deadline rejection with target retirement before unread output was drained. Restoring synchronous writes caused the deadline regression to fail.
- Partial writes, zero-progress writes, cancellation after child close, and late write errors were covered by additional deterministic tests. Success, failure, abort, deadline, and abrupt-owner-exit cleanup were rechecked.
- The 16-file Node 24 Windows/desktop/process/service/install/control suite was passed: **224 passed, 19 platform skips, 0 failures**. Repository syntax checks were passed. Two independent agents were used for implementation critique and native runtime verification; no actionable findings remained.

An already pending OS write cannot be cancelled without closing the caller's global descriptor. Timeout rejection and child retirement are bounded; natural parent exit can still await that write until downstream is drained or closed. TTY detection remains based on `isTTY`; Git Bash/MSYS2-specific console prompts were not validated.

The direct synchronous repair launches in `doctor.mjs` and `control.mjs::runDoctor`, plus scheduler/service-launch changes, remain outside this shared-runner patch. The proposed coordination is #744 first, followed by #674 rebased onto it. [Fresh CI for `2d589d7`](https://github.com/duolahypercho/codex-router/actions/runs/34811966553) has passed on Windows, macOS, and Linux, including both Electron jobs and the native macOS build. Deterministic provider-discovery checks have also passed; the live catalog audit and Homebrew install jobs were skipped.

Related: #565, #571, #674.


duolahypercho:
Reviewed line by line. The direction is right, and I confirmed it does not regress the things #674 got wrong — teardown, job objects, `taskkill /T`, PID validation, the barrier protocol and the POSIX branch are all untouched, and your behavioural test genuinely fails when reverted (`inherited pipes: PowerShell owner — false !== true`). Running `process-tree.test.mjs` on macOS against both trees, the same four owner-signal tests flake under each, and everything green on `main` stays green.

**The gap: this does not remove the window on the configuration that reported the bug.** The node-level spawn still passes `stdio: "inherit"`, and libuv only ORs in `CREATE_NO_WINDOW` when *no* stdio slot is inherited — otherwise it degrades to `SW_HIDE`, which Windows Terminal ignores. So on Win11 + WT (the default terminal via DefTerm handoff), the PowerShell job owner still gets a console and still shows it.

What this PR *does* fix is real and worth having: the **contained** process now gets a true `CREATE_NO_WINDOW` through the PS1 payload, and hosts still on classic conhost are fixed outright.

Concretely still flashing on a Win11 desktop: Control Center → `control.mjs` → `restartRouterServiceIfInstalled` → `invokeService(["restart"], {stdio: "inherit"})` (`src/router-restart.mjs:246`), and `ensureNodeDependencies`' `npm ci` (`src/node-dependency-install.mjs:70`). The complete fix is to stop handing libuv inherited *fds* when there is no terminal — spawn with pipes and forward the streams to `process.stdout`/`process.stderr`. The PowerShell owner duplicates whatever `GetStdHandle` gives it, so the contained command's I/O is unaffected.

Two related notes:

**The Windows test asserts the half that was already going to be hidden.** It probes `GetConsoleWindow()` inside the *contained* process, which proves the flag reached `CreateProcess` — genuine progress — but says nothing about the owner, which is the process that allocates the visible console. It also runs on `windows-latest`, which is conhost, so CI is structurally blind to the Windows Terminal case. Probing the owner, or asserting its creation flags carry `CREATE_NO_WINDOW`, would close that.

**The guard's new exemption comment is not accurate for the set it scans.** It says inherited-stdio calls are skipped "because their terminal attachment must be determined at runtime … that distinction is covered in process-tree.test.mjs" — but `process-tree.mjs`'s own `spawn` is not in the scanned population at all (its argument text never matches the powershell pattern the scan requires). The two tests cover disjoint call sets.

The live example that falls in that hole: `src/doctor.mjs:282` runs `spawnSync("powershell.exe", [... install.ps1 ...], { stdio: repairStdio })` with `repairStdio` resolving to `"inherit"` and **no `windowsHide` at all** — reachable from the Control Center repair button, i.e. from a console-less Electron parent. Exempt from the guard, and untouched here.

One low note: `isTTY` is a proxy for "a console exists", and it is wrong in both directions on Windows — under Git Bash/MSYS2 the std streams are named pipes, so an operator genuinely at a terminal reads as `false`. Output still flows (stdio is still inherited), but a command that prompts on `CONIN$`/`CONOUT$` rather than stdin — `npm ci` hitting a git or ssh passphrase prompt — would render into a hidden console and stall until the deadline fires. Narrow and deadline-bounded. Also worth reordering the check to `stdout || stderr || stdin`, so a long-lived service process does not instantiate `process.stdin` merely to read `isTTY`.

Happy to merge this as a strict improvement if you would rather land it now and extend separately — just say which you prefer.

duolahypercho:
@brianbell-x this overlaps with #674 from @jaakjuu1: both fix the same Windows
console root cause, and both touch `src/process-tree.mjs` and
`test/process-tree.test.mjs`. #674 additionally covers `src/control.mjs`,
`src/service-windows.mjs`, `src/start.mjs` and `src/windows-job-host.pyw`.

This PR is green everywhere, and the TTY check plus the `GetConsoleWindow()`
probe is a clean statement of the problem — an inherited pipe from a desktop
host is not a terminal, and `windowsHide` should still apply.

Rather than land two fixes to the same function, could the two of you settle who
takes which half — #674 rebasing onto this check and keeping the surfaces this
does not touch, or this folding into #674? Whichever you agree on, I will merge
in that order.

brianbell-x:
@duolahypercho @jaakjuu1 The proposed order is **#744 first, followed by #674 rebased onto it**. The shared `runProcessTree` pipe relay and owner/child console regressions are being completed in #744. The separate scheduler and service-launch work can then be retained in #674 without a competing change to the shared function. Would that split be acceptable?

The owner-console gap was confirmed against the previous #744 commit: the contained command had `GetConsoleWindow() == 0`, but a nonzero handle was obtained after attachment to the actual PowerShell parent's console. With the revised three-pipe launch, successful attachment to that same owner now returns a zero window handle as well. This was checked on Windows 11 with Windows Terminal Preview configured as the default terminal.

The stream relay is also being checked for full stdin forwarding, binary output, early input closure, and timeout cleanup with a stalled output reader. A Windows-specific synchronous `process.stdout.write()` stall was caught during review; asynchronous descriptor writes are being used so the operation deadline remains responsive.

The separate `doctor.mjs` and `control.mjs::runDoctor` synchronous repair launches are still outstanding. Their safety is not established by the inherited-stdio exemptions in the source scan, and the scan comment is being narrowed accordingly. They should be covered separately from the shared-runner fix, with installer/service lifetime behavior preserved.

The completed patch and exact validation results will be added here shortly.

brianbell-x:
@duolahypercho @jaakjuu1 The shared-runner revision has been pushed as `2d589d7`.

- Hidden Windows inherited I/O is now relayed through three fresh pipes, so inherited descriptors are no longer passed to libuv for the PowerShell owner. Actual TTY inheritance and explicit `windowsHide: false` are preserved.
- Both the actual owner and contained command are now probed. The exact new owner test was failed against `4cc5a08`, as expected, and passed with the relay on Windows 11 with Windows Terminal Preview configured as the default terminal.
- Binary output, stdin, early input closure, delayed readers, partial writes, and cancellation are covered. A real deadline stall caused by synchronous Windows stdout writes was reproduced on Node 22 and 24; asynchronous descriptor writes were required to resolve it. The committed stalled-reader regression was failed again when synchronous writes were restored.
- The expanded 16-file suite was passed on Node 24: **224 passed, 19 platform skips, 0 failures**. Syntax checks and independent review/runtime verification were completed. The four native stream cases were also passed on Node 22.

The PR description and source-scan comment have been corrected. The earlier zero-window claim was established only for the contained command; it was insufficient for the owner. With the new probe, both handles are zero. Timeout rejection and target retirement were verified before downstream draining; an already pending OS write can still delay natural parent exit until the reader is resumed or closed.

The proposed landing order remains #744, then #674 rebased onto it for the separate service/scheduler work. The direct synchronous repair launches in `doctor.mjs` and `control.mjs::runDoctor` remain outstanding and are explicitly excluded from this PR's claims.

Updated after the CI run: [fresh CI for `2d589d7`](https://github.com/duolahypercho/codex-router/actions/runs/34811966553) has passed on Windows, macOS, and Linux, including both Electron jobs and the native macOS build. Deterministic provider-discovery checks have also passed; the live catalog audit and Homebrew install jobs were skipped. The proposed landing order is still awaiting confirmation.

PR #745, opened 2026-09-13
Add v2 agent application: deepseek/deepseek-v4.1-flash (draft)

kouji648-hub:
## Route

- Routed slug: `deepseek/deepseek-v4.1-flash`
- Upstream model ID: `deepseek-flash` (route config: `gatewayModel = deepseek-v4-1-flash`)
- Router version and test date: 0.5.1, 2026-09-14

## Status

`draft`. This PR makes no v2 claim and intentionally does **not** change any
registry `multiAgentVersion`. Two of the five checks remain open (4 and 5).

## Evidence recorded so far

| Check | Result | Redacted summary |
| --- | --- | --- |
| Official model identity | pending | Still need the official documentation and catalog evidence for the exact upstream model. |
| Streaming Responses | passed (local) | A streamed turn through the installed router completed with status 200 and a terminal completion event on 2026-09-14. |
| Forced function call | passed (local) | A forced shell tool call returned the requested tool name with valid JSON arguments and the exact marker string once. |
| Encrypted relay | pending | Requires a native Codex parent delegating to this route. |
| Marker-return spawn | pending | Same prerequisite. |
| Same-thread follow-up | pending | Same prerequisite. |

## Why checks 4-5 are open

The route is declared `multiAgentVersion: v1` by the registry, so a native
Codex parent cannot delegate to it on our machine. Completing those two checks
needs the native collaboration probe. Once they have been reproduced and the
table is filled in, we are happy to follow up with the registry change
(`status: accepted` plus `multiAgentVersion: "v2"`) in a separate PR.

## Reviewer reproduction

1. Install the router with the DeepSeek provider enabled and this route selected.
2. Run one streamed turn and confirm a normal completion event.
3. Force a single tool call and confirm the requested tool name, valid JSON
   arguments, and the exact marker return.
4. Nominate the route for the native collaboration probe, delegate a child from
   a native Codex parent, and confirm the marker return.
5. Send a same-thread follow-up to that child and confirm the second marker.

No API keys, bearer tokens, raw prompts, response bodies, or encrypted payloads
are included in this PR.


duolahypercho:
Merging this as a draft record — that is the normal resting state for an application here, not a holding pattern. Five of the six applications already on `main` are `draft`, including three DeepSeek routes from 2026-08-22.

Verified on your head commit (`06849b5a`) in a clean worktree: `node scripts/check-v2-agent-applications.mjs` reports `v2-agent applications valid (7)`, and the artifact adds nothing outside `v2_agent/deepseek/deepseek-v4.1-flash/` — no registry change, no v2 claim.

You were right not to attempt checks 4-5. Per `v2_agent/README.md` the reviewer reproduces the marker-return and same-thread steps, and `src/subagent-proofs.mjs` describes the `candidate` state as exactly what you reached: the low-cost stream/tool probe passed, so submit the redacted application for the native encrypted-relay proof. There is no contributor-side path past that point, and there is not meant to be.

Two notes:

1. Your PR description's table still shows `Official model identity | pending`, but commit `06849b5a` updated the artifact to `passed (local)`. The artifact is the one that counts, so this is cosmetic — flagging it only so the description does not mislead a later reader.

2. Your identity evidence is more interesting than the check it satisfies. You recorded that `api.deepseek.com/models` returns exactly two ids, `deepseek-flash` and `deepseek-v4-pro`. `main` ships five `deepseek/*` routes, and three point at ids not in that list: `deepseek-v4-flash`, `deepseek-v4-flash-vision-exp`, and `deepseek-reasoner`. If your catalog read was not account-scoped, that is live drift in the shipped config, and the `Live provider catalog audit` job was `skipping` on this PR so it would not have caught it. If you still have that catalog response, could you open a separate issue with the id list and how you fetched it? That is worth more than this application.

Thanks for following the process as written — checks 1-3 with redacted evidence and no registry change is exactly the right shape.

PR #746, opened 2026-09-14
fix: skip remote compaction v2 while under autoCompact

omarjson:
Skips unnecessary lossy compaction when the request is still inside the route's autoCompact budget. Adds a helper, wires it into the router, and covers it with focused tests.

Verification: 
ode --test test/compaction-limit.test.mjs` passes 5/5 and 
pm run check passes.


duolahypercho:
Thanks for this — the motivation is reasonable, but CI is red on this revision and I think the approach needs a rethink beyond just getting it green.

### What's failing

`test (ubuntu-latest)`, `test (macos-latest)` and `test (windows-latest)` all fail the same two routed-compaction tests in `test/routing.test.mjs`:

```
✖ router synthesizes routed compaction and safely replays it to native models
    actual: 'reasoning'   expected: 'compaction'
✖ compaction never treats reasoning as final text and falls back to chat content
    actual: undefined     expected: 'checkpoint'
```

Both pass at this PR's parent (`0964eefd`). `npm run check` doesn't cover `routing.test.mjs`, which is why this looked green locally.

### Proximate cause

`shouldSkipRemoteCompactV2` returns `true` when `estimatedTokens === undefined`, but `estimateInputTokens` returns `undefined` for any body under `MIN_ESTIMATED_INPUT_TOKENS` (1,000) — see `src/response-usage.mjs`. So *every small* compaction request is skipped, not just ones under `autoCompact`. The fifth test pins the missing-`autoCompact` case, but nothing pins the small-body case.

### The larger objection

Fixing that branch would move the tests' payloads onto a different path, but I don't think it fixes the design. Codex sends `compaction_trigger` and expects a `compaction` item carrying `encrypted_content` back (`handleRoutedCompaction` in `src/router.mjs`). Stripping the trigger makes the router answer the previous user turn again as an ordinary response, and Codex has nowhere to put that — a genuinely under-budget session would hit the same thing. If the goal is avoiding a lossy summary, I think the router still has to answer *with* a compaction item — perhaps a cheap near-lossless checkpoint, though I haven't checked how well that fits what `summarize` currently does.

One more thing to weigh: the `autoCompact` being compared against is the same number the router advertises to Codex as `auto_compact_token_limit` (`src/catalog.mjs`). Codex triggered on provider-reported token counts, so a byte-ratio estimate disagreeing isn't strong evidence that the request is spurious — and `src/response-usage.mjs` is explicit that erring low costs the turn while erring high only costs a summary.

### Minor

`estimateInputTokens` runs twice on the same body (once inside the helper, once for the log line). Compaction bodies are the largest the router handles, so compute it once and pass the value through.

PR #749, opened 2026-09-14
feat(macos): install signed tray release without Xcode

jiemocoder:
## Summary
- publish an optional signed and notarized universal macOS tray app from the release workflow
- download, checksum, validate, and install the matching GitHub Release asset before falling back to a local Xcode build
- bind release metadata to the router version and tray source fingerprint
- document Apple signing secrets and the fallback behavior

## Validation
- targeted Node test suite: 91 passed, 0 failed, 1 skipped
- npm run check
- actionlint .github/workflows/release.yml
- git diff --check

The macOS release job skips cleanly until the repository configures the documented Apple Developer signing and notarization secrets.


duolahypercho:
Thanks for this — the structure is right: the journaled swap is preserved, symlinks are rejected, and leaving hardened-runtime signing to electron-builder instead of a `codesign --deep` re-sign is the correct call. Two things need to change before it can merge.

**1. The release repository defaults to this fork.**

`src/macos-tray-release.mjs:19` and `scripts/download-macos-tray-release.mjs:26` both default to `jiemocoder/codex-router`. Upstream is `duolahypercho/codex-router` (cf. `scripts/update-homebrew-formula.mjs:13`), and `test/macos-tray-release.test.mjs` asserts the fork URL as expected behavior. With the version bumped to 0.5.2 and a `v0.5.2` release already published on the fork, every upstream macOS install would resolve its checksums there.

The surrounding verification doesn't substitute for the right slug:

- `'=anchor apple generic and certificate leaf[field.1.2.840.113635.100.6.1.13] exists'` accepts any Developer ID Application certificate from any team. It needs `and certificate leaf[subject.OU] = "<TEAM_ID>"` in both places it appears (the workflow's notarize step and the downloader).
- `ModelRouterTraySourceFingerprint` is written into `Info.plist` by whoever builds the app, and its expected value is derivable from the public sources. It's sealed by the builder's signature, so it can't be tampered with after signing — but it doesn't attest that the shipped binary came from those sources.

Since clients can't read `MACOS_TEAM_ID`, please export the team identifier as a constant in `src/macos-tray-release.mjs` and have the release job assert `MACOS_TEAM_ID` matches it, so a mismatch fails the release rather than shipping an app every client will reject.

**2. "No package published" and "no network" must stay fallback-eligible.**

`scripts/download-macos-tray-release.mjs:46` throws for any non-200 on SHA256SUMS, and line 43 throws when `curl` itself fails. Only a 404 on the app asset returns `RELEASE_UNAVAILABLE`. Combined with `bin/model-router-tray:368`, that means an unreleased version or an offline machine refuses the local build entirely.

That's the normal state of `main` between tags — after this merges, `v0.5.2` won't exist upstream until it's cut — and it's the permanent state of every branch and dev checkout. The failure is non-fatal to the router (`bin/install` and `src/setup.mjs` catch it), but the companion silently stops installing and updating, and the guidance printed blames a missing Xcode rather than the real cause.

Suggested cut: exit `RELEASE_UNAVAILABLE` for anything that means the package can't be *obtained* — transport failure, 404 on either file, 5xx. Keep the hard refusal for genuine integrity failures: checksum mismatch, a downloaded asset with no `SHA256SUMS` entry, signature or Gatekeeper rejection, metadata/fingerprint mismatch, symlinks, or unexpected archive contents.

**Smaller items**

- `release` now `needs: unified-app-macos`, so a `notarytool submit --wait` failure blocks the Windows and Linux release too. The credentials-absent path is fine, but consider whether a macOS outage should gate the whole release.
- `verifyReleaseMetadata` runs `PlistBuddy` against the extracted bundle before `codesign --verify`. Move signature and Gatekeeper assessment ahead of reading anything out of it.
- `MODEL_ROUTER_PORTABLE_APP=1` drops `ModelRouterSourceRoot` and `router-root` — necessary for a portable build, but it removes the sealed checkout binding that `build-macos-tray-app.sh` documents as deliberate, leaving the Swift host to resolve the checkout from the install manifest or `~/.local/share/codex-router`. Worth calling out in `docs/MACOS-TRAY.md`.
- `CHANGELOG.md` is missing a blank line before the `- **Routed coding clients…**` entry.

CI hasn't run here — both **CI** and **Provider model discovery** are at `action_required` — so the test results in the description are unverified upstream. I'll approve the workflow runs once the two blockers are addressed.

duolahypercho:
This installs signed binaries, so it got a closer security review. It can't merge as it stands. Blocking issues:

1. **Downloads come from a fork by default.** The default release repository is `jiemocoder/codex-router`, in both `scripts/download-macos-tray-release.mjs` and `src/macos-tray-release.mjs`. Installers must only pull from `duolahypercho/codex-router`.
2. **The signature check accepts any developer.** The requirement `anchor apple generic and certificate leaf[field.1.2.840.113635.100.6.1.13] exists` passes any Developer ID, and `spctl` only confirms that some developer notarized the app. The check needs to pin the project's Team ID (for example `certificate leaf[subject.OU] = "<TEAMID>"`). Nothing else makes up for this:
   - `SHA256SUMS` comes from the same release as the zip, so it only catches transfer corruption, not tampering.
   - The bundle id and fingerprint are read from the app's own `Info.plist`, which whoever builds the app controls.

   As it stands, anyone who controls the release source, or can set `CODEX_ROUTER_RELEASE_REPOSITORY`, can ship any notarized app.
3. **Tray updates fail hard with no fallback.** Only a 404 on the zip falls back to a local build. The update exits 1 with no fallback when:
   - the release is missing (the `SHA256SUMS` download 404s first),
   - the machine is offline,
   - the checkout's tray sources differ from the tagged release.

   In each case `bin/model-router-tray` then refuses the local build, so dev checkouts and anyone running `main` between releases can't update the tray.
4. **Housekeeping:**
   - `curl` has no `--max-time`.
   - `CHANGELOG.md` is edited directly; please use a `changelog.d/` fragment.
   - The version bump to 0.5.2 conflicts with the `package.json` and `package-lock.json` files on `main`.
5. **Scope.** The looser `legacy_user_bundle_is_managed` check and the release workflow's need for Apple signing secrets both need the maintainer's decision.

---

Issue #751, opened 2026-09-14
Windows: Control Center tray supervision repeatedly relaunches Codex Router.exe through bursts of visible PowerShell windows


Scrieda:
Summary

On Windows 11 (build 10.0.26200), the Control Center companion repeatedly spawns a
burst of `powershell.exe` processes - each with its own `conhost.exe`, so each one
is a visible console window - and relaunches `Codex Router.exe` from them.

Observed at 12:59:41-12:59:43 local time (WIB, UTC+7): six `powershell.exe`
processes within about two seconds. The same pattern repeats later (for example
13:04:43), so it is not a one-off during install.

## Environment

- codex-router 0.5.1, default Windows install at `C:\Users\<user>\AppData\Local\codex-router`
- Windows 11, OS build 10.0.26200
- Node.js v24.16.0
- Codex desktop app 26.908.40834
- Router service itself stayed healthy the whole time (litellm/python PIDs unchanged
  from 12:33 onward, model requests returning 200)

## Evidence

Process-start sampling using `Win32_Process` (CommandLine + ParentProcessId, 150 ms
poll). Times are local; the router log's `at=` stamps are the same events in UTC.

```text
12:59:41  conhost     (pid 8144)   <- powershell (pid 5632)
12:59:41  conhost     (pid 24256)  <- powershell (pid 15992)
12:59:41  powershell  (pid 5632)   <- Codex Router (pid 25260)
          | C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -NoLogo -NoProfile
            -NonInteractive -ExecutionPolicy Bypass -File C:\Users\<user>\AppData\Local\codex-router\codex-router.ps1 ...
12:59:41  powershell  (pid 15992)  <- Codex Router (pid 25260)  (same command line)
12:59:42  powershell  (pid 8300)   <- Codex Router (pid 25260)  (same command line)
12:59:42  powershell  (pid 3128)   <- ?
12:59:43  powershell  (pid 9988)   <- Codex Router (pid 25260)  (same command line)
12:59:43  powershell  (pid 13732)  <- Codex Router (pid 25260)  (same command line)
12:59:43  powershell  (pid 13216)  <- ?
12:59:43  Codex Router (pid 4540)  <- powershell (pid 9988)
          | "...\codex-router\apps\control-center\release\win-unpacked\Codex Router.exe" ...
12:59:43  Codex Router (pid 6484)  <- powershell (pid 8300)
[...]

PR #754, opened 2026-09-14
fix(windows): accept control's tray verbs in codex-router.ps1

duolahypercho:
The isolated half of #751.

## Problem

`control tray` prints its own usage line as `enable|disable|status|restart|refresh|rebuild` (`src/control.mjs:3002`) and maps the two lifecycle verbs onto the supervisor's spelling before calling it:

```js
const TRAY_COMMANDS = { enable: "install", disable: "uninstall", status: "status", restart: "restart" };
```

The Windows wrapper knew only the supervisor spelling (`codex-router.ps1:953`), so following that usage line on Windows — the documented way to take the companion out of Task Scheduler — died on `Unknown tray action 'disable'`.

## Change

Fold the two control aliases onto the supervisor verbs **before** the action is validated, so both surfaces accept the same words and dispatch the same transaction: `tray enable` reaches the same durable package-and-task replacement `tray install` does, and `tray disable` the same removal as `tray uninstall`. `companion` inherits it, since it forwards to `tray`. The unknown-action message names the aliases.

The test parses `TRAY_COMMANDS` out of `src/control.mjs` and both the alias table and the accepted-action list out of `codex-router.ps1`, then asserts every control verb resolves to the exact supervisor transaction control names for it — so a later rename on either side is caught, rather than one hard-coded spelling.

## Scope

This does **not** address the visible-console half of #751. That is the Control Center's 5-minute poll spawning seven PowerShell-wrapped reads per tick (`apps/control-center/src/App.tsx:316-319` → `electron/command-runner.mjs:900-906`), which #674 fixes by removing the wrapper. #744 fixes a sibling on the CLI side. This PR touches no file that either of those touches.

## Verification

```
node scripts-check.mjs                    → syntax checks passed
node --test test/tray-service-windows.test.mjs test/control.test.mjs \
  test/codex-router-cli.test.mjs test/installer-scripts.test.mjs \
  test/windows-parity.test.mjs test/windows-operations.test.mjs \
  test/codex-router-args.test.mjs test/update-target.test.mjs
  ✔ the Windows tray verbs accept every control tray command
  ℹ tests 147  ℹ pass 140  ℹ fail 0  ℹ skipped 7   (Windows-only suites)
```

Negative control — with the two alias lines removed, the new test fails and nothing else does (`pass 35, fail 1`).

**Authored on macOS: the PowerShell change itself is not executed by any test here.** It is covered by a text-parity test, and `test/codex-router-args.test.mjs` is Windows-only and skipped on this host. The change is a hashtable lookup with PowerShell's default case-insensitive comparer inserted ahead of the existing validator, touching no other branch, but it is unexecuted and deserves a Windows check before merge.


duolahypercho:
Line-by-line review: adopt.

`control tray` already maps enable→install and disable→uninstall. The Windows wrapper only accepted the supervisor spellings, so `tray disable` died on Windows. Folding the aliases before validation is the right place, and the test parses both `TRAY_COMMANDS` and the PowerShell table so a later rename fails.

CI is green on Windows. Merging.

PR #759, opened 2026-09-14
ci: cache the binaries electron-builder downloads

c8dhjp4tyv-bit:
## Why

The three packaging jobs download the Electron binary and electron-builder's own helper binaries on every run, so a bad response from those hosts fails the job outright. Three died that way inside one day:

| when | job | error |
| --- | --- | --- |
| 14 Sep 14:50 | [Electron control center (Windows)](https://github.com/duolahypercho/codex-router/actions/runs/34857987861/job/104022418711) | `Response code 504 (Gateway Time-out)`, right after `downloaded label=nsis-3.0.4.1.7z` |
| 14 Sep 14:53 | [Unified native macOS app](https://github.com/duolahypercho/codex-router/actions/runs/34857987861/job/104022418939) | `Response code 504`, right after `downloaded label=electron` |
| 14 Sep 01:16 | [Electron control center (Linux)](https://github.com/duolahypercho/codex-router/actions/runs/34795374753) | `Response code 500 (Internal Server Error)` |
| 14 Sep 15:36 | [Electron control center (Windows)](https://github.com/duolahypercho/codex-router/actions/runs/34863045029/job/104039870191) | `Response code 500`, again right after `downloaded label=nsis-3.0.4.1.7z` — **this PR's own first run** |

Both 504s are `HTTPError` thrown out of `got` inside `apps/control-center/node_modules`, three minutes apart on two different runners — the download host, not the build. All three cleared on a re-run.

That last one is this branch's own first CI run, and it is worth being plain about: **a cold cache does not protect anything**, so the first run was exposed exactly like any other. It did, however, demonstrate the second half of the problem — see below.

## What changed

`actions/cache/restore` + `actions/cache/save` around the two directories electron-builder already keeps, on all three packaging jobs, keyed on `apps/control-center/package-lock.json` so an Electron bump misses and everything else hits.

**Restore and save are separate steps on purpose.** I first wrote this as a single `actions/cache@v6`, and its post step came back `skipped` on the failing run:

```
9   failure  Package the native beta
...
24  skipped  Post Cache electron-builder downloads
```

The job had already fetched Electron and NSIS successfully before the 500. Discarding those leaves the next run starting cold, so a bad window repeats instead of closing itself. `actions/cache/save` with `if: !cancelled()` keeps what was downloaded even when packaging failed; the save is skipped on an exact-key hit, so a healthy run does not re-upload.

The directories come from `envPaths("electron"|"electron-builder", { suffix: "" }).cache` — read from `node_modules/env-paths/index.js` rather than assumed:

| | Electron binary | helper binaries |
| --- | --- | --- |
| Linux | `~/.cache/electron` | `~/.cache/electron-builder` |
| macOS | `~/Library/Caches/electron` | `~/Library/Caches/electron-builder` |
| Windows | `%LOCALAPPDATA%\electron\Cache` | `%LOCALAPPDATA%\electron-builder\Cache` |

## Verification

Measured locally on Linux against a cleared cache, with the same `--linux` flag CI uses:

- **Cold**: writes 120 MB to `~/.cache/electron` (`electron-v43.4.0-linux-x64.zip`) and 14 MB to `~/.cache/electron-builder` (`appimage-12.0.1.7z`, `7zip-linux-x64.tar.gz` — the same class of helper binary whose sibling NSIS immediately preceded the 504 on Windows). So both directories are the ones that carry the failing downloads.
- **Warm**: the whole packaging run takes 16 seconds and re-downloads nothing. I set the cached zip's mtime to 2020-01-01 before the run and it was still 2020-01-01 afterwards, so the file was read rather than refetched. The `appimage`/`7zip` download lines disappear from the log entirely.

What I could not verify is GitHub's own cache save/restore, or the macOS and Windows paths — I have no runner for either, and the paths above are read from `env-paths`' source rather than observed. If you would rather see it prove itself first, the safest read is the next few runs' "Cache restored from key" lines.

## One thing worth knowing

`ELECTRON_CACHE` and `ELECTRON_BUILDER_CACHE` — the usual advice for this — **do not move the Electron download here**. I set both to absolute, existing directories and the zip still landed in `~/.cache/electron` while both stayed empty. `@electron/get`'s `Cache.js` defaults its root to `envPaths('electron', { suffix: '' }).cache` and reads no environment variable at all, so the override never applies. I mention it because it is the first thing most suggestions reach for, and it would have been a silent no-op.

## Scope

CI only; no source or test change, and no CHANGELOG entry, matching the other workflow-only commits. It reduces exposure rather than removing it — a cold cache still downloads, so an Electron bump landing during an outage can still fail. If you want that closed too, a retry around the packaging step would do it, and I am happy to add one.


c8dhjp4tyv-bit:
CI has now demonstrated both halves of this on its own, which is better evidence than my local measurement.

**The cache works.** On the green run, the Linux job restored and then skipped the helper-binary downloads entirely:

```
Cache restored from key: electron-builder-Linux-994ffbb840...
  • downloaded      label=electron progress=100%
```

That is the whole download section. No `appimage-12.0.1.7z`, no `7zip-linux-x64.tar.gz` — the two lines that are present on a cold run, and the same class of binary as the NSIS fetch that preceded every 5xx here. (The `label=electron` line is logged on a cache read too; locally I confirmed the zip is not refetched by setting its mtime to 2020 and watching it survive the build.)

**And the gap this PR's split closes is not hypothetical — it happened here.** The Windows job on that same run came back cold:

```
Cache not found for input keys: electron-builder-Windows-..., electron-builder-Windows-
  • downloaded      label=nsis-3.0.4.1.7z progress=100%
  • downloaded      label=7zip-win-x64.tar.gz progress=100%
```

It was cold *because its previous run failed on the 500*, and the single-step `actions/cache` I first pushed skipped its post step (`24  skipped  Post Cache electron-builder downloads`). Linux and macOS had succeeded in that same run, saved, and so hit this time — `Save electron-builder downloads` is reported `skipped` on both, which is the exact-key-hit path.

With `actions/cache/save` and `if: !cancelled()`, the Windows job saved this time even though it had to download. A bad window now closes itself after one run instead of repeating.

Still true and worth repeating: this reduces exposure rather than removing it. A cold cache — a fresh Electron bump, or a cache eviction — still has to reach those hosts. If you want that closed too, a retry around the packaging step would do it and I am happy to add one.

duolahypercho:
Line-by-line review: adopt.

CI-only. Restore/save around the directories electron-builder already uses, keyed on the Control Center lockfile. Separate save with `!cancelled()` is the right answer to `actions/cache` skipping its post step on a failed package job. `cache-hit` is exact-key only, so a lockfile bump still downloads the new Electron zip; restore-keys only warm old versioned files.

All three packaging jobs passed on this PR. Merging.

Issue #465, opened 2026-08-26
Windows: router.mjs crashes with exit code 3221226505 (0xC0000409 STATUS_STACK_BUFFER_OVERRUN), suspect zstdDecompressSync


lengocanh2005it:
## Summary

`router.mjs` (the main process listening on `CODEX_ROUTER_PORT`, e.g. 4202) intermittently crashes on Windows with exit code `3221226505` (`0xC0000409` / `STATUS_STACK_BUFFER_OVERRUN`), a native-level crash with no JS stack trace. When it happens, every child process (litellm gateway, oauth forwarders) dies with it and the whole router stays down until something restarts it — Task Scheduler's own `RestartCount`/`RestartInterval` on the installed task did not bring it back on its own in my testing.

## Environment

- OS: Windows 11 Home Single Language 10.0.26200
- Node: v22.22.2 (`C:\nvm4w\nodejs\node.exe`)
- codex-router: `0.4.0-beta.4` (confirmed the relevant code paths below are unchanged as of current `main` / `v0.5.0` — `router.mjs`'s `decodeBody()`/`zstdDecompressSync` and `antigravity-oauth-forwarder.mjs` have no diff between my checkout and `origin/main`)
- Router run via `start-codex-router.cmd` under a Task Scheduler task (logon trigger)

## Observed pattern

Reproduced 3 times over one day, always the same exit code, always at the end of a long session with a lot of traffic:

```
[codex-router] Codex router exited (code=3221226505, signal=null).
```

- The request immediately before the crash always completed normally (`status=200`) — no single "poison" request visible in the log.
- All 3 crashes happened after a sustained streak of successful `antigravity-oauth/gemini-3.7-flash` requests, in sessions where `cached_tokens` had climbed well into the 100k–200k range (i.e. long-running sessions with large accumulated context).
- No stack trace is logged, consistent with a native crash rather than an uncaught JS exception.

## Suspected cause (unconfirmed — no crash dump yet)

`decodeBody()` in `src/router.mjs` (around line 372) calls `zstdDecompressSync` from `node:zlib` to decompress the incoming request body:

```js
if (encoding === "zstd") decoded = zstdDecompressSync(decoded, options);
```

[...]

PR #764, opened 2026-09-14
fix(cursor): preflight and async-inflate the last synchronous zstd path

duolahypercho:
Relates to #465. **This does not fix that crash**, and does not claim to.

## Context

PR #564 hardened `decodeBody` on `/v1/responses`: the Zstandard frame header is read for its declared content size and answered with 413 *before any native decoder sees the body*, and the inflate uses the promisified async `zstdDecompress` so a multi-megabyte frame does not run on the event loop.

`src/cursor-surface.mjs` was missed. It still called the one-shot synchronous `zstdDecompressSync` with no preflight, and it is imported by `src/router.mjs:33`. `git log -S zstdDecompressSync -- src/cursor-surface.mjs` returns exactly one commit — the one that introduced it — so the #564 hardening never reached this file.

It runs only for `/cursor/...` requests, so it cannot account for #465's `/v1/responses` crashes (Windows `0xC0000409` / `STATUS_STACK_BUFFER_OVERRUN`). That root cause remains unconfirmed and blocked on a crash dump. This is simply the last synchronous zstd inflate in the router, and it gets the same defensive treatment for the same reasons — as noted in the issue thread.

## Change

- `decodedRequestBody` is now `async`. Its zstd branch calls `zstdFrameContentSize` from `src/http-utils.mjs` — the existing parser, not a copy — and refuses a frame declaring more than `MAX_CURSOR_BODY_BYTES` before the native decoder is handed anything.
- The inflate is `promisify(zstdDecompress)` under the same `maxOutputLength`. The cap value (32 MiB) is unchanged.
- The three call sites (`handleCursorApp`, plus `RunSSE` and `BidiAppend` in `handleCursorCli`) gain `await`; all three were already `async`.
- gzip, deflate and br are untouched. No synchronous zstd inflate remains anywhere in `src/`.

## Status code

The oversize rejection is a `status`-carrying error, like the 400/404/415 this surface already raises. All three call sites sit outside the local `try`, so it reaches the server-level catch in `router.mjs` and `httpErrorStatus` returns 413 — the same status the equivalent rejection returns on `/v1/responses`.

Deliberately **not** ported: `decodeBody`'s `ERR_BUFFER_TOO_LARGE` -> 413 wrap. Here it would also change the gzip/deflate/br overflow status. A frame that omits its declared size and overflows the cap still returns 502, exactly as before. Porting that is a separate, broader call.

## Verification

```
node --test test/cursor-surface.test.mjs test/http-utils-zstd-frame.test.mjs
tests 14  pass 14  fail 0
```

One regression test, using the file's existing HTTP fixture: a real `zstdCompressSync` body still round-trips and routes, then a hand-built header declaring 64 MiB — followed by bytes that are deliberately *not* a decodable frame body — returns 413 with the upstream call count unmoved. The garbage tail is the point: if the preflight ever stops firing, the decoder fails the inflate instead and the status drops to 500. Verified by disabling the preflight, which produces exactly that failure.


duolahypercho:
Line-by-line review: adopt.

This is the last `zstdDecompressSync` in `src/`. Same preflight as `/v1/responses`: read declared frame size, 413 before native decode, async `zstdDecompress` under the existing 32 MiB cap. gzip/deflate/br unchanged. The three call sites were already async.

The oversize fixture (64 MiB header + undecodable tail) is a real negative control: if the preflight dies, inflate fails and the status is no longer 413. Merging.

Issue #760, opened 2026-09-14
Windows install: `start-codex-router.cmd` reported installed but missing from disk; background-service registration fails with `fetch failed


Owie6789:
## Summary

On Windows, the installer reports `"installed":true` for `start-codex-router.cmd` but the file does not exist on disk. The background-service (Task Scheduler) registration step fails with `fetch failed`, and the install exits with status 1.

## Environment

- **OS:** Windows 11 (PowerShell 7, FullLanguage mode confirmed)
- **Node:** v26.1.0
- **Codex Router version:** 0.5.1
- **Elevated (admin) PowerShell:** Yes
- **Proxy variables (`HTTP_PROXY`, `HTTPS_PROXY`):** Not set
- **Windows Defender / Firewall:** Not the cause (tested)

## Steps to reproduce

1. Run the Codex Router installer on a clean Windows 11 machine (admin PowerShell).
2. Accept all prompts.
3. Observe the install output.

## Actual behavior

```
{"installed":true,"path":"C:\\Users\\<user>\\.codex\\codex-router\\start-codex-router.cmd"}
fetch failed
Background-service installation failed.
At C:\Users\<user>\AppData\Local\codex-router\install.ps1:533 char:30
+ ... STEXITCODE -ne 0) { throw "Background-service installation failed." }
```

- `C:\Users\<user>\.codex\codex-router\start-codex-router.cmd` does **not** exist after install.
- Re-running the installer or `codex-router.ps1 doctor --fix` produces the same failure.
- The "Codex Router" scheduled task is never created.
- `codex-router.ps1 start` then fails because `schtasks /Change /TN "Codex Router" /ENABLE` targets a non-existent task.

## Expected behavior

- `start-codex-router.cmd` is written to disk.
- The "Codex Router" scheduled task is registered.
- `codex-router.ps1 start` / `stop` work.

## Workaround

```powershell
# 1. Create the missing .cmd
New-Item -ItemType Directory -Force -Path "$env:USERPROFILE\.codex\codex-router"
@"
@echo off
node "$env:LOCALAPPDATA\codex-router\src\start.mjs"
"@ | Out-File -FilePath "$env:USERPROFILE\.codex\codex-router\start-codex-router.cmd" -Encoding ascii

# 2. Create the scheduled task
schtasks /Create /TN "Codex Router" /TR "$env:USERPROFILE\.codex\codex-router\start-codex-router.cmd" /SC ONLOGON /RL LIMITED /F
```

[...]

PR #767, opened 2026-09-14
fix: keep a slow-starting service instead of rolling it back (#760)

duolahypercho:
Fixes the cause of #760. #762 fixed a latent defect found in the same area; this is what actually happened to that reporter.

## What happened

Every symptom in the report follows from one path, with no contradiction between them and no misconfiguration:

1. `install.ps1` sets `$ServiceInstalled = $true` **before** calling `service.mjs install`.
2. `service-windows install` wrote both launchers and registered the task. **`installed:true` was accurate.**
3. `service.mjs` then runs `waitForServiceReadiness` (`READINESS_TIMEOUT_MS = 300_000`) — a step `service-windows.mjs install` does not itself contain, which is why the `fetch failed` lands *after* the JSON line rather than before it. The router did not answer; `service-readiness.mjs:80-82` strips the URL, leaving the bare message.
4. Non-zero exit → `install.ps1` throws `Background-service installation failed.`
5. The catch runs `node src/service.mjs uninstall`, which deletes the task **and unlinks `launcherPath` and `wrapperPath`**.

So `start-codex-router.cmd` was deleted by the installer's own rollback, seconds after being written correctly. The operator then finds a log claiming `installed:true`, no file, and no task — all three true, in that order.

`service.mjs:25-28` already anticipates this exactly:

> A shorter wait reports failure while the service is still booting, and the installer's rollback then uninstalls the service and reverts the app config out from under a router that goes on to come up healthy seconds later.

The guard added for it gates the teardown on `service_was_installed` / `$ServiceWasInstalled`, which protects a **reinstall over a working router**. A first install on a clean machine has no prior state to compare against, so it was torn out regardless — and a first install is precisely when a LiteLLM cold start with a large model set is slowest. It also explains why the reporter's manual workaround succeeds where the installer failed: by then the gateway is warm.

## Change

- `waitForServiceReadiness` tags health-wait exhaustion as `readinessTimeout`. The crash-loop and dead-launcher throws are deliberately **left untagged** — those are broken rather than slow, and a rollback is still right for them, or a task that relaunches a failing router at every logon survives the install.
- `service.mjs` converts only the tagged failure into exit **75** (`EX_TEMPFAIL`) and says the service is installed and still starting. Everything else keeps exiting 1.
- Both installers branch on 75 and skip the teardown for **service and client config alike** — a router that comes up moments later needs its config still pointing at it — then stop before `wait-health.mjs`, which would only spend a second full budget failing the same way on the same cold start.

The install still fails, loudly, with a non-zero status. What changes is that it no longer destroys what it correctly built.

## Verification

```
node --test test/installer-scripts.test.mjs test/service-readiness.test.mjs \
  test/service-lifecycle.test.mjs test/service-render.test.mjs
tests 99  pass 95  fail 0  skipped 4   (Windows/PowerShell-only)
```

New coverage: a health wait that runs out is tagged retryable; a crash loop and a dead launcher are **not**; `service.mjs` converts only the tagged case; and both installers branch on 75, initialise the flag before the rollback reads it (under `set -u` an unset variable would abort the trap itself), guard ahead of the teardown, and decide before the second health wait. None of the four markers the new assertions match exists on `main`, so every one of them fails pre-fix.

**Authored on macOS.** `pwsh` is not available on this host, so the PowerShell edit rests on the installer-script assertions plus CI's Windows job rather than a local parse — worth a reviewer's eye on the bare `throw` rethrow in the catch.

## What this does not cover

Why that machine's router needed more than 300 seconds is still unknown, and this PR does not shorten it — it only stops the installer from reacting destructively. If cold starts routinely overrun the budget, the budget itself is the next question.


duolahypercho:
Line-by-line review: adopt.

The installer set `service_installed` before the 300s health wait, then rolled back a first install that had already written launchers and registered the task. Tagging only health-wait exhaustion as `readinessTimeout`, exiting 75, and skipping teardown for that case is the correct split: crash loops and dead launchers still roll back.

POSIX initializes the flag before `rollback` (needed under `set -u`). Windows rethrows before uninstall and only restores env in `finally`. Tests cover the tag, the untagged failure paths, and both installer guards. Merging.

PR #778, opened 2026-09-15
opencode Zen protocol variants + omit-tool-choice for Qwen behind Go Messages

cjchanh:
# opencode Zen protocol variants + omit-tool-choice for Qwen behind Go Messages

Follow-up to #772, rebased on v0.6.0.

## Zen Messages / Responses variants

Zen serves Anthropic, OpenAI, xAI and Meta models over their native wire protocols, exactly as Go does, but only the Go family had the `-messages` / `-responses` provider variants. Every curated Zen model from those families answered HTTP 500 through the chat route (probed live: 22 of 70 Zen models worked, all 48 failures were those families). This adds `opencode-zen-messages` and `opencode-zen-responses`, mirroring the Go variants (same credential, `baseUrl` `zen/v1`, `variantOf: opencode-go` because a variant may not be a variant of a variant). After curating under the right variant, Claude, GPT, Grok and Muse on Zen answer; Gemini on Zen does not on any route (Google-native protocol, no adapter here) and is left out.

## omit-tool-choice

opencode Go's Messages route now answers HTTP 400 (body is only the model id) for every Qwen model when the request carries `tool_choice` in any form, `auto` and `none` included, and calls the listed tools when the field is absent. MiniMax on the same route still accepts the field, so this is per model. `auto-tool-choice` no longer keeps these routes alive: Codex sends `tool_choice` on every tool turn, so every such turn failed. New request profile `omit-tool-choice` deletes the field and keeps the tools; applied to the five Qwen `opencode-go-messages` entries (hashes bumped) and exposed as a curatable profile for resellers with the same quirk. Verified live on all five after the change.

## Tests

`test/opencode-go-messages-qwen-tool-choice.test.mjs` (profile applied, profile known and curatable); registry loads with the new variants; related suites green.


duolahypercho:
Thanks for this, and sorry it sat so long.

Both halves of this have since landed on `main` through other changes:

- The Zen protocol variants exist as `opencode-zen-messages` and `opencode-zen-responses` in `config/opencode/opencode.json`, wired through `src/opencode-session.mjs` and `src/opencode-curation.mjs`, and documented in AGENTS.md.
- The `omit-tool-choice` request profile exists in `src/request-profiles.mjs` and is already applied to all five `config/opencode/go-messages/qwen*.json` routes.

So the config changes here are no-ops against current `main`, and the diff conflicts with the equivalent change that landed. Closing as superseded rather than leaving it open — the diagnosis was right, it just arrived twice.

If you still see a Zen or Qwen failure on current `main`, please do open a fresh issue or PR with the repro; that would be very welcome.

PR #798, opened 2026-09-16
fix: keep OpenCode Go Messages tools and streams compatible

tonyjzhou:
## Summary
- Sanitize OpenCode Go Messages tools to named functions with object schemas so Codex hosted/custom leftovers no longer 400 the Anthropic Messages route (live Union Alpha `tools[116]`).
- Apply LiteLLM's Chat Completions reasoning-summary repair to Anthropic protocol providers. Those routes still use `use_chat_completions_api: true`, so thinking-only `reasoning_text` closes were classified as empty completions and 502'd after retry.
- Keep `reasoning_text` / thinking parts out of empty-completion *content* so a silent retry can still recover the next attempt's text.

Live exact-route stream on `opencode-go-messages/union-alpha` at max returned HTTP 200 with `CODEX_ROUTER_STREAM_OK`.

Conversation: https://app.warp.dev/conversation/c1d4960c-1ef9-4c62-8bc9-e57fc5856675

## Test plan
- [x] `node --test test/empty-completion-guard.test.mjs test/grok-reasoning-summary-compat.test.mjs test/namespace-relay.test.mjs test/chat-tool-surface.test.mjs test/namespace-relay-routing.test.mjs` (260 pass)
- [x] Live compatibility-test stream at max on `opencode-go-messages/union-alpha` (200, marker, completed)
- [ ] Review CI on this PR


duolahypercho:
Thanks for this, and apologies for the delay.

Most of it has since landed on `main` via commit `f6ff49b2` ("make Union Alpha usable on OpenCode Go Messages"), which introduced essentially the same fixes:

- `anthropicFunctionTools` in `src/namespace-relay.mjs` for named-function tools with `input_schema`
- the `reasoning_text`/`reasoning`/`thinking` exclusion in `partHasContent` in `src/empty-completion-guard.mjs`
- the non-`openai`/non-`anthropic` protocol gate in `src/grok-reasoning-summary-compat.mjs`

Because those landed in the same functions, this branch now conflicts semantically rather than textually, so closing as superseded.

**One piece here is still not on `main`,** and is worth a fresh PR on its own: adding `"opencode-go-messages"` to `BOUNDED_TOOL_NAME_PROVIDERS` in `src/chat-tool-surface.mjs` (currently only `commandcode`/`commandcode-messages`) for the 64-character tool-name bound. If you can still reproduce that failure on current `main`, please send just that change — it would go in quickly.

PR #799, opened 2026-09-16
Publish the real GPT-5.6 effort ladder for kiro-prism

gchahal1982:
## The problem

The three GPT-5.6 tiers carried a single rung:

```json
"reasoningLevels": [{ "effort": "medium", "description": "Gateway default reasoning" }]
```

So no client routed through codex-router could select an effort for `gpt-5.6-sol`, `-terra`, or `-luna`, even though Prism advertises the full scale.

## Why it was wrong

That single-rung entry was recorded when Prism itself reported an **empty** effort ladder for these models. Prism's `effort_levels()` read only `output_config.effort` — the Claude container — while GPT-5.6 carries effort in `reasoning.effort` ([Kiro docs](https://kiro.dev/docs/models/effort/)). Prism now resolves both containers and reports the real ladder:

```
gpt-5.6-sol     effort_levels=['none','low','medium','high','xhigh','max']
gpt-5.6-luna    effort_levels=['none','low','medium','high','xhigh','max']
gpt-5.6-terra   effort_levels=['none','low','medium','high','xhigh','max']
```

## The change

Publishes the five **rankable** rungs. `none` is deliberately omitted: it is not on this repository's `EFFORT_LADDER` (`minimal..ultra`) and is an off-state rather than a depth the picker should rank. `clampEffort` passes off-ladder values through unchanged, so carrying it would put an unrankable entry in the picker.

`defaultEffort` stays `medium`, matching Prism's own per-family default for these tiers.

Claude entries are untouched — their effort has always lived in `output_config`, so they already advertised the correct ladder.

## Verified

After `refresh-catalog`, the merged catalog Codex actually reads:

```
kiro-prism/gpt-5.6-sol      default=medium  levels=['low','medium','high','xhigh','max']
kiro-prism/gpt-5.6-luna     default=medium  levels=['low','medium','high','xhigh','max']
kiro-prism/gpt-5.6-terra    default=medium  levels=['low','medium','high','xhigh','max']
kiro-prism/claude-opus-5    default=medium  levels=['low','medium','high','xhigh','max']
```

Cherry-picked onto `origin/custom/v0.5.1` in a temporary worktree so unrelated uncommitted work in the main checkout was never touched.


gchahal1982:
Opened against the wrong repository by mistake — this belongs on the gchahal1982 fork, not upstream. Closing; no review needed.

PR #815, opened 2026-09-17
fix: opt-in withholding of Codex app-connector tools on routed chat providers

haytham10:
## Problem

Codex registers every installed app connector (Airtable, Notion, Slack, GitHub,
Canva, Figma, Gmail, Drive, ...) as an eager namespace on routed requests:
nothing carries `defer_loading` and no `tool_search` control is sent. A live
capture of a one-word turn (DeepSeek on OpenCode Go, Union Alpha on OpenCode Go
Messages) showed 38 tool entries across 25 namespaces with no deferral marker
at all. `flattenNamespaceTools` flattens Codex's whole connector registry into
the prompt: 576 provider-facing tools, ~1.09 MB of JSON Schema, ~200,000 input
tokens billed before the model reads the first word. Connectors disabled in
the Codex UI are still sent. This cost is silent to the operator. Native
(Responses-native) models don't pay it, because they take the deferred/code-mode
path Codex already supports.

## Change

`CODEX_ROUTER_APP_CONNECTORS` opts in to withholding these tools, in three modes:
- unset (default): unchanged behaviour, every connector declared eagerly.
- `none`: withholds every `mcp__codex_apps__*` namespace from a routed
  chat/messages surface, whether or not Codex marked it deferred and whether or
  not a search relay exists.
- a comma-separated list (e.g. `airtable,gmail`, trimmed, case-insensitive,
  unknown ids ignored): keeps those connectors eager and withholds the rest.

`mcp__codex_app` (threads, automations, navigation), `collaboration`, the
repls, the template picker, `image_gen`, and every plain tool always stay
eager. A withheld connector keeps its namespace registration, so stored calls
are still renamed outbound and restored inbound. Its definition is re-declared
the moment a stored `function_call`, a forced `tool_choice`, or an
`allowed_tools` entry names it, so an in-progress session never cites a tool
the request doesn't declare. That restoration is charged against a bounded
provider's tool ceiling like any other definition (a Groq surface that can't
hold them all still fails locally), and it is repaired again for routes whose
validators reject a sibling `$ref` (Kimi). The variable is read per request, so
no restart is needed. Responses-native routes are unchanged. Default behaviour
is unchanged.

## Measurements

Captured fixture: 576 tools / 1,091,492 bytes -> 76 tools / 121,002 bytes with
`none` (most of the remainder is the app snapshot the router injects itself).

Live before/after:
- DeepSeek fresh turn: 253,813 -> 45,576 input tokens.
- Union Alpha on OpenCode Go Messages: 241,388 -> 41,019 input tokens.

## Tests

Ran the five test files touched by this change:

```
node --test test/namespace-relay.test.mjs test/chat-tool-surface.test.mjs \
  test/namespace-relay-routing.test.mjs test/codex-app-tools.test.mjs \
  test/namespace-relay-custom.test.mjs
```

217 pass, 0 fail. The full suite was not run locally. On unmodified `main`,
four pre-existing/environmental failures are seen and are unrelated to this
change:
- `chatgpt-profile-switch` crash-boundary: 180s timeout.
- `login-free-app-server`: EPERM on Windows temp cleanup.
- `desktop-panel`: missing playwright.
- `catalog-publication-lock` and `config-manager`: pass in isolation.

## Open question for the maintainer

Whether to flip the default to `none` later, once operators have had a chance
to see the token savings against their own connector usage.


duolahypercho:
This passes on its own head (219 tests, and `npm run check`). After merging with current `main`, it breaks in two places:

1. `test/chat-tool-surface.test.mjs` imports `flattenToolSearchHistory` twice. That's a SyntaxError, so `npm run check` fails.
2. With the duplicate import removed, "a chat route omits deferred connector tools behind the bridged search relay" still fails. The PR collides with 6a0dc78 (honor deferred app tools with tool search), so it needs a rebase onto that commit.

**The default behaviour changes, contrary to the description.** Whenever the client sends `tool_search`, tools marked `defer_loading` are withheld even with `CODEX_ROUTER_APP_CONNECTORS` unset. `mergeCodexAppTools` also now strips `defer_loading`. Please either keep the default truly unchanged, or state the change plainly.

**Scope.** The PR also carries an unrelated port-pool change (`test/port-pool.mjs` and its test) and an extra Moonshot schema-repair pass. Please split those into their own PRs.

**Changelog.** Please use a `changelog.d/` fragment instead of editing `CHANGELOG.md`. That edit is why GitHub reports this PR as conflicting.

Whether the project wants the new environment variable, and whether its default should ever become `none`, is the maintainer's decision.

---

Issue #770, opened 2026-09-15
invalid_argument on model list — Devin CLI 3000.x likely broke wire format


nsgreenw:
codex-router devin-cli probe
node: v26.6.0
platform: darwin/arm64
os: Darwin 25.6.0
devin cli: devin 3000.10.21 (611c1cba)
mode: free

PASS  devin cli session: read from ~/.local/share/devin/credentials.toml
INFO  cascade host: server.codeium.com
PASS  request encoding: 11 field(s) written, 498 bytes, all on their declared wire types
FAIL  model list: the upstream refused the request: an internal error occurred
      observed: http=400 connect-code=invalid_argument known=true retriable=false content-type=application/json
      fix: Paste the observed line; the connect-code names which assumption broke.

VERDICT FAIL — 2 pass, 1 fail, 0 warn, 0 skipped

bin/curate-models devin-cli fails identically:
codex-router curate-models: Devin upstream refused the request: an internal error occurred
Plan type: Individual Pro
Model list in official Devin CLI: 48 models shown — account has healthy entitlements, so this doesn't look like a plan restriction.

Likely cause: Devin CLI jumped from the 2025.x series the router was built against to 3000.10.21. request encoding passes the self-consistency check, but the upstream rejects it — points at the protobuf shape not matching what Cognition's backend currently expects.

PR #818, opened 2026-09-17
fix(devin): match CLI metadata and stream tool arguments

nsgreenw:
## Summary

Follow-up to #807 / #770 after testing the merged fix against Devin CLI `3000.10.31`.

`GetCliModelConfigs` is the correct method, but Devin rejects the request with `invalid_argument` when `metadata.extension_version` and `metadata.ide_version` are absent. The official CLI also identifies as `chisel` / `0.0.0-dev`, not `windsurf` / `1.0.0`.

This patch:

- sends the required model-discovery metadata
- matches the official CLI client identity for chat turns
- assembles Devin's streamed tool calls, where the named header arrives first and unnamed argument fragments follow
- emits only each new argument fragment to OpenAI-compatible streaming clients
- adds regressions for metadata and fragmented tool arguments

## Verification

Against Devin CLI `3000.10.31` on macOS:

- `./bin/devin-probe`: **PASS** (209 enabled model UIDs discovered)
- `./bin/devin-probe --live --tools --model swe-2-high --max-tokens 128`: **12 pass, 0 fail**
- `./bin/test-model 'devin-cli/swe-2-high' --live --yes`: **5 pass, 0 fail** (non-streaming, streaming, tools, tool-result replay, compaction)
- focused Node tests: **46 pass, 0 fail**
- `npm run check`: pass

A paid model was also attempted, but that account's weekly paid quota was exhausted. The free `swe-2-high` model completed the full routed harness.


nsgreenw:
Closing because the real Codex client path is not reliably supported by this patch.

PR #828, opened 2026-09-18
fix: stop refusing routed children on the OpenCode Go routes

rafaelwall85:
## What

Two measured refusals on the OpenCode Go routes, each fixed on the exact provider that proved it.

- **`opencode-go-responses`: a delegated child died on its first turn.** Console Go implements the public Responses input schema only. It answers ` ``input[N]` did not match any supported type ` for Codex's `agent_message` collaboration item and ` ` input[N].content` ... ` for the `encrypted_content` part inside it, so a Muse Spark 1.3 Contributor child was refused on its first turn while the very same route served its parent normally. `consoleGoResponsesInput` converts the handoff into the ordinary user message the endpoint accepts and carries a still-opaque value as visible text; it runs on the turn and on a delegated transcript's compaction.

- **`opencode-go` Chat: `access_programs` refused the whole request.** Codex sends that top-level field because the merged catalog publishes the model's `available_access_programs`; Console Go answers `invalid request body: json: unknown field "access_programs"` before inference. Measured on `opencode-go/glm-5.3-flash` and `opencode-go/glm-5.3` as delegated children. The field is now dropped for that provider only -- the Responses variant accepted the identical field unchanged in the same session.

## Evidence

Reproduced and isolated on the local caller endpoint before the change:

| Shape | Before | After |
|---|---|---|
| `agent_message` whose visible text ends at `Payload:` | 400 `input[1]` | 200 |
| `agent_message` with no visible text | 400 `input[1]` | 200 |
| `agent_message` whose visible text does not end at `Payload:` | 400 `input[1]` | 200 |
| `access_programs` on `opencode-go/glm-5.3-flash` | 400 `unknown field` | 200 |
| `access_programs` on `opencode-go-responses/muse-spark-1.3-contributor` | 200 (unchanged) | 200 |

End to end after the change, with the routes confirmed in the router meter (`isSubagent: true`) rather than in the child's own claim: a routed parent delegated to `opencode-go/deepseek-v4.1-flash` (max), `opencode-go-responses/muse-spark-1.3-contributor` (xhigh) and `opencode-go/glm-5.3-flash` (max), and each child returned its exact marker with status 200.

## Tests

- New route tests in `test/namespace-relay-routing.test.mjs`: `OpenCode Go Responses carries a delegated task as an ordinary user message` and `Console Go Chat drops the access_programs field its endpoint refuses`. The second also asserts that the Responses route keeps the field, so the gate stays exact.
- `node --test test/namespace-relay-routing.test.mjs` -> 35/35
- `node --test test/namespace-relay.test.mjs` -> 132/132
- `node --test test/routing.test.mjs test/router-resilience.test.mjs test/deepseek-tool-message-compat.test.mjs test/subagent-certify.test.mjs` -> 264/264
- `npm run check` -> v2-agent applications valid (7), syntax checks passed

## Limits

- Both new gates are exact to the provider that was measured. Paid Zen and other Responses providers keep their native contract until their own endpoint proves the same restriction.
- The five-check v2 certification still cannot complete under a ChatGPT account (the native parent is refused), so these routes stay promoted by operator selection. This PR does not change that.


duolahypercho:
This now conflicts with `main` — could you rebase?

#830 landed in `bf55dec2` and deletes `access_programs` unconditionally on every routed path, which is a strict superset of the `opencode-go`-only gate here:

- `src/router.mjs:2896` — `delete body.access_programs;` (compaction)
- `src/router.mjs:3762` — `delete routed.access_programs;` (routed turn)
- plus `src/api-forwarder.mjs:985` and `src/oauth-forwarder.mjs:60`

So the rebase is: **drop `rejectsAccessPrograms()` and both of its call sites** — the field is already gone before your code runs, and the live measurement in your comment (`unknown field "access_programs"` on `opencode-go/glm-5.3-flash`) stays fixed by #830.

Everything else in this PR is independent and still wanted: `consoleGoResponsesInput()` in `src/namespace-relay.mjs`, the `routedInput` conversion, and the compaction-path fix. That is the actual fix for delegated children dying on their first turn with `` `input[N]` did not match any supported type ``, and nothing on `main` covers it.

Note that #831 (`fix/console-go-muse-handoff`) also merged and touches the same Console Go paths in `src/router.mjs` and `src/openai-adapters.mjs`, so it's worth re-running `test/namespace-relay-routing.test.mjs` after the rebase.

duolahypercho:
Thanks for this. Both fixes have since landed on main separately, so I'm closing this as superseded:

- stripping `access_programs` is done for every routed request (and compaction) by #830;
- converting delegated-child `agent_message` input for Console Go Responses is done by #831, deliberately scoped to the Muse contributor routes where the contract was measured (other Console Go models answered with a different contract).

The remaining difference here would widen that conversion to every opencode-go-responses model without a measurement, and would turn undecodable `encrypted_content` parts into visible text. If you can capture a real `agent_message` whose `encrypted_content` survives unwrapping on a Muse route, a narrow follow-up extending `consoleGoCompatibleInput` (dropping opaque tokens rather than exposing them) would be very welcome.

PR #831, opened 2026-09-18
fix: make Console Go Muse routes usable in collaboration

lorianlee98-spec:
Two defects made the paid Console Go Muse Contributor routes unusable inside a
multi-agent session, one on each side of the relationship. Both were measured
against the live routes, and both are fixed here.

## Delegating to a Muse child

A delegated child on these routes died before its first token:

```
opencode rejected the request for Muse Spark 1.3 Contributor (opencode Go).
(HTTP 400: {"model":"muse-spark-1.3-contributor","error":{"param":"input[5]",
"type":"invalid_request_error","message":"Error from provider (Console Go):
Upstream request failed: [invalid_request_error] `input[5]` did not match any
supported type"}})
```

Codex hands work to a child as an `agent_message` item. When the parent is
routed, the task text is stored as plaintext under `encrypted_content`;
`normalizeRoutedAgentInput` recovers it and replaces that part with `input_text`,
but the item keeps `type: "agent_message"`, which is a Codex collaboration item
and not part of the public Responses schema. Console Go rejects that item type by
name on these two upstreams -- a request whose only input item is an
`agent_message` answers 400 with exactly the message above, while message,
reasoning and function items are accepted.

The Zen free pair and the native DeepSeek route already convert the handoff to an
equivalent user message (`zenFreeCompatibleInput`, `deepSeekResponsesInput`).
`consoleGoCompatibleInput` does the same on the exact upstreams measured
rejecting it (`muse-spark-1.2-contributor`, `muse-spark-1.3-contributor` behind
`opencode-go-responses`), at both places that build that route's body: the routed
turn (`buildRoutedRequest`) and compaction (`summarizeWith`), so a child session
cannot pass its first turn and then die on its first summary.

Sibling Console Go routes are not silently affected: `opencode-go-responses/gpt-5.6-luna`
answers the same item with its own contract (`Missing required parameter:
'input[0].author'`), and `grok-oauth/grok-4.6` reaches its provider through
LiteLLM's Chat bridge, which normalizes the item. They keep the item unchanged.

## Collaboration calls made from a Muse parent

With Muse Spark 1.3 as the parent, every collaboration tool answered
`unsupported call: collaboration__spawn_agent`. The stream described the same
call two different ways:

```
response.output_item.added  -> { name: "spawn_agent",   namespace: "collaboration" }
response.output_item.done   -> { name: "collaboration__spawn_agent" }
```

Codex executes the terminal item, so the restored `added` event never helped.
`normalizeResponsesEvent` restored the namespace only for
`response.output_item.added`; the `done` item and the `response.completed`
snapshot were forwarded exactly as the provider wrote them.

Both places now restore the same call, which also keeps the three descriptions of
one call in agreement on any Responses route that relays flattened names.

## Validation

- New test in `test/namespace-relay-routing.test.mjs`: the handoff reaches both
  Muse Contributor routes as a user message carrying the recovered task text,
  while a sibling Console Go route keeps the item it was written for. With
  `src/router.mjs` reverted the test fails; with the fix it passes.
- New test in `test/qwen-plan-namespace-restoration.test.mjs`: the terminal item
  and the completion snapshot come back namespaced. With `src/openai-adapters.mjs`
  reverted the test fails; with the fix it passes.
- `npm run check`.
- Affected suites green: `openai-adapters`, `qwen-plan-namespace-restoration`,
  `namespace-relay`, `namespace-relay-custom`, `namespace-relay-routing`,
  `commandcode-forwarder` and `grok-oauth-forwarder` (270 tests together).
- A full local `npm test` has three failures -- `test/desktop-panel.test.mjs`,
  "real Codex app-server completes signed-out turns through fallback and
  preserved providers" and "retry policy is bounded by one operation deadline and
  cancellation". They reproduce unchanged on `main` without these commits
  (checked in a clean worktree at c6f58a3), so they are environmental rather than
  caused by this change.
- Delegation on a running router: a routed parent delegated to Muse Spark 1.3
  children through the `general-worker` and `code-mapper` presets. One returned
  its marker, the other ran a shell command and reported its output; the route
  metered 200 for both.
- Wire check on the same router: a forced `collaboration__spawn_agent` call on
  `opencode-go-responses/muse-spark-1.3-contributor` now returns
  `{ name: "spawn_agent", namespace: "collaboration" }` in `added`, `done` and
  `completed.output`. Before the change `done` carried the flattened name, which
  is what Codex reported as `unsupported call`.
- Not verified here: a real Codex app session with a Muse parent running a
  collaboration tool end to end. The wire shapes above are what the app reads, so
  that step is a retest in the app rather than more code.


Yormede:
Independent confirmation of the failure, from a different machine and a different patch: a single request to a local router carrying the same `agent_message` item returned `400 input[4] did not match any supported type` on `opencode-go-responses/muse-spark-1.3-contributor`, and `200` once the plaintext payload is recovered and the item presented as a user message. All three Codex subagent roles failed identically before the change and ran afterwards.

Your narrower gate looks like the safer one. I had written the same conversion provider-wide, which also rewrites the input for `grok-4.6` and `gpt-5.6-luna` on that provider. I never tested those siblings, so the two-upstream restriction here is better grounded than what I had.

Nothing needed from me -- just adding the reproduction in case it helps the review.

PR #841, opened 2026-09-19
feat: add OpenRouter Decisions routing for Jev

moinerus:
## Summary
- register an OpenRouter Decisions provider variant and hidden Jev route at openrouter-decisions/jev-latest
- expose caller-capability protected POST /v1/decisions with the existing request/response limits, cancellation, metering and credential isolation
- forward the native body unchanged to OpenRouter's alpha Decisions endpoint, translating only the routed model id
- refuse redirects and retries for billable Decisions requests, and keep the model out of conversational pickers

## Verification
- registry and endpoint-policy tests: 3 passed
- hermetic router to forwarder to upstream Decisions test: passed
- npm run check: passed

The end-to-end test asserts loopback capability gating, no caller/internal credential forwarding, model translation, size limits, cancellation, and redirect refusal.


moinerus:
## Additional evidence: live Decisions route proof

A bounded live run started temporary loopback Codex Router and API-forwarder processes from this PR branch. The Router-owned OpenRouter credential and caller capability remained process-only.

- Route: `/v1/decisions`
- Model: `openrouter-decisions/jev-latest`
- Completed in **961 ms**
- Full pruner path: **63,793 bytes -> 2,326 bytes (96.35%)**
- The original output archive was byte-for-byte exact and the final `tests=184 passed=184 failed=0` status was retained.
- No credential or caller capability was emitted or persisted by the evidence harness.

I also ran a targeted mutation campaign against this change:

- Killed: forwarding the Router gateway model rather than `~typesafe/jev-latest`
- Killed: removing both request-size guards
- Killed: making the model picker-visible, caught by registry validation
- One raw-size-limit-only mutant survived because the independent decompressed-body size guard still returned `413`. That is intentional defence in depth; removing both guards was killed.

This is focused evidence, not a claim that the project's entire cross-platform test suite is green.

PR #842, opened 2026-09-19
fix(grok): lead flattened tool schemas with id so automation_update selects

Ryanm218:
## Summary
Grok emits object keys in the order the parameter schema lists them. Codex's live `automation_update` view branch declares `mode` then `id`, and the union flatten kept that order, so view/update/delete calls arrived with `mode` first.

`objectRootToolSchema` now leads with `id` when the field exists, on both merged unions and already-object-rooted copies. `id` stays optional (create omits it). Schemas with no `id`, or with `id` already first, stay by identity.

## Test plan
- [x] `node --test test/tool-schema-root.test.mjs` (72/72)
- [x] `node --test test/grok-oauth-forwarder.test.mjs test/namespace-relay.test.mjs` (203/203)
- [x] `npm run check`
- [x] Live flatten of `automation_update` now starts `id, mode, name, prompt, rrule`; required remains `["mode"]`


Ryanm218:
Closing: this change is fork-local for my Grok/Codex Router checkout and does not apply upstream. Kept on Ryanm218/codex-router only (merged as https://github.com/Ryanm218/codex-router/pull/1).

PR #845, opened 2026-09-19
fix(router): deliver the subagent handoff to Console Go Responses

Yormede:
### What is broken

A Codex subagent pinned to a Console Go Responses route cannot be spawned. The
parent's first item is a native `agent_message`, Console Go rejects the item's
discriminator, and the child dies before it reads its task:

```
{"model":"muse-spark-1.3-contributor","error":{"param":"input[4]",
 "type":"invalid_request_error","message":"Upstream request failed:
 [invalid_request_error] `input[4]` did not match any supported type"}}
```

Measured on `opencode-go-responses/muse-spark-1.3-contributor` at router
0.6.0 (`5e1b49e6`). All three subagent roles fail identically -- the index
moves with the transcript, the rejection does not.

### Why

`agentMessagesAsUserMessages` already exists for exactly this, but
`zenFreeCompatibleInput` only reaches it for the free Zen pair. The paid
Console Go route keeps the native item, so the strict upstream sees a type it
does not implement. The payload is not encrypted either: Codex stores the
readable handoff under `agent_message.content[].encrypted_content`, and the
value is a plain string rather than a Fernet token.

### The change

`consoleGoCompatibleInput` renders a non-Fernet `agent_message` payload as text
through the existing `normalizeAgentMessageForNative`, then presents the item
as the ordinary user message Console Go accepts. It is wired into both entry
points -- the ordinary routed turn and routed compaction -- so a child can be
spawned, resumed, and compacted on these routes.

The gate is `needsConsoleGoResponsesToolCompatibility`, so `opencode-go`,
`opencode-go-messages`, paid Zen, and every other Responses provider keep
their current wire payload byte-for-byte. A native Fernet token is left
untouched rather than rendered as ciphertext.

### Regression test

`test/routing.test.mjs` gains "a routed subagent handoff reaches Console Go as
a user message". It posts the handoff to a Console Go route against a mock
gateway and asserts the upstream body carries no `agent_message` and that the
non-Fernet payload arrives as readable `input_text`.

Verified both directions: the test passes on this branch and fails on
`5e1b49e6` with `the native agent_message discriminator reached the strict
upstream`. `npm run check` passes.

### Live evidence

Before the change, a direct request to the router carrying the same
`agent_message` item returned `400 input[4] did not match any supported type`.
After it, the same request returns `200` and the model answers with exactly the
marker the payload asked for.

Then with real Codex subagents, one per role:

| Role | Task | Result |
| --- | --- | --- |
| fast_explorer | list the `.toml` files in the agents directory | correct |
| code_worker | count usage-event rows, two independent methods | correct |
| deep_reasoner | analyse this patch and cite line numbers | correct |

All three children recorded
`opencode-go-responses/muse-spark-1.3-contributor`, every request in the
sequence returned 200, and no Console Go request has returned a non-200 since
the change.

### Relationship to existing guidance

`AGENTS.md` ("Routed subagent regression prevention") already requires the
reverse direction -- a non-Fernet `encrypted_content` part of an
`agent_message` rendered as `input_text` before a native Responses endpoint.
This applies the same rule to the Console Go Responses surface.


Yormede:
Withdrawing this one: it duplicates #828 and #831, both opened a day earlier and both broader.

The core is the same fix, not a similar one. #831 defines a function with the same name (`consoleGoCompatibleInput`), gated on `opencode-go-responses`, and calls it at the same two sites this branch uses -- the routed turn and `summarizeWith`. #828 reaches the same result through `consoleGoResponsesInput` in `namespace-relay.mjs`, and additionally drops `access_programs` on the Go Chat routes.

There is also a real defect in my version that #831 caught. My gate is provider-wide, so it also rewrites the input for `opencode-go-responses/grok-4.6` and `opencode-go-responses/gpt-5.6-luna`. #831 restricts it to the two Muse Contributor upstreams and reports that a sibling route answers the handoff with its own contract instead. I measured the Muse failure independently but never tested those siblings, so the narrower gate is the one with evidence behind it.

Independent reproduction, in case it is useful to the reviewers: one request to a local router carrying the same `agent_message` item returned `400 input[4] did not match any supported type` before any change and `200` after recovering the plaintext payload, on `opencode-go-responses/muse-spark-1.3-contributor`. All three Codex subagent roles failed identically before, and all three ran afterwards.

Closing in favour of #828 and #831. My local fix stays only as an operator-side patch until one of them lands.

Issue #297, opened 2026-08-17
GLM-5.3 starts with ~52% context used because standalone tool search is not enabled


PracticalSwan:
## Summary

`zai-coding/glm-5.3` currently starts fresh Codex sessions with roughly **480K–498K estimated input tokens** (about **52%** of the effective 1M context window) on a tool/plugin-heavy Codex installation.

The generic routed `tool_search` bridge already exists on `main` via #255, but the checked-in GLM-5.3 registry entry does not opt into it. Adding:

```json
"searchTool": {
  "mode": "standalone"
}
```

to `config/zai/coding/glm-5.3.json` reduced fresh-session startup input from about **491K to ~82K** in live A/B testing, while deferred MCP discovery, normal tool calling, streaming, compaction, and V2 subagents continued to work.

## Current upstream state

Current `main` defines `zai-coding/glm-5.3` with the verified 1M context / 900K auto-compact values, but no `searchTool` declaration:

```json
{
  "slug": "zai-coding/glm-5.3",
  "contextWindow": 1000000,
  "autoCompact": 900000,
  "inputModalities": ["text"],
  "requestProfile": "glm-thinking",
  "compHash": "zai-coding-glm-5-3-v1"
}
```

The repository already supports per-model standalone search/tool discovery. #167 documented the opt-in shape, and #255 added the native `tool_search` relay for routed models.

This therefore appears to be a missing per-model compatibility declaration rather than missing router infrastructure.

## Reproduction

Environment used for the measurements:

- Windows 11
- Codex Router `0.4.0-beta.4`
- model: `zai-coding/glm-5.3`
- context window: 1,000,000
- auto-compact: 900,000
- Codex installation with many MCP/plugin tools enabled

Across fresh GLM-5.3 sessions before the change, first-request router estimates were:

```text
489,946
493,797
495,094
494,559
480,203
497,685
492,800
497,270
490,700
```

Median: ~493.8K.

The narrow clustering strongly suggests a fixed startup surface rather than leaked conversation history.

Codex displays this as roughly 52% context used immediately because the usable context is about 950K after headroom.

## Why this happens

Without `searchTool.mode = "standalone"`, the merged model catalog advertises the routed GLM model without search/tool-discovery support, so Codex exposes a much larger direct tool/schema surface in the initial request.

This is especially expensive for installations with many MCP servers, plugins, and connector tools.

[...]

PR #847, opened 2026-09-19
fix: defer Z.ai Coding Flash tools and raise compaction floor

### LINKED ISSUE #846: Z.ai Coding GLM-5.3-Flash can compact-loop because the 400K threshold is below the subagent tool-surface floor
## Summary

`zai-coding/glm-5.3-flash` inherits the 400,000-token auto-compaction threshold that was originally chosen from OpenCode Go observations. On current Codex Desktop subagents, that threshold can sit **below the normal post-compaction request floor** once the full agent tool schema is reattached.

The result is a deterministic compact loop:

```
ordinary child turn (~450K-474K input)
-> Codex compacts successfully (~35K-55K compacted history)
-> next ordinary child turn reattaches the tool surface (~450K+ again)
-> immediately above the 400K threshold
-> compact again
-> repeat
```

This is provider-specific evidence for Z.ai Coding and is distinct from #820's Union Alpha reserve discussion and #266's zero-token estimate bug.

## Environment

- Windows Codex Desktop: `26.915.4065.0`
- Desktop-owned app-server: `0.155.0-alpha.9.2`
- codex-router live runtime: `0.5.1` plus selected merged backports
- clean reference investigated at `5e1b49e6e3fea1ada57f0e7fa6ca7612e9d17300`
- route: `zai-coding/glm-5.3-flash`
- advertised context: 1,000,000
- current auto-compact: 400,000
- effective Codex context event: 950,000

No synthetic provider probe was used; the measurements below came from naturally occurring routed subagent traffic.

## Evidence

Two independent Flash child tasks showed the same sequence repeatedly.

Representative provider-reported ordinary-turn input counts:

- 436,200
- 450,196
- 454,565
- 458,562
- 463,584
- 468,299
- 472,092
- 474,443

The router ledger contains **19 successful Z.ai Coding Flash requests above 400K**. The maximum observed successful prompt was **474,443 input tokens**, with 7,252 output tokens and HTTP 200.

After each successful compaction, Codex's token event dropped to roughly 35K-55K. But the next ordinary child turn immediately returned to roughly 450K-474K because ordinary turns reattach the large tool schema while the compaction request itself omits live tools.

The child rollout records contain valid top-level `compacted` records with small replacement histories, and projection/integrity checks have zero gaps, duplicate ordinals, or backward ordinals. So this is not transcript/projection corruption.

## Root cause

The Z.ai Coding descriptor currently says:

```json
"contextWindow": 1000000,
"autoCompact": 400000
```

The 400K threshold was copied from the same-model OpenCode Go conservative pin. Current Z.ai Coding traffic disproves 400K as a usable threshold for this client/tool surface: the provider repeatedly and successfully serves requests above it.

Because compaction omits the live tool schema, it reduces the history but cannot reduce the ordinary-turn static tool prefix. If that ordinary-turn floor is already above `autoCompact`, every successful compaction recreates the condition that triggered it.

## Proposed provider-specific fix

Raise **only** `zai-coding/glm-5.3-flash` to:

```json
"autoCompact": 500000
```

and bump its comp hash.

500K is deliberately conservative:

- it is the smallest round threshold above the measured 474,443-token successful floor;
- it still reserves half of the advertised 1M window;
- it does not change OpenCode Go, OpenRouter, Z.ai API, Ollama Cloud, or Command Code thresholds without provider-specific evidence.

A regression test should make the threshold provider-specific rather than asserting 400K for every Flash route.

I have a tested patch ready: syntax checks pass and the relevant registry/catalog/agent suite passes 134/134.

PracticalSwan:
## Summary

Fixes #846.

This PR addresses the reproduced Z.ai Coding GLM-5.3-Flash compact loop at both layers:

1. the route's 400K auto-compaction threshold sat below its demonstrated successful ordinary-turn floor;
2. Flash lacked the standalone tool-discovery metadata already proven on full-size Z.ai Coding GLM-5.3, so Codex eagerly attached a very large direct tool/schema surface; and
3. once standalone discovery was enabled, Z.ai/LiteLLM returned bridged special function calls with `namespace: null`, while the response relay accepted only an omitted namespace. That prevented valid `tool_search` calls from being restored to native Codex `tool_search_call` items.

## Reproduction

On Codex Desktop `26.915.4065.0` / app-server `0.155.0-alpha.9.2`, two independent `zai-coding/glm-5.3-flash` child threads repeatedly showed:

```
ordinary turn (~436K-472K provider-reported input)
-> compact
-> ~35K-54K compacted turn
-> next ordinary turn returns to ~450K
-> compact again
```

The append-only router ledger contains **19 successful Z.ai Coding Flash requests above 400K**. The highest observed successful prompt was **474,443 input tokens** with non-empty output and HTTP 200.

The compacted rollout records themselves are valid and small; projection checks show no gaps, duplicate ordinals, or backward ordinals. This is not transcript corruption.

The full-size `zai-coding/glm-5.3` precedent in #297 is directly relevant: enabling `searchTool.mode = "standalone"` reduced a tool-heavy fresh-session surface from roughly 491K to 82K while preserving deferred MCP discovery and V2 subagents.

## Changes

Scoped route metadata for **`zai-coding/glm-5.3-flash`**:

- `autoCompact: 400000 -> 500000`
- enable `searchTool: { "mode": "standalone" }`
- use the already-proven Coding Plan GLM behavior template: `gpt-5.6-sol`
- use the `efficient-agentic` instruction overlay
- bump `compHash`
- document why this provider differs from the other Flash routes
- add provider-specific regressions

500K is deliberately conservative: it is the smallest round threshold above the measured 474,443-token successful floor and still reserves half of the advertised 1M window. Other Flash providers retain their separately proven 400K thresholds.

Response relay:

- treat provider `function_call.namespace === null` as the same unqualified wire identity as an omitted namespace
- apply this only at the provider-facing response boundary; native restored identities stay strict
- cover tool-search, custom-tool, function-relay, plain-identity, and raw-codec paths
- add streamed and non-streamed `tool_search` regressions plus custom-tool coverage

This PR deliberately **does not** add checked-in `multiAgentVersion: "v2"`. Repository policy requires an accepted exact-route `v2_agent` proof artifact; family similarity and successful ordinary tool turns are not sufficient evidence.

## Live verification

Installed Z.ai Coding Flash at max reasoning:

- basic response — HTTP 200
- streaming — HTTP 200
- forced function call — HTTP 200
- stateless tool-result continuation — HTTP 200
- compaction — HTTP 200
- subagent-capability tool calling — HTTP 200
- subagent-capability streaming — HTTP 200

Generated Codex catalog after refresh:

- `supports_search_tool = true`
- `context_window = 1000000`
- `auto_compact_token_limit = 500000`
- text + image input
- behavior base instructions match the full-size Coding Plan GLM execution template apart from model identity
- `comp_hash = zai-coding-glm-5-3-flash-v4`

### Deferred discovery A/B

Before the relay fix, a live Flash request correctly called the provider-facing `tool_search` function but the router returned:

```json
{"type":"function_call","name":"tool_search","namespace":null}
```

instead of native `tool_search_call`.

After the relay fix, the same two-turn live probe passes end to end:

1. Flash calls the bridged search control.
2. Router restores native `tool_search_call` with `execution:"client"`.
3. A synthetic search result exposes only `mcp__calendar.delete_event`.
4. The next Flash turn calls `delete_event` in namespace `mcp__calendar` with the requested id.
5. Both turns return HTTP 200.

That verifies deferred discovery itself, rather than only the metadata flag.

## Tests

- Flash + namespace/custom relay focused suite: **146/146 pass**
- installed live-router namespace relay suite after surgical backport: **119/119 pass**
- `npm run check` — pass in upstream worktree and installed live router
- `git diff --check` — pass
- broader registry/catalog/agent/native-drift/tool-surface suites — pass
- full repository `npm test` was run. One untouched Antigravity onboarding assertion fails with `null !== 1`; running that exact test against untouched current `main` reproduces the identical failure, so it is baseline/environmental and unrelated to these changed files. Hosted CI remains authoritative for that unrelated test.

No provider keys, prompts, private thread ids, or capability-bearing router URLs are included here.


PracticalSwan:
Parity hardening follow-up at `1937678`: added a regression that directly compares `zai-coding/glm-5.3-flash` with full `zai-coding/glm-5.3` for the execution contract that should remain shared — 1M context, low/high/max reasoning/default max, `glm-thinking`, standalone discovery, GPT-5.6-Sol behavior template, efficient-agentic overlay, reasoning-summary/apply-patch/parallel-tool capability declarations.

The same test explicitly locks the intentional differences instead of accidentally inheriting them: full GLM remains text-only in registry / 900K compaction / checked-in V2, while Flash remains native text+image / 500K compaction / no checked-in V2 claim. Local selected-mode promotion can still expose Flash as V2 without fabricating an upstream certificate.

Fresh focused suite after this addition: 147/147 pass; `npm run check` and `git diff --check` pass.

PracticalSwan:
Root cause follow-up at `6a0dc78`: the compact-loop pressure was not only tool-search relay correctness. The router was undoing Codex's deferred loading before Z.ai saw the request.

A live-captured-style request already carries client `tool_search`, core/collaboration tools, selected MCP namespaces, and only three active `codex_app` definitions. `chatProviderToolSurface()` unconditionally called `mergeCodexAppTools()` for non-Groq Chat providers, which rehydrated the entire bundled `codex_app` + `plugin_management` snapshot before `flattenToolSearchHistory()`.

Measured with the captured reduced shape:
- client/provider surface intended by Codex: **8 tools / 805 bytes**
- previous routed Z.ai surface: **23 tools / 52,710 bytes**
- eager injection: **+15 definitions / +51,905 bytes**
- after this fix: **8 tools / 805 bytes**, with the tool-search relay still present.

The change is deliberately generic but narrow: when a request contains client-executed `tool_search`, non-Groq Chat routes preserve the reduced client app surface and add back only deferred app definitions already referenced by stored history or an explicit forced choice. Providers without client tool_search retain the existing full-snapshot compatibility fallback. MCP namespaces are not globally pruned or synthesized; the router continues to flatten exactly the MCP namespaces Codex sent.

Verification:
- RED: 2 new deferred-surface regressions failed against the old eager merge.
- GREEN: `test/chat-tool-surface.test.mjs` 19/19.
- full Z.ai Flash router integration proves the mock gateway receives only client-visible app tools.
- namespace/app/custom relay suite: 164/164.
- Flash parity suite: 7/7.
- direct `scripts-check.mjs` passes (`v2-agent applications valid (7)`, syntax checks passed).
- `git diff --check` passes.

This complements the earlier 500K compaction floor and namespace-null relay fix: discovery can now actually reduce the provider-facing schema instead of coexisting with a fully injected app catalog.

PracticalSwan:
Root cause follow-up at `6a0dc78`: the compact-loop pressure was not only tool-search relay correctness. The router was undoing Codex's deferred loading before Z.ai saw the request.

A live-captured-style request already carries client `tool_search`, core/collaboration tools, selected MCP namespaces, and only three active `codex_app` definitions. `chatProviderToolSurface()` unconditionally called `mergeCodexAppTools()` for non-Groq Chat providers, which rehydrated the entire bundled `codex_app` + `plugin_management` snapshot before `flattenToolSearchHistory()`.

Measured with the captured reduced shape:
- client/provider surface intended by Codex: **8 tools / 805 bytes**
- previous routed Z.ai surface: **23 tools / 52,710 bytes**
- eager injection: **+15 definitions / +51,905 bytes**
- after this fix: **8 tools / 805 bytes**, with the tool-search relay still present.

The change is deliberately generic but narrow: when a request contains client-executed `tool_search`, non-Groq Chat routes preserve the reduced client app surface and add back only deferred app definitions already referenced by stored history or an explicit forced choice. Providers without client tool_search retain the existing full-snapshot compatibility fallback. MCP namespaces are not globally pruned or synthesized; the router continues to flatten exactly the MCP namespaces Codex sent.

Verification:
- RED: 2 new deferred-surface regressions failed against the old eager merge.
- GREEN: `test/chat-tool-surface.test.mjs` 19/19.
- full Z.ai Flash router integration proves the mock gateway receives only client-visible app tools.
- namespace/app/custom relay suite: 164/164.
- Flash parity suite: 7/7.
- direct `scripts-check.mjs` passes (`v2-agent applications valid (7)`, syntax checks passed).
- `git diff --check` passes.

This complements the earlier 500K compaction floor and namespace-null relay fix: discovery can now actually reduce the provider-facing schema instead of coexisting with a fully injected app catalog.

PracticalSwan:
Follow-up at `98ab15c`: the first broad Ubuntu matrix exposed two pre-existing integration assertions that encoded eager app injection as part of the expected behavior.

- The 149-tool expectation was exactly 15 tools above the new discovery-first surface. It now expects 134 = 110 core + tool_search + 3 client app tools + 20 already-discovered tools, and explicitly rejects undiscovered `create_thread`/plugin uninstall injection.
- The non-streaming mock returned `codex_app__send_message_to_thread` without that tool ever being advertised, discovered, referenced, or forced. The fixture now adds a real historical `codex_app/send_message_to_thread` call/output before expecting namespace restoration, which exercises the intended "reintroduce only a referenced deferred definition" path.

Fresh cross-provider integration coverage passes 59/59, direct `scripts-check.mjs` passes, and `git diff --check` passes. A new full local `npm test` and hosted CI run are in progress from this stable head.

The separate macOS Grok protocol failure on the previous head is unrelated: that fixture contains no `tool_search`, Ubuntu+Windows passed it, and macOS timed out only while waiting for an existing 60s cancellation-close invariant. The GitHub integration lacks Actions write permission to rerun the failed job (403), so no unrelated cancellation code was changed.

PracticalSwan:
Live quota-backed validation found one additional Chat-route compatibility defect after the deferred-catalog fix: native `tool_choice: {type:"tool_search", execution:"client"}` reached Z.ai unchanged even though the tool definition had already been flattened into a function. Z.ai rejected it as an invalid OpenAI tool choice.

Root cause: `router.mjs` flattened `tool_choice` only for Groq/Command Code inside the generic namespace-flattening path. Commit `53a4d47` extends that existing rewrite to all Chat-completions routes while preserving the existing provider-specific cases.

Evidence:
- live Z.ai Flash probe reproduced HTTP 500 with the untranslated native tool_search choice.
- new full-router Z.ai regression failed RED with actual `{type:"tool_search", execution:"client"}` vs expected `{type:"function", name:"tool_search"}`.
- after the one-line condition fix, the exact regression passes.
- combined relay/model suite: 192/192.
- `scripts-check.mjs`: passes (7 v2-agent applications; syntax clean).
- `git diff --check`: clean.

This is separate from, but complementary to, the eager catalog-injection fix: discovery now has both a reduced provider-facing catalog and a provider-valid forced search choice.

PracticalSwan:
Post-restart live validation is now complete for the new discovery path.

Installed into the customized live 0.5.1 runtime with scoped rollback coverage:
1. deferred app/plugin catalog preservation from `6a0dc78`
2. generic Chat-route tool-choice flattening from `53a4d47`

Quota-backed exact-route Z.ai Flash result after the managed restart:
- initial provider surface: **8 tools / 805 bytes**
- eager `codex_app__create_thread`: absent
- eager `plugin_management__uninstall_plugin`: absent
- turn 1 returned native `tool_search_call` with client execution
- after supplying a tool_search_output containing only `codex_app/send_message_to_thread`, turn 2 returned a restored native `function_call` with `namespace: codex_app`, `name: send_message_to_thread`
- both routed turns returned HTTP 200
- router health after the probe: HTTP 200, idle, degraded=[]
- projection critical_count=0; targeted current-generation router error counters all zero

The same live probe had previously reproduced the forced-search bug as HTTP 500, so this is a direct before/after validation of the second fix.

PracticalSwan:
Follow-up optimization verification on current head 227543813b79e3c10aad893f5c7038475bde2a78:

- The Flash route-level default remains `max`; routine-effort policy is intentionally local to user specialist configs and is not part of this PR.
- Generic privacy-bounded usage telemetry records final provider-facing `reasoningEffort`, `providerToolCount`, and `providerToolSchemaBytes`.
- Live exact-route Z.ai Flash verification after managed restart passed deferred discovery with 8 provider-visible tools / 805 serialized schema bytes, restored native client `tool_search_call`, and restored `codex_app/send_message_to_thread`.
- Live usage rows recorded `high` and `max` effort plus provider-facing tool counts/schema bytes, with no prompt/input/tool-schema/argument/session fields persisted.
- Routed tool-result aging is enabled through the existing supported feature path; native aging remains off. Focused installed tests pass and `scripts-check.mjs` passes.
- The branch was merge-forwarded (no rebase/history rewrite) onto upstream `9c0db45676238d032757370ec4010b66b6759dd8`. The final PR delta against current main remains the intended 15-file Flash/deferred-tool/telemetry patch.
- Fresh merged-tree local verification passes: `npm run check`, `scripts-check.mjs`, `git diff --check`, and 296/296 combined Flash/routing/usage/aging/subagent/reviewer/failover tests.
- Fresh hosted verification on `2275438` is fully green: Provider Model Discovery, Grok apply_patch protocol, Ubuntu/Windows/macOS core tests, Windows/Linux Electron control center, Unified native macOS app, and Homebrew formula all pass. The Homebrew install job is intentionally skipped by workflow conditions.

No personal agent-tiering policy is included in the upstream diff.

Issue #804, opened 2026-09-17
feature:  Support Custom Router Models as Codex Subagents


ifquant:
I would like Codex to support using custom Router models as delegated subagents while keeping the primary Codex session on a different
  model.

  For example, the main session could continue running on gpt-6-astra with medium reasoning, while a delegated subagent runs on deepseek/
  deepseek-v4.1-flash with high reasoning for implementation, testing, or code review tasks.

  The requested capability should allow:

  - Selecting an available Router model as a subagent model.
  - Configuring the subagent’s reasoning effort independently from the parent session.
  - Running the subagent with the parent session’s workspace, tools, permissions, and task context.
  - Returning the subagent’s changes, findings, verification results, and status to the parent Codex session.
  - Preserving clear task and usage attribution between the parent and delegated model.
  - Providing a clear explanation when a model requires additional compatibility or collaboration certification.

  I am not asking Codex to silently treat an uncertified model as a native v2 subagent. I would like a supported, controlled delegation path
  for available custom models, with explicit compatibility checks and transparent failure reasons when a route cannot be used.

duolahypercho:
This is implemented on `main`:
- `control subagents mode proven|selected|all` and `subagents set <slug> on|off` choose which routed models can be subagents.
- #811 (d7a78fd) explains why a route can't be delegated to.
- #773 and #851 apply the configured subagent reasoning effort.

The `codex-subagents` skill and the AGENTS.md section "Subagent capability is researched, not asserted" document the whole flow.

Two limits are deliberate. A route becomes a full (v2) subagent only once its checked-in registry entry has passed the capability checks. Subagent effort is one global setting, not set per subagent.

Closing as implemented. Please open a narrower issue if something specific is still missing.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #849, opened 2026-09-19
fix: stop the bundled skill documenting a subagent pin the router removed

Corpocore:
## Problem

An explicit `spawn_agent.model` is kept; only a call that omits the model inherits the routed parent. That behavior shipped in #742, together with a rewrite of `.claude/skills/codex-subagents/SKILL.md` and an update to `docs/HOW-IT-WORKS.md`.

Two places kept describing the behavior #742 removed:

- `skills/codex-router/SKILL.md`: "In-session subagents are always pinned to the routed parent's model because their tool arguments are model-generated and must not silently cross a provider or billing boundary."
- `src/namespace-relay.mjs`, the comment above `SPAWN_MODEL_TOOLS`: "An in-session subagent is always pinned to the routed parent: its model argument is generated by the parent model, not an independent user choice..."

The first one is not internal documentation. `skills/codex-router/SKILL.md` is installed into every user's `~/.codex/skills/codex-router/`, so it is what an operator, or a delegated agent reading its own instructions, sees while diagnosing a child that ran on the wrong model. It says the router discards the override on purpose.

## Why nothing caught it

The suite exercises the rewrite, not the prose describing it. `test/subagent-model-inherit.test.mjs` covers the keep-or-inherit split at the wire level and passes on both sides of the drift.

## Change

- Correct the skill's inheritance paragraph: an explicit `spawn_agent.model` is kept, delegating to another model means naming it, only an omitted model inherits the parent, and Codex validates the name against what it advertises for `spawn_agent`.
- Correct the comment above `SPAWN_MODEL_TOOLS` to match.
- Add `test/subagent-override-docs.test.mjs`, a source assertion over the two shipped skills and that comment, so the stale claim cannot come back silently.
- CHANGELOG entry under Unreleased.

No behavior change. In particular this does **not** touch `sanitizeSpawnAgentModel`, which still clears a model outside the client's advertised enum, after which the call inherits the parent; that guard is #632's and stays as it is.

## Not in this PR

The `subagents explain <model>` command that the troubleshooting path really wants is #804, still on `feat/804-subagent-explain` and not on `main`. Nothing here depends on it.

## Testing

- `node --test test/subagent-override-docs.test.mjs test/namespace-relay.test.mjs test/subagent-model-inherit.test.mjs` -- **150 pass, 0 fail**, including the two new assertions.
- `npm run check` -- syntax checks clean, `v2-agent applications valid (7)`.


duolahypercho:
Thanks — removing documentation for a subagent pin the router no longer honors is the right call. CI is red on this revision, and it is caused by the change itself rather than by flake or drift.

### What's failing

`test (ubuntu-latest)`, `test (macos-latest)` and `test (windows-latest)` all fail the same case:

```
✖ every pack skill has valid frontmatter, a trigger description, and stays short
```

That test guards the bundled skill files this PR edits, so the edit has pushed one of them outside what the contract allows — most likely the frontmatter shape or the length bound rather than the prose itself.

### What it needs

Run that test locally against the edited skill and adjust to fit the contract. If the pin really has to be documented somewhere longer-form, the skill body is the wrong home for it given the length bound — the docs tree is the better place, with the skill keeping only the short trigger description.

duolahypercho:
Thank you @Corpocore — this landed via #904, which carries your commit unchanged (authorship preserved) plus one small commit trimming skills/codex-router/SKILL.md back under its 100-line budget, which is what failed this PR's CI on all three OSes. Closing as merged through #904.

PR #850, opened 2026-09-20
Bound the image payload on the routed request path

areebbajwa:
## Why

A conversation replays every image it still holds on every following turn, so a
session that works from screenshots grows until one request crosses the
provider's per-request image ceiling and the turn fails outright. Measured on
OpenRouter from this host on 2026-09-17: about 12MB of decoded image content was
accepted and about 31MB was refused with `HTTP 413: Downloaded image content
cannot exceed 30MB`. The reference incident was a session that had accumulated
1,791 image references before its turns started failing - a stalled agent, not a
slow one.

The per-image token estimate added in #813 bounds what one image is charged, but
nothing bounded how many images ride along. This bounds the payload itself.

## What it does

On the routed request path (`prepareRoutedRequest`), after tool-result aging,
the payload is bounded before it leaves:

- **Byte budget: 20MB of decoded image content** (`IMAGE_PAYLOAD_BUDGET_BYTES`),
  deliberately below the provider's 30MB ceiling so a provider that counts
  differently, or the text of the same request, still fits.
- **Token budget: 128K image tokens** (`IMAGE_PAYLOAD_BUDGET_TOKENS`), charged at
  the route's per-image bound (`maxImageTokensForRoute`: 4096 on a resold route,
  1024 on the three documented direct DeepSeek Flash models). Bytes bound the
  request; tokens bound the recurring per-turn bill that a byte cap alone does
  not, since under 20MB a hundred small screenshots still ride along and are
  charged again every turn.
- Oldest image references become short text receipts until the turn fits both
  budgets; **the newest two images are always kept**, even if one of them would
  exceed the budget alone.
- Images inside tool results are bounded too. Remote `https` image URLs are
  **not** measured (only `data:image/` references are); Codex sends data URLs, so
  that path is not reachable today, but it is a known limit if a surface ever
  forwards a remote image.
- The counts - `imageReferencesSeen`, `imageReferencesDropped`,
  `imageBytesSaved`, `imageTokensSaved` - are recorded on the usage event.

The bound is a flat constant rather than a fraction of the route's context
window on purpose: a bigger window does not make an image cheaper, it only
delays compaction, and what is being bounded is recurring spend per turn.

## Known gap (deliberately not covered here)

Only the routed path is bounded in this change. There are three `ageToolResults`
call sites in `router.mjs` - the routed path, the native path (the `else` branch
of `if (route)`, around line 4337) and the compaction path - and only the routed
one bounds images. The
native and compaction paths age tool results by size and leave images entirely
unbounded. That is not cosmetic: the native path leaves as a stateless full
conversation, so an old image costs its full size on every turn of that path
too, which is the same failure this change addresses. Those two paths are
knowingly left for a follow-up PR rather than wired blind here, because the
native path was not exercised in this change.

## Notes for review

- `...imageBudget` is spread on all five usage-event rows, not just the success
  row, so the counts survive on the error and retry rows. It is `undefined` on
  paths where the bound did not run, and spreading `undefined` adds nothing.
- The trim summary log line stays inside the existing `if (!QUIET)` gate.
  Nothing upstream hard-sets `CODEX_ROUTER_QUIET`, so the gating is correct for
  every deployment.
- The per-image default comes from `DEFAULT_IMAGE_TOKEN_BOUND` in
  `prompt-image-usage.mjs` (added in #813) rather than a second local 4096.

## Tests

`node --test test/prompt-image-budget.test.mjs` - 11/11.
`node --test test/response-usage.test.mjs` - 47/47.
Run under the runtime's Node v24 (the `JSON.parse` reviver `context` argument the
estimate path uses needs Node 21+; system Node 20 gives false failures there).


areebbajwa:
The `macos-latest / python 3.12` check is red on this PR. I do not think it is
caused by this change, and I cannot re-run it from a fork, so I am leaving the
evidence here rather than force-pushing.

The annotation points at the apply_patch guidance step, but the guidance checks
all pass in the log. The failure is the last cancellation sub-check:
`cancellation did not reach mock upstream within 60000 ms on darwin`
(verify-grok-apply-patch-guidance.mjs:778). The two earlier cancellation checks
in the same run closed in 12 ms and 18 ms, and ubuntu and windows passed on this
same commit.

The same job failed today on an unrelated branch, run 35494529067 on
fix/zai-flash-compaction-loop, so it does not look specific to this PR.

For what it is worth on the diff itself: with no image references
`boundImagePayload` returns the input array unchanged, so a text-only request
leaves this path byte-identical to before.

Happy to rebase or change anything if you would rather; otherwise a re-run of
that one job would confirm it.

duolahypercho:
Reviewed and this looks good to merge — the trim logic keeps the newest two images, the estimate is conservative, and what is logged matches what is sent. The noted gap (native and compaction paths still unbounded) is fine to leave as a follow-up.

It cannot merge as-is: #817 and #836 landed first and GitHub now reports conflicts in `CHANGELOG.md` and `src/router.mjs`, and its own "Update branch" button cannot resolve them.

I merged current `main` into your branch locally and it resolved cleanly, with `test/prompt-image-budget.test.mjs` and `test/usage-events.test.mjs` passing 27/27 afterwards. The result is pushed here if it saves you time: `rescue/850-image-token-budget` in this repository.

Could you rebase or merge `main` into `image-token-budget` and push? Then this goes straight in. (I could not push to your fork from here.)

PR #855, opened 2026-09-21
fix: bound the upstream connect phase and derive the retry budget from it

4Liberty:
## What

`src/upstream-retry.mjs` lists `UND_ERR_CONNECT_TIMEOUT` as retryable, but the loop's pre-retry budget was a fixed 5s while the socket was allowed undici's 10s default to connect. By the time the failure arrived the attempt had already spent the budget, so no retry was ever started and the blip was relayed to the client as a 502.

Measured on 2026-09-21: one incident logged **454 connect timeouts and zero connect retries**, each failure costing its caller ~10.2s. A blackholed address measured 10.0s to fail.

## Change

- `fetchDispatcherOptions()` carries an explicit `connectTimeout` (3s; `CODEX_ROUTER_CONNECT_TIMEOUT_MS`, clamped 0.5–30s) and races the resolved addresses (`autoSelectFamily`, 250ms per attempt) instead of serializing one dead anycast address in front of a healthy one.
- `src/upstream-retry.mjs` derives its default budget from that same bound (`3 x connectTimeout`, 9s by default) so the two cannot drift apart again. A fixed budget against a configured connect bound was the actual defect.
- Worst case for a request that does fail is unchanged at ~10s — three bounded attempts plus backoff, where one undici-default attempt used to be. A *slow* failure is still relayed untouched: a 504 the edge spent half a minute producing is not retried.
- `AGENTS.md` rule 3, `CHANGELOG.md` and `docs/TROUBLESHOOTING.md` updated; the troubleshooting entry tells a bound that is too small apart from a network path that is actually broken.
- The spawned-router test waits now tolerate a loaded host (5s → 30s, still polling at 50ms and still failing immediately when the child exits), because that gate was itself timing out under load and blocking the very fix it guards.

## Verification

- New tests: a bounded connect timeout is affordable for every retry (three attempts under the default budget and backoff); a connect timeout that outlives the budget is still relayed; the dispatcher carries the configured bound and clamps the override; the retry budget covers every bounded attempt it allows. Existing tests for relayed bytes, aborts and slow failures are unchanged.
- `node --test test/fetch-transport.test.mjs test/native-retry.test.mjs` → 32/32.
- Blackholed connect (`10.255.255.1:443`) measured **3.5s** with the new bound, against **10.0s** before.

## Risk

A connect timeout cannot double-execute work: no connection was established, so nothing was sent. The existing `canRetry`/`headersSent` backstop is untouched, and the budget still refuses slow failures.


duolahypercho:
Heads-up: this now conflicts with `main` because of a merge underneath you, and the overlap is substantive rather than incidental.

#874 ("cap in-flight streams and match the header idle bound") landed earlier today and changes the same transport bounds:

```
CONFLICT (content): Merge conflict in src/fetch-transport.mjs
CONFLICT (content): Merge conflict in test/fetch-transport.test.mjs
```

Specifically, `installStableFetchTransport` now sets `headersTimeout` alongside `bodyTimeout` when a caller passes `bodyTimeoutMs`, on the reasoning that a Grok turn can stay silent until the first response headers and Undici's 300s headers default would abort that wait while the body bound still had room. Your PR bounds the connect phase and derives the retry budget from it, which is aimed at the same window from the other side.

Worth rebasing and then deciding what is still needed: some of the connect-phase bound may be covered by the headers bound now, or the two may compose — but the retry-budget derivation looks orthogonal and still valuable. Apologies for the churn; that is the merge order's fault, not yours.

PR #856, opened 2026-09-21
fix: keep startup gates from killing the service on a loaded host

4Liberty:
## What

Three startup gates treated a scheduling artifact as a defect, and a one-minute task heartbeat turned each failure into a loop that could not converge:

1. **The gateway config ACL hardening was fatal.** `writeLiteLlmConfig()` hardened four files per start through a short-lived PowerShell helper and threw when that helper timed out. On a busy host every start died there.
2. **The interpreter probe was fatal on a timeout.** `venvRuntimeProblem()` already documents that a `spawnSync` timeout "is not proof of a broken virtual environment", but `start.mjs` threw on it anyway.
3. **The gateway's cold start was allowed 300s.** A starved Python import can take longer; the launching scheduled task then relaunched `start.mjs` after each failure and killed the in-flight import — the failure mode the code's own comment warns "loops forever".

Measured on 2026-09-21: the router was down for ~40 minutes on a host that was busy with unrelated work, because each attempt killed the previous import at the 300s mark.

## Change

- `src/litellm-config.mjs`: the ACL hardening is attempted, warned about on failure, and never fatal on the startup path. The files still inherit the owner-only ACL of the state directory they are written into, and callers that must not continue without the hardening still receive the error.
- `src/venv-runtime.mjs`: new `isTransientVenvProblem()`; `src/start.mjs` logs a warning for a timed-out probe and lets the gateway's own readiness gate be the judge. A genuinely broken environment (missing interpreter, non-zero exit) is still fatal.
- Cold-start allowance raised from 300s to 900s in `src/start.mjs`, with the matching waits in `src/service.mjs`, `src/service-readiness.mjs`, `src/router-restart.mjs` and `src/wait-health.mjs`, and the `src/health-backoff.mjs` comment updated to match.

## Verification

- `node --check` on every touched file; focused test set 78/78 (`node --test test/responses-heartbeat.test.mjs test/empty-completion-router.test.mjs test/fetch-transport.test.mjs test/native-retry.test.mjs`).
- With the change: a cold start on a loaded host completes instead of being killed mid-import; a start after the ACL helper times out proceeds with a warning.
- Simulated a fresh upstream checkout and rebuilt on top of it: the resulting tree is identical across two independent hosts.

## Risk

The longer allowance only changes how long a start waits for a gateway that is still initializing — a broken environment still fails, and the readiness gate that decides "healthy" is unchanged. The ACL change keeps the attempt; it only stops a transient timeout from taking the whole service down.


duolahypercho:
Thanks — the motivation is sound (a starved host should not have its own startup gate kill the service), and the `isTransientVenvProblem` split plus the best-effort ACL hardening both read well. CI is red though, and it is a real failure rather than a flake.

### What's failing

`test (ubuntu-latest)`, `test (macos-latest)` and `test (windows-latest)` all fail, roughly twenty distinct tests each, including:

```
✖ an impossible inherited restart deadline is refused before overlay mutation
✖ an environment-backed mutation waits for service ownership through publication
✖ a live foreground router blocks environment pool publication before metadata changes
✖ a completed local pull checks tool-capable models on automatically
✖ API forwarder never replays embeddings through a provider API-key pool
```

### Cause

The PR raises `300_000` to `900_000` in `router-restart.mjs`, `service-readiness.mjs`, `service.mjs`, `start.mjs` and `wait-health.mjs` but ships no test changes. Several suites pin that arithmetic — `ROUTER_SERVICE_RESTART_MINIMUM_MS` is derived from `SERVICE_READINESS_ALLOWANCE_MS`, so tests asserting the minimum restart deadline (and the "impossible deadline is refused" case built on it) now compute a different number. `test/router-restart.test.mjs` and `test/service-readiness.test.mjs` are the direct ones; the pool/publication tests inherit the same budget.

### What it needs

Update the tests that encode the old budget alongside the constants. Worth deciding explicitly whether the tests should assert the *derivation* rather than a literal, so the next tuning pass does not have to touch them again.

One question on the change itself: 900s is a long time to hold a start attempt open on a host that is genuinely broken rather than merely busy. The commit message motivates it from a Windows task relaunching every 60s and killing the in-flight import — is the right fix there the longer health budget, or the heartbeat relaunching a start that is still making progress? Happy either way, just want the reasoning recorded.

duolahypercho:
This still fails its tests. CI is red on all three OSes. On a trial merge with the #890/#892/#895 stack, `router-restart` has 1 failure, and `model-overlay-publication` plus `provider-api-key-control` have 12, including "impossible inherited restart deadline". Those are the tests the earlier review asked to update. Ideally they should check how the budget is derived rather than assert a literal number.

Two more things:
- **ACL hardening.** Wrapping `protectPrivateFile` makes hardening warn-only across all of `litellm-config`, and it also swallows `chmod` failures on POSIX. #890 adds an opt-in `hardenFailure: "warn"` and leaves the default at throw. Please rebase onto that and opt in only where it's needed. Also, the comment says the config "holds provider keys", but it stores `os.environ/…` references, not the keys themselves.
- **Changelog.** There's no new test and no changelog fragment (`changelog.d/<short-slug>.md`).

The maintainer's design question is still open: extend the budget to 900 s, or stop the heartbeat from relaunching a start that is still in progress? #895's back-off now covers part of the second option.

---

PR #839, opened 2026-09-19
Improve Simplified Chinese coverage across desktop and panel UI

MJYKIM99:
## Summary

This builds on the Simplified Chinese support contributed in #181 and integrated in #194. Thanks to Yuxin-Qiao and the maintainers for the existing localization foundation.

With Simplified Chinese selected, some Control Center pages, native menus, widget text, and browser-panel messages still appear in English. This PR fills those coverage gaps using the existing language selection and fallback behavior.

- Localize all eight Control Center pages, shared controls, search, dialogs, accessible labels, chart labels, dates, and known setup/progress messages.
- Localize Electron menus while preserving their native roles and shortcuts; extend the macOS tray/notch and widget strings, including widget configuration/gallery metadata.
- Fill browser-panel gaps for activity states, quota labels, reasoning-effort labels, and tool-result compaction summaries.
- Add regression coverage for Chinese rendering, language persistence, fallback behavior, and compatibility with existing widget snapshots.

## Compatibility and review notes

Model/provider identifiers, commands, paths, user content, and unknown provider diagnostics pass through unchanged. Localized effort labels keep their original submitted values (for example, `最高 (max)` still submits `max`). English remains available, and other existing locales retain their previous English fallback for newly covered copy; the additional dictionary entries preserve the repository's locale-key parity checks.

The widget snapshot gains an optional presentation-language field. Older snapshots still decode, absent values remain omitted, and a language change triggers a new snapshot. Both the Xcode project and its `project.yml` source include the Chinese widget resource.

This does not change provider routing, credentials, authentication, model selection, or the network protocol. The scope is UI presentation plus the backward-compatible widget language field.

## Initial validation (e1a3459)

The original localization commit was validated on macOS 27 / Apple silicon, Node 24.19, Swift 6.4 and Xcode 27. That commit has exactly the same Git tree as the build that was tested, installed, and inspected locally. The following results describe that original build, not a fresh execution of those suites for the subsequent test-only correction.

| Check | Result |
| --- | --- |
| Root syntax check and Control Center TypeScript check | Passed |
| Focused Chinese localization/menu/backend regression tests rerun for PR preparation | 16 passed |
| Control Center and affected root tests | 109 passed; 1 Windows-only skip |
| Full Control Center suite during packaging, including renderer scenarios | 101 passed; 1 Windows-only skip |
| Browser panel unit/browser tests | 54 passed |
| Native boundary/source/install tests | 72 passed |
| Swift tray tests | 136 passed in 21 suites |
[...]

PR #857, opened 2026-09-21
feat(i18n): complete the zh-TW interface for the Control Center and tray panel

coolka-hsu:
## What this changes

Completes the Traditional Chinese (zh-TW) interface for both desktop surfaces. Pages the earlier localization pass had not reached still rendered hardcoded English while the language was set to zh-TW: the whole Settings "ChatGPT accounts" section, every action label behind `runAction` and the optimistic toggles, the model directory, the shared formatters in `lib.ts`, and the tray panel's status, usage, connections, vision, local-model, and Ollama-catalog surfaces. The Windows tray context menu was English too, because the main process owns no dictionary of its own.

## Details

- **Dictionaries.** EN and ZH_TW both define every key: 1188/1188 in the Control Center and 386/386 in the panel. The other locales keep their existing overlays and fall back to English for the new keys, which is what they rendered before this change. The panel's key test now states that contract explicitly: zh-TW must stay complete, and another locale may omit a key because `t()` inherits the English source text.
- **Shared helpers.** `effortLabel` moves into `lib.ts` so the Models, Local, and Settings pages agree. The panel's Ollama-catalog "fit" column keeps a stable id for sorting and translates only the label beside it, so wording can no longer reorder the list.
- **Tray menu.** The context menu is built in the main process, which has no dictionary, so the renderer hands it the two labels it already resolved over a validated `setTrayLabels` IPC channel. English stays as the pre-render default for a tray-only launch.
- **Provider plan notes.** The panel's zh-CN-only note table becomes a per-language table and gains zh-TW.
- **Fixed in passing:** the zh-TW `models.subagentSummary` entry appended the English plural "s" to a Chinese sentence.

## Tests

- Source-copy assertions that pinned English strings to page sources now check the key the page renders plus its dictionary entry. The same stale marker fix is applied in `test/chatgpt-session-desktop.test.mjs`.
- The panel suites declare the language they assert on: `test/panel-ui.test.mjs` pins `en`, and the browser E2E passes `locale: "en-US"`. Both previously compared translated labels against English expectations, so they failed on a machine whose locale is not English.
- New: the tray context menu follows the interface language the renderer resolved.

## Verification

`npm run check`, plus the control-center, panel, and desktop-panel browser E2E targets, pass locally on Windows.

## Unrelated note

While rebuilding the packaged companion on Windows, `codex-router.ps1`'s Control Center update transaction could not complete: `Restore-ControlCenterTaskSnapshot` compares the task SDDL byte-for-byte, but Windows normalises the DACL (`D:AI` plus ACE reordering) as soon as the task is re-registered, so the restore never matches. Out of scope for this PR; happy to open a separate issue with the reproduction.


MJYKIM99:
Heads-up: the Traditional Chinese work here is being carried forward in #839.

The Chinese UI integration in #839 includes #857 and keeps the original commit `d795cb9` in history, so the authorship is intact. It is not a re-implementation. It is combined with the Simplified Chinese catalogs so there is one translation engine with separate zh-CN / zh-TW catalogs, instead of two parallel i18n paths that would conflict on the shared locale files.

No action is needed from the author. Maintainers can decide whether to close #857 in favor of #839. Thanks for the original contribution.

duolahypercho:
Heads-up, and thank you — your Traditional Chinese work has landed, but not through this PR, and you should hear that from us rather than discover it.

#839 merged today and carries your translations with attribution. `apps/panel/messages.mjs` opens with:

```
// Includes Traditional Chinese translations from PR #857 by @coolka-hsu (MIT).
```

zh-TW is now complete in both surfaces — 1545/1545 keys in the Control Center dictionary and 545/545 in the panel catalog, with placeholder parity verified against English.

So this PR is likely redundant now. Before closing it, it is worth a pass to see whether anything here is *not* covered by what #839 absorbed — if #839 took the panel and Control Center strings but missed something you had done elsewhere, that part is still wanted and we would rather land it than lose it. If it is fully covered, feel free to close, and sorry it got there by a longer road than it should have.

Separately: your #858 merged today as well, so thanks twice over.

